PR #201, opened 2014-07-01
Add "interactive" option to zendmd

ssbunyk:
It could allow performing some initialization before using zendmd.


ssbunyk:
Could be done using this tip: http://wiki.zenoss.org/ZenDMD_Tips#ZenDMD_configuration_file

PR #235, opened 2014-07-29
correct sql dump file

suvorovd:


tgjonestx:
Closing since the code has changed enough, it's not clear if the proposed change is necessary

PR #383, opened 2014-09-13
Coerce dicts to ObjectMaps when set via a RelMap.  This fixes a regressi...

iancmcc:
...on in this behavior


youngklee:
Closing due to the lack of any follow-up activities

PR #515, opened 2014-10-21
ZEN-7489: (port) If one instance of wrapped function is already running,...

smousa:
... other instances will be paused and then use its output


youngklee:
Closing due to the lack of activities

PR #572, opened 2014-11-11
Removed callback functions for fixing ZEN-14350

ss-mskrynnikov:
Some changes that fix disabling of the button


jpeacock-zenoss:
Fixed by another PR.

PR #664, opened 2015-01-05
fixes ZEN-15857 was looking up the severity icon for every tree object

jhanson:


Products/Zuul/tree.py
```diff
@@ -53,7 +53,7 @@ def __init__(self, ob, root=None, parent=None):
         self._parent = parent or None
         self._severity = None
         # allow showing the event severity icons to be configurable
-        if not hasattr(self._root, 'showSeverityIcons'):
```
jhanson: There was a typo that was causing the severity icon lookup to happen for every node instead of just the root.

Products/Zuul/routers/__init__.py
```diff
@@ -174,7 +174,14 @@ def asyncGetTree(self, id=None, additionalKeys=()):
             childData = Marshaller(child).marshal(keys)
             # set children so that there is not an expanding
             # icon next to this child
-            if len(child.children) == 0:
+            # see if there are any subtypes so we can show or not show the
+            # expanding icon without serializing all the children.
+            # Note that this is a performance optimization see ZEN-15857
+            organizer = child._get_object()
+            # this looks at the children's type without waking them up
+            hasChildren = organizer.meta_type in [o['meta_type'] for o in organizer._objects]
```
epowell: I suggest using a safe `get` call rather than `o['meta_type']` here - I suspect that the TreeRouter may be inherited and used for less well-behaved objects.

epowell: Actually, you don't need a whole list comprehension here - stopping on the first one that matches would be enough. How about an `any` call?

Products/Zuul/tree.py
epowell: :+1:

Products/Zuul/routers/__init__.py
```diff
@@ -174,7 +174,14 @@ def asyncGetTree(self, id=None, additionalKeys=()):
             childData = Marshaller(child).marshal(keys)
```
epowell: You could move this to an `else` below the `hasChildren` check to avoid marshalling when it's not needed.

epowell: I was looking at `childData['children'] = []` and thinking it was `childData = []`
:+1:

PR #670, opened 2015-01-06
fixes ZEN-15927 if the component tree isn't loaded then add a listener

jhanson:


Products/ZenUI3/browser/resources/js/zenoss/devdetail.js
```diff
@@ -743,8 +743,13 @@ Ext.define('Zenoss.DeviceDetailNav', {
                     var pieces = token.split(Ext.util.History.DELIMITER),
                     type = pieces[0], rest = pieces[1];
                     var tosel = componentRootNode.findChild('id', type);
-                    selectOnRender(tosel, sm);
-                    return Ext.getCmp('component_card').selectByToken(rest);
+                    if (!tosel) {
```
jhanson: Was regressed by https://github.com/zenoss/zenoss-prodbin/pull/656/files the component tree is now loaded later than it previously was.

PR #673, opened 2015-01-08
Fix docstrings in zenmodeler which are lying.

ssbunyk:
I've checked the type using print. :)


tgjonestx:
Closing all zenoss-prodbin PRs created prior to 2017 because the backlog of open (abandoned?) PRs is creating extra, unnecessary work in our CI server.  This PR can be reopened if/when it's ready to be merged into the latest code base.

PR #685, opened 2015-01-19
ZEN-15995 When MultiRealm Zenpack is installed, event transform to chang...

romanbakaleyko:
...e device does not change the rest of the device context.


romanbakaleyko:
port https://github.com/zenoss/4.2.5-RPS/pull/330

PR #697, opened 2015-01-28
Add a dummy license

alapidas:
This is necessary to run Zenoss w/devimg, and should be replaced before
shipping the actual product


wplu:
ship

PR #700, opened 2015-01-28
fixes ZEN-16089: Mismatch of Columns & Data in Performance Reports -> Availability Report

itshane:


Products/ZenReports/reports/Performance_Reports/Availability_Report.rpt
```diff
@@ -21,8 +21,10 @@
                          zem.defaultAvailabilityEnd());
     objects python:zem.getAvailability(tableState);
     batch python:tm.getBatch(tableName,objects);
-	exportFields python:['device', 'systems', 'availability'];
-	">
+    exportFields python:[('title', 'Device'),
```
epowell: Does this work OK on devices without a title? Usually we use device.titleOrId() for things like this.

itshane: http://pix.my/dD81WJc0

PR #716, opened 2015-02-03
ZEN-15723: Add Export button to device list view on Infrastructure page

byshovets:
For https://jira.zenoss.com/browse/ZEN-15723

### DIFF
--- a/Products/ZenUI3/browser/configure.zcml
+++ b/Products/ZenUI3/browser/configure.zcml
@@ -137,6 +137,13 @@
     />
 
 <browser:page
+    name="export_devices"
+    class=".device_export.DeviceExporter"
+    for="*"
+    permission="zope2.Public"
+    />
+
+<browser:page
     class=".pages.DeviceDetails"
     name="devicedetail"
     for="Products.ZenModel.Device.Device"
--- a/Products/ZenUI3/browser/device_export.py
+++ b/Products/ZenUI3/browser/device_export.py
@@ -26,7 +26,7 @@ class DeviceExporter(BrowserView):
 
     def _query(self, params):
         device_router = DeviceFacade(self.context)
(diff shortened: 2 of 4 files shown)


keruzu:
While I understand the desire to export devices, the preferred export format should be zenbatchload format.  See ZenModel/BatchDeviceDumper.py BatchDeviceDumper class, along with unit tests for dumping and loading.

The exportXml() -style method of doing exports is going to lead to some serious issues (special device classes like VMware, component export, relations, etc). ie the ones that caused the dumpdevices and deviceload scripts to be deprecated in 4.x.

jhanson:
So this export is more for reporting, like the event exports. It is definitely not intended to be used to be importable back into Zenoss. Right now it is only supposed to export basically what is visible in the grid.

byshovets:
Pull request reopened because zenbatchload format is not a requirement

keruzu:
Event XML exports *can* be imported by zensendevent, and is used a lot for
regression testing, prod to staging testing etc.


Creating a homegrown XML format for 'reporting' seems to be pretty fishy,
especially when built-in functionality that does something more useful is
available.
Further, given the scale of the amount of data that can be generated for
larger systems, the Zope timeouts, this is probably better done as a job,
anyway.

On Tue, Feb 3, 2015 at 9:53 AM, Joseph Hanson <notifications@github.com>
wrote:

> Merged #716 <https://github.com/zenoss/zenoss-prodbin/pull/716>.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/zenoss/zenoss-prodbin/pull/716#event-229300183>.
>



-- 


*Kells Kearney *

Developer, OSS Tools

​kells.kearney@sungardas.com

---------------------------------------

[image: cid:part1.07070106.05070600@sungard.com] <http://www.sungardas.com/>
---------------------------------------

PR #718, opened 2015-02-04
ZEN-16381 Reports fail to load, send user to login page (zope perms issu...

romanbakaleyko:
...e)


romanbakaleyko:
port https://github.com/zenoss/4.2.5-RPS/pull/355

PR #721, opened 2015-02-05
Handle arguments through the Goto redirect and the device devault view

epowell:
Fix wanted for Impact 5.0 - please pick for RM 5.0.1.


epowell:
The GotoRedirect view does not currently honor any query string arguments. Neither does the default view for the Device in backcompat.py. These need to be updated for devices to work.
Components are next.

PR #724, opened 2015-02-05
fixes ZEN-15056 monkeypatch _findUsers to allow us to use other plugins ...

jhanson:
...when finding properties if one fails


jhanson:
I think I found a better fix for this

PR #727, opened 2015-02-06
fixes ZEN-16496: Cannot delete device from add device option.

itshane:
Zenoss.remote.JobsRouter.deleteJobs doesn't return the response.


Products/ZenUI3/browser/resources/js/zenoss/quickstart/view/AddDeviceView.js
```diff
@@ -136,9 +136,7 @@
                         Zenoss.remote.JobsRouter.deleteJobs({
                             jobids: [record.get('uuid')]
                         }, function(response) {
-                            if (response.success) {
```
jhanson: I think the best approach would be to make sure the deletejobs api returns a success flag. It should be 

return DirectResponse.succeed() 
instead of nothing 
https://github.com/zenoss/zenoss-prodbin/blob/develop/Products/Zuul/routers/jobs.py#L64

PR #750, opened 2015-02-19
ZEN-16648 Locking doesn't work for deletion devices or its components.

romanbakaleyko:


romanbakaleyko:
port from https://github.com/zenoss/pm-resmgr-4.2.4/pull/772

PR #771, opened 2015-02-24
ZEN-15658: ZenCollector workers.py log improvements (pid, name)

serge30:
https://jira.zenoss.com/browse/ZEN-15658

### DIFF
--- a/Products/ZenCollector/utils/workers.py
+++ b/Products/ZenCollector/utils/workers.py
@@ -130,7 +130,7 @@ class ProcessWorkers(object):
         """
         no_problem_children = True
         for worker in self._workers.values():
-            log.info("Stopping worker %s..." % worker)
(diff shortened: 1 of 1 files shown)


Products/ZenCollector/utils/workers.py
```diff
@@ -130,7 +130,7 @@ def _shutdownWorkersNicely(self):
         """
         no_problem_children = True
         for worker in self._workers.values():
-            log.info("Stopping worker %s..." % worker)
+            log.info("Stopping worker %s (%s)" % (worker.pid, worker.name) )
```
jpeacock-zenoss: This should be written in this form:
`log.info("My message %s %s", var1, var2)`
instead of the `%` form.

This is OK because logging can delay the interpolation until later when it's determine that it will actually write the message to the log file.

PR #779, opened 2015-02-25
Remove unnecessary migrate script

alapidas:


alapidas:
In conjunction with https://github.com/zenoss/platform-build/pull/480

PR #780, opened 2015-02-26
ZEN-15883: Catalog issues stop data collection

byshovets:
Fixing https://jira.zenoss.com/browse/ZEN-15883

### DIFF
--- a/Products/ZenCollector/services/config.py
+++ b/Products/ZenCollector/services/config.py
@@ -237,16 +237,11 @@ class CollectorConfigService(HubService, ThresholdMixin):
     def _getDevices(self, deviceNames=None, deviceFilter=None):
 
         if not deviceNames:
-            devices = filter(deviceFilter, self._prefs.devices())
-        else:
-            devices = []
-            for name in deviceNames:
-                device = self.dmd.Devices.findDeviceByIdExact(name)
-                if not device:
-                    continue
-                else:
-                    if deviceFilter(device):
-                        devices.append(device)
+            return filter(deviceFilter, self._prefs.devices())
+        devices = []
+        for device in self._prefs.devices():
+            if device.id in deviceNames:
+                devices.append(device)
         return devices
 
     @translateError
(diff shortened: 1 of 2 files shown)


Products/ZenCollector/services/config.py
```diff
@@ -237,16 +237,11 @@ def remote_getDeviceNames(self, options=None):
...
-                if not device:
-                    continue
-                else:
-                    if deviceFilter(device):
-                        devices.append(device)
+            return filter(deviceFilter, self._prefs.devices())
+        devices = []
+        for device in self._prefs.devices():
+            if device.id in deviceNames:
+                devices.append(device)
```
jplouis: The devices need to be filtered just like the  code did before, without filtering it might break collector  workers.

Products/ZenCollector/services/config.py
```diff
@@ -237,16 +237,11 @@ def remote_getDeviceNames(self, options=None):
...
-            for name in deviceNames:
-                device = self.dmd.Devices.findDeviceByIdExact(name)
-                if not device:
-                    continue
-                else:
-                    if deviceFilter(device):
-                        devices.append(device)
+            return filter(deviceFilter, self._prefs.devices())
+        devices = []
+        for device in self._prefs.devices():
```
jplouis: This is a very expensive call.  Some collectors can have thousands of devices.

Products/ZenCollector/services/config.py
```diff
@@ -237,16 +237,11 @@ def remote_getDeviceNames(self, options=None):
...
+            if device.id in deviceNames:
```
jplouis: deviceNames is a list and can be a huge list as well.  You should not scan a list for every device id

Products/ZenHub/services/ModelerService.py
```diff
@@ -30,6 +30,13 @@ def __init__(self, dmd, instance):
         PerformanceConfig.__init__(self, dmd, instance)
         self.config = self.dmd.Monitors.Performance._getOb(self.instance)
 
+    def get_devices(self, names):
+        return [x for x in self.config.devices() if x.id in names]
```
jplouis: These are very expensive calls

Products/ZenCollector/services/config.py
keruzu: "The performance of a broken system is zero"
And since the catalog might be utterly broken or completely unresponsive, it might not return EVER.  This causes data collection issues and silent failures.  Been there and had to pick up the pieces MULTIPLE times.

PR #788, opened 2015-03-03
Create Trendline thresholds 

jhanson:


Products/ZenModel/RRDTemplate.py
```diff
@@ -464,9 +464,13 @@ def getDataSourceInstance(self, id, dsOption):
...
         from Products.ZenModel.MinMaxThreshold import MinMaxThreshold
         from Products.ZenModel.ValueChangeThreshold import ValueChangeThreshold
-        thresholdClasses = [MinMaxThreshold, ValueChangeThreshold]
+        from Products.ZenModel.TrendlineThreshold import TrendlineThreshold
+        thresholdClasses = [MinMaxThreshold, ValueChangeThreshold, TrendlineThreshold]
         for zp in self.dmd.ZenPackManager.packs():
-            thresholdClasses += zp.getThresholdClasses()
+            try:
+                thresholdClasses += zp.getThresholdClasses()
+            except Exception, e:
```
jhanson: Do not stack trace if there is a zenpack we are unable to pull the thresholds from.

Products/ZenModel/TrendlineThreshold.py
```diff
@@ -0,0 +1,90 @@
...
+from ThresholdInstance import ThresholdContext
+from Products.ZenEvents.ZenEventClasses import Capacity
+
+import logging
+log = logging.getLogger('zen.TrendlineThreshold')
+
+PROJECTION_TIME_UNITS = ['days', 'weeks', 'months']
+PROJECTION_ALGORITHMS = ['linear']
+
+class TrendlineThreshold(MinMaxThreshold):
```
jhanson: Eventually this will make a datasource that stores the projection variables (e.g. "m" and "b" for mx+b) but for right now just signifies to the visualization library to render a projection.

Products/ZenModel/TrendlineThreshold.py
```diff
@@ -0,0 +1,90 @@
...
+                                      eventFields=self.getEventFields(context),
+              )
+        return mmt
+
+
+
+InitializeClass(TrendlineThreshold)
+
+
+class TrendlineThresholdInstance(MinMaxThresholdInstance):
```
jhanson: This will be filled out when we actually alert against these things

Products/ZenUI3/browser/resources/js/zenoss/JobsWidget.js
```diff
@@ -15,7 +15,10 @@ Ext.ns('Zenoss');
 var REMOTE = Zenoss.remote.JobsRouter;
 
 function lengthOrZero(o) {
-    return o===null ? 0 : o;
+    if (o) {
```
jhanson: Unrelated error that was showing "undefined" jobs remaining under some circumstances

Products/ZenUI3/browser/resources/js/zenoss/form/TimeSpan.js
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * This content is made available according to terms specified in
+ * License.zenoss under the directory where your Zenoss product is installed.
+ *
+ ****************************************************************************/
+
+
+/* package level */
+(function() {
+    Ext.define("Zenoss.form.TimeSpan", {
```
jhanson: This widget has a number and a timespan (weeks months and days). It is used for inputting the amount of data to fetch as well as the amount of time to project to see if the threshold is breached.

Products/ZenUI3/browser/resources/js/zenoss/form/graphPanel.js
```diff
@@ -18,6 +18,7 @@
     function syntaxHighlight(json) {
         json = JSON.stringify(json, undefined, 4);
         json = json.replace(/&/g, '&amp;').replace(/</g, '&lt;').replace(/>/g, '&gt;');
+        json = json.replace(/ /g, '&nbsp;');
```
jhanson: The chart definitions were hard to read because the space wasn't being respected.

Products/ZenUI3/browser/resources/js/zenoss/form/graphPanel.js
```diff
@@ -426,11 +428,6 @@
             var delta = Math.round(rangeToMilliseconds(gp.drange)/this.pan_factor);
             var newstart = gp.start + delta > 0 ? gp.start + delta : 0;
             var newend = newstart + rangeToMilliseconds(gp.drange);
-            var currTime = now();
```
jhanson: This allows the chart to go past "now"

Products/Zuul/routers/template.py
```diff
@@ -177,7 +177,7 @@ def getThresholdDetails(self, uid):
         """
         facade = self._getFacade()
         thresholdDetails = facade.getThresholdDetails(uid)
-        form = IFormBuilder(thresholdDetails).render(fieldsets=False)
```
jhanson: Previously no thresholds used fieldsets to organize its fields so it was disabled.

Products/ZenModel/TrendlineThreshold.py
```diff
@@ -0,0 +1,90 @@
...
+from Globals import InitializeClass
+from MinMaxThreshold import MinMaxThreshold, MinMaxThresholdInstance
+from ThresholdInstance import ThresholdContext
+from Products.ZenEvents.ZenEventClasses import Capacity
+
+import logging
+log = logging.getLogger('zen.TrendlineThreshold')
+
+PROJECTION_TIME_UNITS = ['days', 'weeks', 'months']
+PROJECTION_ALGORITHMS = ['linear']
```
epowell: This should eventually be a zope named utility I think.

Products/ZenUI3/browser/resources/js/zenoss/form/graphPanel.js
epowell: Shouldn't this just be a html-text-encode call? There are likely other characters than '>', '<', ' ' that would need preserving.

Products/Zuul/facades/templatefacade.py
```diff
@@ -268,7 +268,7 @@ def getThresholdDetails(self, uid):
         """
         threshold = self._getObject(uid)
         template = threshold.rrdTemplate()
-        info = IThresholdInfo(threshold)
+        info = IInfo(threshold)
```
epowell: Why? Was something incompatible with IThresholdInfo? Looking at IThresholdInfo, I think we'd want to keep using it.

You'll get an IThresholdInfo in most cases anyway, I just think we should error visibly rather than in the js silently.

Products/Zuul/interfaces/template.py
```diff
@@ -155,6 +155,35 @@ class IMinMaxThresholdInfo(IThresholdInfo):
     explanation = schema.TextLine(title=u'Explanation', order=2)
     resolution = schema.TextLine(title=u'Resolution', order=2)
 
+class ITrendlineThresholdInfo(IInfo):
```
epowell: Why not IThresholdInfo? Field ordering?

Products/ZenModel/TrendlineThreshold.py
jhanson: Yep, you are definitely right. I'll make a note of it for the next iteration.

Products/Zuul/facades/templatefacade.py
jhanson: This was tricky but basically I needed more flexiability because the "schema form builder" thing is additive. Which means I can't get different fields than the parent field without doing janky things like making certain fields hidden. 
It was cleaner to make a new interface type that descended from IInfo and added my fields there.

Products/Zuul/interfaces/template.py
jhanson: See above comment. I didn't want all the IThresholdInfo fields.

Products/ZenUI3/browser/resources/js/zenoss/form/graphPanel.js
jhanson: I still wanted " and ' characters so the syntax highlighting regex still worked and html encoding was overly aggressive.

PR #789, opened 2015-03-04
ZEN-14324 ZenPack install fails to add all device classes, if the same n...

romanbakaleyko:
...ame is used within the same hierarchy

### DIFF
--- a/Products/ZenRelations/ImportRM.py
+++ b/Products/ZenRelations/ImportRM.py
@@ -272,6 +272,8 @@ for a ZenPack.
         try:
             if id.startswith('/'):
                 obj = getObjByPath(self.app, id)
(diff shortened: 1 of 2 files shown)


Products/ZenRelations/ImportRM.py
```diff
@@ -272,6 +272,8 @@ def createObject(self, attrs):
         try:
             if id.startswith('/'):
                 obj = getObjByPath(self.app, id)
+                if obj.getPrimaryUrlPath() != id:
+                    obj = None
```
jpeacock-zenoss: It seems that if `getObjByPath` were fixed to _not_ return an object for a path that doesn't exist, then the check for the object's path would be unnecessary.

PR #791, opened 2015-03-05
ZEN-16953: There is no mechanism to update zep db schema in an upgrade

fraibacas:


jplouis:
When running an upgrade the new zenoss image will be downloaded before the upgrade command is run.  This means zep and other daemons get started first before the migrate happens.  Won't running zep with new  code before the database is updated possibly cause problems?

fraibacas:
Other alternative could be to add a run to MariaDB or Zope to upgrade the schema and call it before zeneventserver starts. The problem is that we still don't support service definition upgrades.
This was a problem for upgrading from CA2 to 5.0.1. The problem does not happen with CA3, thus we could move this fix until we support service definitions upgrades so we can add the new run.

PR #797, opened 2015-03-10
ZEN-14324 ZenPack install fails to add all device classes, if the same n...

romanbakaleyko:
...ame is used within the same hierarchy

### DIFF
--- a/Products/ZenUtils/Utils.py
+++ b/Products/ZenUtils/Utils.py
@@ -286,38 +286,19 @@ def getObjByPath(base, path, restricted=0):
                 if not validated:
                     raise Unauthorized( name )
         else:
-            if restricted:
-                next = guarded_getattr(obj, name, marker)
-            else:
-                next = _getattr(obj, name, marker)
-                ## Below this is a change from the standard traverse from zope
-                ## it allows a path to use acquisition which is not what
-                ## we want.  Our version will fail if one element of the
-                ## path doesn't exist. -EAD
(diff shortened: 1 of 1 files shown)


Products/ZenUtils/Utils.py
```diff
@@ -198,6 +198,7 @@ def getObjByPath(base, path, restricted=0):
     @return: object pointed to by the path
     @rtype: object
     """
+    opath = path
     if not path:
         return base
```
jpeacock-zenoss: Move the `opath = path` statement after the `if` statement.  Don't need to set `opath` until it's needed.

Products/ZenUtils/Utils.py
```diff
@@ -284,28 +286,19 @@ def getObjByPath(base, path, restricted=0):
...
-                    obj, obj, _none, next):
-                    raise Unauthorized( name )
+            ## Below this is a change from the standard traverse from zope
+            ## it allows a path to use acquisition which is not what
+            ## we want.  Our version will fail if one element of the
+            ## path doesn't exist. -EAD
+            #if hasattr(aq_base(obj), name):
+            #    next = _getattr(obj, name, marker)
+            #else:
+            #    raise NotFound, name
```
jpeacock-zenoss: I believe that lines 293-296 (the commented out code) is no longer necessary.  The comment itself should stay (but remove one of the `#` from each line).

jpeacock-zenoss:
Everything else looks reasonable (assuming it produces correct results).

PR #811, opened 2015-03-19
commands were using the wrong template fixes ZEN-15024

jhanson:


jhanson:
screenshot 2015-03-19 12 32 51

PR #815, opened 2015-03-27
Fix getTree call on monitor router

alapidas:


alapidas:
```queryPerformanceMonitors()``` doesn't exist anymore

File: Products/Zuul/routers/monitor.py [hunk 1]
```diff
@@ -37,7 +37,7 @@ def getTree(self, id):
         @return:  Object representing the tree
         """
         facade = self._getFacade()
-        monitors = facade.queryPerformanceMonitors()
+        monitors = facade.queryMonitors()
```
jhanson: I don't see query monitors in the facade either https://github.com/zenoss/zenoss-prodbin/blob/develop/Products/Zuul/facades/monitor.py

PR #819, opened 2015-04-08
Allow --restore to skip the erase of some zenpacks

achow-zenoss:


alapidas:
```--keep``` is kind of ambiguous here; It could mean don't delete the zenpack from disk like we normally would if it's not found in zodb, keep an older version of the pack around instead of upgrading it, keep the database objects in tact but do not re-install the zenpack, etc.

Please make the description of the option more descriptive, or extend the option's functionality to cover all the cases above (and anything else).

PR #820, opened 2015-04-09
Router changes to move devices

alapidas:
+ Allow moving a collector's worth of devices at once to another
collector

### DIFF
--- a/Products/Zuul/routers/device.py
+++ b/Products/Zuul/routers/device.py
@@ -846,6 +846,24 @@ class DeviceRouter(TreeRouter):
             log.exception(e)
             return DirectResponse.exception(e, 'Failed to change priority.')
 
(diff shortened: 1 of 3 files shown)


Products/Zuul/routers/device.py
```diff
@@ -846,6 +846,24 @@ def setPriority(self, uids, priority, hashcheck, uid=None, ranges=(),
             log.exception(e)
             return DirectResponse.exception(e, 'Failed to change priority.')
 
+    def moveCollectorDevices(self, srcCollector, dstCollector, hashcheck, uid=None, ranges=(),
+                     params=None, sort='name', dir='ASC', moveData=False,
```
wwwtyro: Can we use something other than "dir" here, so we don't shadow the builtin?

PR #821, opened 2015-04-10
ZEN-12319 removing snmp communities from add multiple devices hangs zenj...

romanbakaleyko:
...obs.


romanbakaleyko:
port https://github.com/zenoss/4.2.5-RPS/pull/422

PR #822, opened 2015-04-10
ZEN-14074 Jobs auto-refresh happening on users with no permissions to jo...

romanbakaleyko:
...bs


romanbakaleyko:
port https://github.com/zenoss/4.2.5-RPS/pull/388

PR #824, opened 2015-04-10
ZEN-16957: IPv4 networks on Network page doesn't load.

olst:
4.2.5 port: https://github.com/zenoss/4.2.5-RPS/pull/436/files

### DIFF
--- a/Products/ZenModel/UserInterfaceSettings.py
+++ b/Products/ZenModel/UserInterfaceSettings.py
@@ -34,6 +34,7 @@ class UserInterfaceSettings(ZenModelRM):
         {'id': 'showEventSeverityIcons', 'type': 'boolean', 'mode': 'w'},
         {'id': 'enableTreeFilters', 'type': 'boolean', 'mode': 'w'},
         {'id': 'showPageStatistics', 'type' : 'boolean', 'mode':'w'},
+        {'id': 'renderNetworksFromCatalog', 'type' : 'boolean', 'mode':'w'},
         {'id': 'deviceGridBufferSize', 'type': 'int', 'mode': 'w'},
         {'id': 'componentGridBufferSize', 'type': 'int', 'mode': 'w'},
         {'id': 'eventConsoleBufferSize', 'type': 'int', 'mode': 'w'},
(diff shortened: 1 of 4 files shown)


Products/ZenModel/UserInterfaceSettings.py
```diff
@@ -34,6 +34,7 @@ class UserInterfaceSettings(ZenModelRM):
         {'id': 'showEventSeverityIcons', 'type': 'boolean', 'mode': 'w'},
         {'id': 'enableTreeFilters', 'type': 'boolean', 'mode': 'w'},
         {'id': 'showPageStatistics', 'type' : 'boolean', 'mode':'w'},
+        {'id': 'renderNetworksFromCatalog', 'type' : 'boolean', 'mode':'w'},
```
jhanson: I'm not sure why this should be a user configurable option. If using the catalog is more performant then it should always be used. Under what circumstances would a user prefer one versus the other?

Products/ZenUI3/browser/resources/js/zenoss/HierarchyTreePanel.js
```diff
@@ -729,8 +729,16 @@
             // is the top level organizer (e.g. Locations, Groups)
             return orgPieces[3] === targetPieces[3];
         },
-        refresh:function (callback, scope) {
-            this.getStore().load({
+        refresh:function (callback, scope, shift, direction) {
```
jhanson: This is too high level here since it is only for the network tree. This should be in the network tree subclass.

Products/ZenUI3/browser/resources/js/zenoss/network.js
```diff
@@ -152,6 +152,13 @@ function treeselectionchange(sm, newnodes) {
     }
     var newnode = newnodes[0];
     var uid = newnode.data.uid;
+    if (uid.match(/:backward:|:forward:/)) {
```
jhanson: Use separate fields (updating the model and what the server sends) instead of overloading the uid field.

Products/Zuul/routers/network.py
```diff
@@ -114,6 +117,46 @@ def deleteNode(self, uid):
         audit('UI.Network.DeleteSubnet', subnet=uid)
         return DirectResponse.succeed(tree=self.getTree())
 
+    def _marshal(self, netid=None, shift=0, limit=None):
+        root = self.catalog.search(IpNetwork, depth=0, paths=[netid]).results.next()
```
jhanson: If we know the id of the IpNetwork that is the root then a catalog search is unnecessary.

Products/ZenModel/UserInterfaceSettings.py
byshovets: Unfortunately, I have not found the way to get search through catalog works fine -- it takes a lot of time. This option gives user a choice: whether have Network page available in case of extra-large networks with limited search, or do not change anything and have fully-functional search.

dougsyer: Sorry to invade the thread but on 4.2.5 sp 273 (plus hotfixes) the network page is unusably slow.  I read that there are performance impeocements in the latest rps havent tried that yet.

Actually would be great for us if those searches where permissioned.  We use ip realms but cant restrict access per realm or subnet now so i cant give customers access to the networks but up until now since it was slow..

the maps arent slow so I assumed it was using catalog searches for that so i again assumed the problem was that the right side is using catalog but the tree structure on the left wasnt and thats why it was slow..

PR #833, opened 2015-04-15
Using pip to uninstall zenpack package

achow-zenoss:


Products/ZenUtils/zenpack.py
```diff
@@ -755,12 +757,12 @@ def buildOptions(self):
                                action="store_true",
                                default=False,
                                help='List installed ZenPacks')
-        self.parser.add_option('--keep-packs',
+        self.parser.add_option('--keep-pack',
                                dest='keepPackNames',
```
alapidas: You may want to change this to ```keepPackName``` to be consistent

achow-zenoss: Because the option can be specified multiple times and will generate a pack name list so it should be ok here.

PR #834, opened 2015-04-16
Fix async flag on device router's setCollector

alapidas:


Products/Zuul/facades/devicefacade.py
```diff
@@ -524,12 +539,8 @@ def setCollector(self, uids, collector, moveData=False, asynchronous=True):
                 'fromCollector': info.collector,
             })
             info.collector = collector
-            notify(IndexingEvent(info._object))
```
alapidas: This is removed because an ```IndexingEvent``` happens inside the setter for the ```collector``` attribute

PR #837, opened 2015-04-20
Fix multi select in tree panels

alapidas:
Due to EXTJSIV-6824, which is fixed in the 4.2 timeframe.  This code
should be removed once we upgrade ExtJS

### DIFF
--- a/Products/ZenUI3/browser/resources/js/zenoss/ExtOverrides.js
+++ b/Products/ZenUI3/browser/resources/js/zenoss/ExtOverrides.js
@@ -649,6 +649,11 @@ Ext.override(Ext.util.Sorter, {
         }
     });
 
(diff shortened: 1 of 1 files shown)


Products/ZenUI3/browser/resources/js/zenoss/ExtOverrides.js
```diff
@@ -649,4 +649,389 @@ Ext.override(Ext.util.Sorter, {
         }
     });
```
jhanson: Add a comment stating what the patch is for please so we can review it when upgrading ExtJs

alapidas: np.

PR #840, opened 2015-04-22
ZEN-16570 Zencommand JSON parser fails silently

romanbakaleyko:


romanbakaleyko:
port https://github.com/zenoss/pm-resmgr-4.2.4/pull/781/files

PR #843, opened 2015-04-23
ZEN-15235: "Sort Column" and "Sort Sense" don't work on Custom Device Re...

serge30:
...port

Port: https://jira.zenoss.com/browse/ZEN-15235

### DIFF
--- a/Products/ZenWidgets/ZenTableManager.py
+++ b/Products/ZenWidgets/ZenTableManager.py
@@ -143,6 +143,15 @@ class ZenTableManager(SimpleItem, PropertyManager):
                                         default=default, reset=reset)
 
 
+    def deleteTableState(self, tableName):
+        """delete an existing table state"""
+        request = self.REQUEST
+        tableStates = self.getTableStates()
+
+        if tableName in tableStates:
+            del tableStates[tableName]
+
+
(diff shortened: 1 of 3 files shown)


Products/ZenWidgets/ZenTableManager.py
```diff
@@ -143,6 +143,15 @@ def setReqTableState(self, tableName, attrname, default=None, reset=False):
                                         default=default, reset=reset)
 
 
+    def deleteTableState(self, tableName):
+        """delete an existing table state"""
+        request = self.REQUEST
```
jpeacock-zenoss: Is there a reason for this line of code?

serge30: ups... no reasons, removed.

jpeacock-zenoss:
Why was the `zmanage_editProperties` (and `deleteTableState`) method added.  How does that pertain to the solution?

serge30:
Sorting parameters (field and direction) are stored in 2 places. One is the Custom Report properties which are edited on UI and stored in ZODB. The second one is table state which holds last column used for sorting. The table state has a priority over Custom Report properties. 

So after editing the report properties, new sorting parameters were not applied and this method resets table state.

PR #860, opened 2015-05-08
integrates service migration sdk with zenoss upgrade path

wwwtyro:


bin/zenrun.d/upgrade.sh
```diff
@@ -9,8 +9,14 @@
 ##############################################################################
 
 doUpgrade() {
+    SERVICE_MIGRATION_PATH=`python /opt/zenoss/bin/service-migrate.py begin` || return "$?"
```
iancmcc: Does the return work with the subshell?

PR #866, opened 2015-05-13
Adds migration script for 5.1.0 that imports zproxy to all collector …

wwwtyro:
…daemons. Tested this with zenrun upgrade.sh doUpgrade, seems to work fine. Not sure I got the versioning right, though.

### DIFF
--- a/Products/ZenModel/migrate/collectorDaemonsImportZproxy.py
+++ b/Products/ZenModel/migrate/collectorDaemonsImportZproxy.py
@@ -15,7 +15,7 @@ sm.require("1.0.0")
 class CollectorDaemonsImportZproxy(Migrate.Step):
     "Import zproxy in all collector daemons."
 
-    version = Migrate.Version(5,1,0)
+    version = Migrate.Version(5,0,70)
 
     def cutover(self, dmd):
         ctx = sm.ServiceContext()
@@ -35,6 +35,12 @@ class CollectorDaemonsImportZproxy(Migrate.Step):
 
         # Add the endpoint to each daemon.
         for daemon in collectorDaemons:
+
+            # Make sure we're not already importing zproxy.
+            zpImports = filter(lambda ep: ep.name == "zproxy" and ep.purpose == "import", daemon.endpoints)
+            if len(zpImports) > 0:
+                continue
+
             daemon.endpoints.append(endpoint)
 
         # Commit our changes.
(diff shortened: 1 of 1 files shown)


Products/ZenModel/migrate/collectorDaemonsImportZproxy.py
```diff
@@ -0,0 +1,43 @@
...
+
+import Migrate
+import servicemigration as sm
+sm.require("1.0.0")
+
+
+class CollectorDaemonsImportZproxy(Migrate.Step):
+    "Import zproxy in all collector daemons."
+
+    version = Migrate.Version(5,1,0)
```
wwwtyro: Is this versioning right?

wplu: we normally make it (5, 0, 70) on develop branch

Products/ZenModel/migrate/collectorDaemonsImportZproxy.py
```diff
@@ -0,0 +1,43 @@
...
+            purpose = "import",
+            application = "zproxy",
+            portnumber = 8080,
+            protocol = "tcp",
+            addressConfig = sm.AddressConfig(0, "")
+        )
+
+        # Add the endpoint to each daemon.
+        for daemon in collectorDaemons:
+            daemon.endpoints.append(endpoint)
```
jhanson: Do we need to make sure the endpoint isn't already in the collection first?

Products/ZenModel/migrate/collectorDaemonsImportZproxy.py
```diff
@@ -0,0 +1,43 @@
...
+
+        # Commit our changes.
+        ctx.commit()
```
jhanson: Should our context commits be contingent upon the ZODB transaction being successful? In this example we are not doing any ZODB mutation but I could see other migrate scripts where we are doing ZODB changes and Service changes that depend upon each other.

alapidas: I'd say yes.  Also, keep in mind that when you run ```zenmigrate``` all migrate scripts run in one big transaction.  If any one migrate script fails, we bail on the whole thing.  FWIW we brokw this out into separate transactions for zup installs, you could do something like this if you really wanted to (https://github.com/zenoss/pm-resmgr-4.2.4/blob/master/scripts/zup/post/run_migrates)

dougsyer: Whatever you guys do with the migration scripts whenever possible if we can get some idea that things are still processing and not hung or looping...that is super important.  if there is a big change happening things can get hung up (like if we make a mistake and have a hub still connected to zodb).  the zup actually does a good job with it but the individual zenpacks dont always...from and end user perspective, the worst scenerio is you have planned downtime of x hours and you are running straces to see if stuff is hung because you are creaping up on your change backout time.  i know this isnt probably directly relevant to what you are doing here ..

PR #872, opened 2015-05-14
ZEN-17256 Triggers using custom details cause error messages, fails t…

romanbakaleyko:
…o process trigger


romanbakaleyko:
port from 4.2.5

PR #875, opened 2015-05-14
Add a migration step for updating zentrap config files in 5.1.0.

wwwtyro:


Products/ZenModel/migrate/updateZentrapConfigs.py
```diff
@@ -0,0 +1,43 @@
...
+
+class UpdateZentrapConfigs(Migrate.Step):
+    "Add zentrap.filter.conf and alter zentrap.conf."
+
+    version = Migrate.Version(5,0,70)
+
+    def cutover(self, dmd):
+        ctx = sm.ServiceContext()
+
+        zentrap = filter(lambda s: s.name == "zentrap", ctx.services)[0]
```
jhanson: Will this work with zentraps that are on other collectors?

wwwtyro: Ah, no. I was rather simple-mindedly assuming a single collector. I'll perform this operation on all services named zentrap.

PR #891, opened 2015-05-21
ZEN-16855: Linux_ifconfig Command Plugin fails on CentOS 7

vmashkov:
Changes:
 - Updated flags 
 - Renamed interfaces names exactly as for ifconfig
 
Retested again, Screenshot https://www.dropbox.com/s/imov2b062jzuvrd/Screenshot%202015-05-22%2000.33.52.png?dl=0


dougsyer:
If you guys are revampin things here imo best to combine release and kernel into the device os poller.

This is what i replaced the default cmd modeller with for rhel/centos 5 and 6.  Could be more elegant but having distribution/release and kernel in one place is super helpful.

https://gist.github.com/dougsyer/f13aa0a86642592a7508

PR #897, opened 2015-05-28
Support ZenPack service migration.

wwwtyro:


Products/ZenModel/ZenPack.py
```diff
@@ -1201,67 +1202,54 @@ def getServiceTag(self):
...
+                services = self.getServiceDefinitionsFromFiles()
+            elif self.getDaemonNames():
+                services = self.getDefaultCollectorServiceDefinitions()
+        cpClient = ControlPlaneClient(**getConnectionSettings())
+        serviceTree = ServiceTree(cpClient.queryServices("*"))
+        ctx = servicemigration.ServiceContext()
+        for service in services:
+            parentServices = serviceTree.matchServicePath(ZenPack.currentServiceId, service["servicePath"])
+            for parentService in parentServices:
+                ctx._ServiceContext__deployService(json.dumps(service["serviceDefinition"]), parentService.id)
```
jpeacock-zenoss: curiosity: why call the '__deployService' method?  Seems like this API should be a no-underscore (public) method name.

wwwtyro: Yeah, I'm not particularly pleased with that, either. The service migration api is meant to be user facing, and only officially support a specific format for services offered by the api. This function sidesteps that, and so shouldn't be shown to the user. This was the unhappy result.

jpeacock-zenoss: Feels like that if you had to sidestep it, at some point users will feel the need as well, officially supported format or not.

Why is this 'migrated' service not in the officially supported specific format?

wwwtyro: Oh yes, we're counting on it. Our plan is to support the official interface, and people can of course muck about with the python to change whatever they want, but we can't support it.

The issue is that ZenPacks deploy their service definitions by referencing the ID of the parent service. Since we don't expose the parent ID in the service migration API, it doesn't make sense to expose a function that uses it.

Products/ZenModel/ZenPack.py
```diff
@@ -354,10 +356,9 @@ def remove(self, app, leaveObjects=False):
         if not leaveObjects:
             self.removeZProperties(app)
             self.removeCatalogedObjects(app)
-
-        # Once ZEN-16599 is fixed, make sure this exits cleanly even if
-        # the service to delete does not exist
-        self.removeServices(self.getServiceTag())
+            # Once ZEN-16599 is fixed, make sure this exits cleanly even if 
+            # the service to delete does not exist
```
jafcrocker: Can we remove this comment?

Products/ZenModel/ZenPack.py
jafcrocker: Does this even work?  Shouldn't there be a period between ServiceContext and __deployService?

jpeacock-zenoss: yes.  Python takes '__name' names and renames them to '_class__name' dynamically.  However, you're not supposed to reference the name directly in actual code.  This is the reason for the earlier discussion in this PR.

PR #908, opened 2015-06-04
Zenpack service migration support.

wwwtyro:


Products/ZenModel/ZenPack.py
```diff
@@ -1382,37 +1383,53 @@ def installServiceDefinitions(self, serviceDefs, servicePaths):
...
+            for tag in sd["Tags"]:
+                paths.append(sp + "/" + tag)
+            sd["servicePaths"] = paths
+
+        services = []
+        for sd in serviceDefs:
+            matched = False
+            for pd in serviceDefs:
+                if pd is sd:
+                    continue
```
jafcrocker: I prefer 
```
for x in collection if x is not y:
```
to 
```
for x in collection:
    if y is y: 
        continue

File: Products/ZenModel/ZenPack.py [hunk 2]
```diff
@@ -1382,37 +1383,53 @@ def installServiceDefinitions(self, serviceDefs, servicePaths):
...
+        for sd, sp in zip(serviceDefs, servicePaths):
+            sd["parentServicePath"] = sp
+
+        for sd in serviceDefs:
+            sd["Services"] = []
+            sp = sd["parentServicePath"]
+            sp = "" if sp == "/" else sp
+            paths = [sp + "/=" + sd["Name"]]
+            for tag in sd["Tags"]:
+                paths.append(sp + "/" + tag)
```
jafcrocker: I don't think this is sufficient.  We should cover the name matching ('=foo') and parent ('..') functionality. See https://github.com/zenoss/zenoss-prodbin/blob/develop/Products/ZenUtils/controlplane/servicetree.py#L86-L94

Products/ZenModel/ZenPack.py
```diff
@@ -1382,37 +1383,53 @@ def installServiceDefinitions(self, serviceDefs, servicePaths):
...
+            sd["servicePaths"] = paths
```
jafcrocker: What if, instead of adding 'servicePaths' to the service dictionary, you have a separate map: servicePath -> service.  That's really what you want.  It would make the look up code cleaner and you wouldn't have to remember to remove the entry from the service.

You could make a similar argument about the parentServicePath.  Should that be a map: service -> parentPath?

PR #911, opened 2015-06-05
ZEN-18163 When adding a device, several zProperties get locally set t…

romanbakaleyko:
…o blank values on it (only started happening in RM 5.0.3).

### DIFF
--- a/Products/ZenModel/Device.py
+++ b/Products/ZenModel/Device.py
@@ -1124,7 +1124,7 @@ class Device(ManagedEntity, Commandable, Lockable, MaintenanceWindowable,
 
         # apply any zProperties to self
         for prop, value in zProperties.items():
-            if value is not None:
+            if value is not None and value != "":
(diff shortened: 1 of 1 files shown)


Products/ZenModel/Device.py
```diff
@@ -1124,7 +1124,7 @@ def updateDevice(self,**kwargs):
 
         # apply any zProperties to self
         for prop, value in zProperties.items():
-            if value is not None:
+            if value is not None and value != "":
```
jpeacock-zenoss: If `value` is always a string (or `None`), this can be shortened to
```
    if value:
        ...
```
because `None` and empty strings are considered False in boolean expressions.

jpeacock-zenoss: The value of `value` can also be a boolean and `False` is a legitimate value for a zProperty.  The fix for ZEN-16601 is to support non-string values and the change above is to exclude two specific values that aren't generally useful for zProperties.

PR #922, opened 2015-06-16
ZEN-17256 ZEN-18175 : Allow triggers to use custom ZP details in rules

fraibacas:


fraibacas:
Zep changes: https://github.com/zenoss/zenoss-zep/pull/90

PR #923, opened 2015-06-16
Feature/zen 17516

fraibacas:


fraibacas:
zep PR: https://github.com/zenoss/zenoss-zep/pull/91

PR #928, opened 2015-06-22
ZEN-15412 Page Command can be Edited Without Password Re-Entry

romanbakaleyko:


Products/ZenModel/DataRoot.py
```diff
@@ -876,6 +876,19 @@ def _updateEmailNotifications(self, REQUEST):
...
+        if not self.dmd.ZenUsers.authenticateCredentials(curuser, curpasswd):
+            messaging.IMessageSender(self).sendToBrowser(
+                'Error',
+                'Confirmation password is empty or invalid. Please'+
+                ' confirm your password for security reasons.',
+                priority=messaging.WARNING
+            )
+            return self.callZenScreen(REQUEST)
+
+
```
jpeacock-zenoss: I think all of this should be performed only for requests from the browser, otherwise, `zendmd` scripts be blocked because it doesn't provide a _REQUEST_ object.

Products/ZenModel/DataRoot.py
```diff
@@ -876,7 +876,20 @@ def _updateEmailNotifications(self, REQUEST):
...
         """
+        
         if REQUEST:
+            curuser = self.dmd.ZenUsers.getUser().getId()
+            curpasswd = REQUEST.get('curPasswd')
+            
+            if not self.dmd.ZenUsers.authenticateCredentials(curuser, curpasswd):
+            messaging.IMessageSender(self).sendToBrowser(
+                'Error',
+                'Confirmation password is empty or invalid. Please'+
```
jpeacock-zenoss: You don't need a `+` here.  Python automatically concatenates strings found inside parenthesis.

Products/ZenModel/skins/zenmodel/editSettings.pt
```diff
@@ -210,6 +210,18 @@ loader.insert();
         </td>
     </tr>
 
+    <tr>
+        <td style="color:darkred;width:275px;height:50px" class=tableheader align="left">
+            <strong>Enter current password to confirm changes</strong>
+        </td>
+        <td align="left" style="padding:4px 0 0 4px;" class="tableheader" colspan=3>
```
jpeacock-zenoss: Styling change here.  HTML/XML specifies that all attribute values should be quoted, which includes the `colspan` attribute.  HTML is forgiving for legacy reasons which is why it didn't blow up on you.

PR #939, opened 2015-06-26
Added handling for uptime formatted as X hrs

kevinjcash:
For ZEN-18497, plus potential edge cases.

Appended to test strings to account for new check.


kevinjcash:
I apologize for the excessive whitespace removal

PR #941, opened 2015-06-30
ZEN-15407 Systemic Authorization Bypasses

romanbakaleyko:


Products/ZenModel/ZenModelRM.py
```diff
@@ -86,6 +86,7 @@ def zmanage_editProperties(self, REQUEST=None, redirect=False, audit=True):
         return ZenModelBase.zmanage_editProperties(self, REQUEST, redirect, audit)
 
 
+    security.declareProtected('Manage DMD', 'zmanage_addProperty')
```
jhanson: Make sure we are using the constants and not magic strings when declaring permissions

jhanson:
This is better than doing nothing but it doesn't close the ticket. The ticket is for the systemic problem, not just those instances. We need to do one of the following:
1. Figure out how to make sure that the default permission for methods is private 
2. Write a script that audits all the python methods and lists which ones are public
3. At zope startup time use inspection to force all methods to be defaulted to private.

PR #972, opened 2015-07-24
Core support for the ComponentGroup ZenPack

jhanson:


Products/Zuul/facades/__init__.py
```diff
@@ -261,7 +261,12 @@ def moveOrganizer(self, targetUid, organizerUid):
         parent.moveOrganizer(targetUid, [organizer.id])
         target = self._getObject(targetUid)
         # reindex all the devices under the organizer
-        for dev in parent.getSubDevices():
+        if isinstance(parent, DeviceOrganizer):
+            childObjects = parent.getSubDevices()
+        else:
```
jhuie-zenoss: can we use an elif here and check for ComponentOrganizer

PR #978, opened 2015-07-30
ZEN-15411 Denial of Service from User-Supplied Regular Expression

romanbakaleyko:


romanbakaleyko:
For me it looks like that this method should not be called directly from the web. I found only one mention in zenoss + zenpacks where we call this method
https://github.com/zenoss/zenoss-prodbin/blob/develop/Products/ZenModel/skins/zenmodel/scSortedTableGetBatch.py#L76 but nothing about where do we call scSortedTableGetBatch.py what is odd.

If current solution is not acceptable we can use third party regex module https://pypi.python.org/pypi/re2/ which is a wrapper of https://github.com/google/re2. Google's regex library is resistant to the nasty regex.

PR #985, opened 2015-08-04
ZEN-15413: Weak Password Hashing Algorithm

serge30:
https://jira.zenoss.com/browse/ZEN-15413

Time measurements of `PBKDF2-SHA256` hashing function on the XX-Large Mem (8 CPU / 61GB / 100GB / High Network) AWS box:

```
In [1]: from Products.ZenUtils import pbkdf2
In [2]: s = pbkdf2.PBKDF2DigestScheme()
In [5]: salt = s._generate_salt(24)

In [14]: %timeit -n 50 s._encrypt('password', salt, iterations=100000, hash_len=24)
50 loops, best of 3: 214 ms per loop

In [16]: %timeit -n 50 s._encrypt('password', salt, iterations=200000, hash_len=24)
50 loops, best of 3: 444 ms per loop

In [17]: %timeit -n 50 s._encrypt('password', salt, iterations=300000, hash_len=24)
50 loops, best of 3: 668 ms per loop

In [18]: %timeit -n 50 s._encrypt('password', salt, iterations=400000, hash_len=24)
50 loops, best of 3: 887 ms per loop
```

### DIFF
--- a/Products/ZenUtils/pbkdf2.py
+++ b/Products/ZenUtils/pbkdf2.py
@@ -37,10 +37,10 @@ class PBKDF2DigestScheme:
 
             <iterations>$<base64_salt>$<base64_hash>
         """
-        salt = self._generate_salt(DEFAULT_SALT_LEN)
+        salt = _generate_salt(DEFAULT_SALT_LEN)
         iterations = DEFAULT_ITERATIONS_COUNT
 
-        password_hash = self._encrypt(pw, salt, iterations, DEFAULT_HASH_LEN)
+        password_hash = _encrypt(pw, salt, iterations, DEFAULT_HASH_LEN)
 
         return '%s$%s$%s' % (iterations, base64.b64encode(salt),
                              base64.b64encode(password_hash))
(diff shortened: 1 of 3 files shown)


jpeacock-zenoss:
A security review may identify that storing the iteration count, salt, and hash in the same place as a weakness.  An attacker getting a database dump would have two pieces of unencrypted data (iteration count and salt) needed to speed up brute force attacks.

jpeacock-zenoss:
I read material on cryptographic storage and keeping the salt and hash together seems to be standard practice. There's less material on how to store the iteration count for 'slow' hashing.  Some sites argue that the iteration count should be random as well.  Many sites are non-committal on whether storing the salt and iteration count separately or in parts is worth the additional code complexity.

jpeacock-zenoss:
This site, https://www.owasp.org/index.php/Password_Storage_Cheat_Sheet, provides reasonable recommendations and this PR follows the relevant recommendations.  This more than could be done, but it's outside the scope of improving the password hashing algorithm.

jpeacock-zenoss:
However, there is one problem this PR does not address and it's a weakness of Zope's authentication design: while new passwords will benefit from the improved hashing algorithm, existing passwords still suffer from having used an older insecure algorithm and Zope does not have the framework in place to force users to reset their passwords.

Products/ZenUtils/pbkdf2.py
```diff
@@ -0,0 +1,86 @@
...
+
+        if iterations <= 0:
+            return False
+
+        attempt_hash = self._encrypt(attempt, salt, iterations,
+                                     len(password_hash))
+
+        return constant_time.bytes_eq(attempt_hash, password_hash)
+
+    def _generate_salt(self, salt_len):
```
jpeacock-zenoss: This function does not depend on any state in the class instance, and since nothing else depends on it being a member of the class, it should be module level function.

Products/ZenUtils/pbkdf2.py
```diff
@@ -0,0 +1,86 @@
...
+        """Generates random string for salt."""
+        return os.urandom(salt_len)
+
+    def _encrypt(self, password, salt, iterations, hash_len):
```
jpeacock-zenoss: This function does not depend on any state in the class instance, and since nothing else depends on it being a member of the class, it should be module level function.

PR #1002, opened 2015-08-17
Add metric name to eventKey when sending threshold related events. So…

fraibacas:
…me tranforms need it


fraibacas:
Example of transform that uses the metric name:
https://github.com/zenoss/zenoss-prodbin/blob/develop/Products/ZenModel/data/events.xml#L5227

PR #1003, opened 2015-08-17
ZEN-19198: Add metric name to eventKey when sending threshold related…

fraibacas:
… events. Some tranforms need it

### DIFF
--- a/Products/ZenUtils/metricwriter.py
+++ b/Products/ZenUtils/metricwriter.py
@@ -183,7 +183,7 @@ class ThresholdNotifier(object):
             if 'eventKey' in thresh_event_data:
                 eventKeyPrefix = [thresh_event_data['eventKey']]
             else:
-                eventKeyPrefix = [context_id]
+                eventKeyPrefix = [metric]
             for ev in self._thresholds.check(context_uuid, metric, timestamp, value):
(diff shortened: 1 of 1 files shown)


fraibacas:
Example of transform that uses the metric name:
https://github.com/zenoss/zenoss-prodbin/blob/develop/Products/ZenModel/data/events.xml#L5227

Products/ZenUtils/metricwriter.py
```diff
@@ -186,6 +186,8 @@ def notify(self, context_uuid, context_id, metric, timestamp, value, thresh_even
                 eventKeyPrefix = [context_id]
             for ev in self._thresholds.check(context_uuid, metric, timestamp, value):
                 parts = eventKeyPrefix[:]
+                if metric not in parts:   # We add the metric name to the eventKey. Some transforms use it
+                    parts.append(metric)
```
jplouis: I don't think this is correct. Look at the old code. https://github.com/zenoss/4.2.5-RPS/blob/master/src/core/Products/ZenCollector/daemon.py#L361  the datapoint name was added only eventKey wasn't present in thresh_event_data.  

I think the path rsplit in the old code is what added the datapoint name.

PR #1009, opened 2015-08-19
ZEN-16579 Physical paths exposed when sent malformed requests

ed-:
Patch catches zExceptions.Unauthorized and does not use the
zenoss_feedback_error_message template because unauthorized users
cannot access /zport/dmd resources. Instead, the generic error
template is used.

### DIFF
--- a/Products/ZenModel/DataRoot.py
+++ b/Products/ZenModel/DataRoot.py
@@ -15,6 +15,7 @@ data objects.  It can be used as a global acquisition
 name space.
 """
 
+import httplib
 import re
 from persistent.list import PersistentList
 from zope.interface import implements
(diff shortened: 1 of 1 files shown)


Products/ZenModel/DataRoot.py
```diff
@@ -411,6 +412,9 @@ def zenoss_error_message(self,error_type,error_value,
         elif isinstance(error_value, MySQLConnectionError) \
                 or isinstance(error_value, connectionFactory.exceptions.Error):
             return self.zenoss_mysql_error_message(error_value=error_value)
+        elif isinstance(error_value, Unauthorized):
+            self.REQUEST.response.status = 403
```
jhanson: use httplib.UNAUTHORIZED

PR #1013, opened 2015-08-20
ZEN-17408 preserve backwards compatablity with zenpack graph definiti…

jhanson:
…ons but still allow people to specify the dimensions of a graph


jhanson:
screenshot 2015-08-20 13 43 24

PR #1015, opened 2015-08-21
ZEN-17897: zenmib ignores 0 index in OID

fraibacas:


jplouis:
What was the decision on how this reconciles with this ticket?  https://jira.zenoss.com/browse/ZEN-3770

fraibacas:
From what I read in the ticket comments, oids containing zeros are valid oids so we need to revert what was done for ZEN-3770.

jplouis:
Please verify with Brett or William.

fraibacas:
William is ok with this approach:
"Seems good to me.  I also had Glenn and Ian take a look and they were fine with this approach."

tgjonestx:
:+1:

PR #1016, opened 2015-08-24
Don't handle container commit logic via return code.

wwwtyro:


bin/zenrun.d/zenmib.sh
```diff
@@ -9,19 +9,9 @@
 ##############################################################################
 
 __DEFAULT__() {
-    local nocommit=0
-    if echo "$@" | egrep -- '--nocommit|--keeppythoncode'; then
```
tgjonestx: The help statement at line 20 still mentions `--nocommit` - I'm guessing that help needs to be updated also.

bin/zenrun.d/zenpack-manager.sh
```diff
@@ -0,0 +1,124 @@
+#! /usr/bin/env bash   
+##############################################################################
+#
+# Copyright (C) Zenoss, Inc. 2014, all rights reserved.
```
tgjonestx: 2015

Products/ZenModel/migrate/convertRunsToCommands.py
```diff
@@ -0,0 +1,125 @@
...
+        "Command": "${ZENHOME:-/opt/zenoss}/bin/zenrun zenpack-manager.sh --service-id={{.ID}}",
+        "CommitOnSuccess": True
+    },
+    "zenpack": {
+        "Command": "${ZENHOME:-/opt/zenoss}/bin/zenrun zenpack.sh --service-id={{.ID}}",
+        "CommitOnSuccess": False
+    },
+    "install-percona": {
+        "Command": "${ZENHOME:-/opt/zenoss}/bin/zenrun percona.sh install",
+        "CommitOnSuccess": True
```
tgjonestx: Previously, the Runs for `zendmd` and `install-percona` specified `NOCOMMIT=true`, so for backwards-compatibility, I would expect the replacement Commands to use `CommitOnSuccess=False`.  So either, I'm misinterpreting a double-negative somewhere, or the old definition was wrong and needs to be changed (in which case this PR is correct), or these definitions are not backwards compatible.

tgjonestx:
Done with my comments for this one.

PR #1018, opened 2015-08-25
ZEN-18249 Zenbatchload incorrectly outputs "Modeled" instead of "Proc…

romanbakaleyko:
…essed"


romanbakaleyko:
port https://github.com/zenoss/4.2.5-RPS/pull/583

PR #1021, opened 2015-08-26
ZEN-18231: Maintenance windows with large number of devices can fail …

serge30:
…on a zodb conflict error

https://jira.zenoss.com/browse/ZEN-18231

Port from 4.2.5. Adds batching to maintenance window execution.

### DIFF
--- a/Products/ZenEvents/Schedule.py
+++ b/Products/ZenEvents/Schedule.py
@@ -162,7 +162,9 @@ class Schedule:
                     device=self.monitor,
                     prodState=prodState,
                 ))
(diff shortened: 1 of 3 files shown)


dougsyer:
By the way events for entering /exiting windows dont get a device context at least not with multirealm.  If you add the monitor and ip of the device it should work.  Also i think the agent may be set to zenobs ans not zenactiond

dougsyer:
For devices with no managed ip i just set a fake ip on the event and thay works likw vspheres

PR #1033, opened 2015-08-31
ZEN-14139: Improve getChildLinks performance

fraibacas:
New data structure add to dmd.ZenLinkManager that stores the networks a device belongs to. Havin this information ready improves the overall performance of getChildLinks.
The data structure chosen is the following:
```
        OOBTree
            Key:    Device.id
            Value:  OOBtree
                        Key:    Network
                        Value:  number of ip addresses that belong to that network
```

### DIFF
--- a/Products/ZenModel/LinkManager.py
+++ b/Products/ZenModel/LinkManager.py
@@ -16,8 +16,10 @@ from Globals import InitializeClass
 from AccessControl import ClassSecurityInfo
 
 from OFS.Folder import Folder
+from BTrees.OOBTree import OOBTree
 
 from json import dumps
+from Products.AdvancedQuery import Eq
 from Products.CMFCore.utils import getToolByName
 from Products.ZCatalog.ZCatalog import manage_addZCatalog
 from Products.ZenModel.Device import Device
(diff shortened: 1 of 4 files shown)


fraibacas:
Results:
```
In [10]: len(dmd.Networks)
Out[10]: 44560

In [11]: len(dmd.Locations.children())
Out[11]: 451

In [12]: len(dmd.Devices.Ping.devices())
Out[12]: 20947


(zenoss)[zenoss@39b544f82833 Products]$ zendmd --script=test_getChildLinks.py

--------------------------------------------------------
             getChildLinks test run 0
--------------------------------------------------------
    New implementation found 56 links
    New implementation took 21.294604063 seconds

    Old implementation found 56 links
    Old implementation took 135.238978148 seconds

    Comparing results....
    Both Implementations returned the same result!

--------------------------------------------------------
             getChildLinks test run 1
--------------------------------------------------------
    New implementation found 56 links
    New implementation took 1.57696509361 seconds

    Old implementation found 56 links
    Old implementation took 47.8688080311 seconds

    Comparing results....
    Both Implementations returned the same result!

--------------------------------------------------------
             getChildLinks test run 2
--------------------------------------------------------
    New implementation found 56 links
    New implementation took 1.5495660305 seconds

    Old implementation found 56 links
    Old implementation took 49.0893828869 seconds

    Comparing results....
    Both Implementations returned the same result!
```

Products/ZenModel/LinkManager.py
```diff
@@ -143,6 +145,63 @@ def getAddresses(self):
...
+        if device_dict:
+            network_value = device_dict.get(network_id, 0) - 1
+            if device_dict.has_key(network_id):
+                if network_value > 0:
+                    device_dict[network_id] = network_value
+                else:
+                    del device_dict[network_id]
+
+    def remove_device(self, device_id):
+        if self.cache.get(device_id):
```
iancmcc: Recommend "if device_id in self.cache". It is cheaper.

iancmcc:
This looks pretty good to me, and the results clearly speak for themselves. My only concern is a potential hot spot in the cache, but I don't see how there's any way around that, honestly, unless we make it eventually consistent.

PR #1059, opened 2015-09-14
ZEN-17811 Default zNmapPortscanOptions are invalid

romanbakaleyko:


Products/ZenModel/migrate/fixNmapArgsTemplate.py
```diff
@@ -0,0 +1,27 @@
...
+
+import Migrate
+
+
+
+
+class fixNmapArgsTemplate(Migrate.Step):
+    " Update default args for nmap command "
+
+    version = Migrate.Version(5,0,70)
```
jpeacock-zenoss: I think the version number should reference a real version.

PR #1075, opened 2015-09-21
ZEN-19130 mibs are unable to find their dependencies

romanbakaleyko:


Products/ZenModel/MibOrganizer.py
```diff
@@ -235,7 +235,7 @@ def handleUploadedFile(self, REQUEST):
         if not mypath:
             mypath = '/'
         commandArgs = [binPath('zenmib'), 'run', savedMIBPath,
-                '--path=%s' % mypath]
+                '--path=%s' % mypath, '--mibdepsdir=%s' % zenPath("/var/ext/uploadedMIBs")]
```
jpeacock-zenoss: Please make `"/var/ext/uploadedMIBs"` a module level constant.

Why pick /var/ext/uploadedMIBs as the location instead of something more Zenoss specific?

Also, since `zenmib` has to be run in a container, does that path have an export to the local filesystem or will the user need to 'upload' the MIB file into that directory in the container?

Products/ZenModel/MibOrganizer.py
```diff
@@ -227,7 +229,7 @@ def handleUploadedFile(self, REQUEST):
         """
         filename = REQUEST.upload.filename
         mibs = REQUEST.upload.read()
-        savedMIBPath = zenPath('/var/ext/uploadedMIBs/%s' % filename)
+        savedMIBPath = zenPath(_pathToMIB'/%s' % filename)
```
jpeacock-zenoss: This is a syntax error, but you should be able to write the `zenPath` call like this:
```
savedMIBPath = zenPath(_pathToMIB, filename)
```
`zenPath` behaves like `os.path.join` with regard to multiple arguments.

PR #1087, opened 2015-09-25
ZEN-17121: Custom command with streaming output renders an entire Zop…

serge30:
…e instance inaccessible

https://jira.zenoss.com/browse/ZEN-17121

Migration for Zope's threads count resetting to default value.


jplouis:
did you test the migration?  https://sites.google.com/a/zenoss.com/engineering/home/faq/work-with-rm/howtorunzenmigrate 

Otherwise it looks good

serge30:
@jplouis, yes, I tested the migration.

PR #1088, opened 2015-09-25
Raise appropriate error in clientFinished

ed-:
Fixes ZEN-19186


Products/DataCollector/SshClient.py
```diff
@@ -28,6 +28,7 @@
 from twisted.conch.ssh import common, channel
 from twisted.conch.ssh.keys import Key
 from twisted.internet import defer, reactor
+from twisted.python.failure import Failure
```
ed-: don't actually need that

PR #1089, opened 2015-09-25
revert change for ZEN-18147

jplouis:


jplouis:
ZEN-18147 is a bad fix and causes breakage with the new session manager. ZEN-18147 Should no longer be an issue with the new session manager since we now have only 1 session manager, previously there were 2 session managers and sessions weren't being cleared causing the user to be able to login with the old credentials.

PR #1097, opened 2015-09-30
Check preconditions in patch.sh before using quilt

ed-:
Adds a few clauses zenrun's patch.sh to check that the patch and series
exist, and that there are any patches to rewind and then apply.

This should prevent apply-custom-patches from erroring out with the cryptic
message "No series file found".

### DIFF
--- a/bin/zenrun.d/patch.sh
+++ b/bin/zenrun.d/patch.sh
@@ -12,33 +12,48 @@ applyPatches() {
     
     # No need to subshell here for 'cd' because this whole process needs to 
     # happen in $ZENHOME
+    local RC
 
     cd $ZENHOME
-    if ! which quilt > /dev/null 2>&1; then
+    which quilt > /dev/null 2>&1; RC=${PIPESTATUS[0]}
+    if [[ $RC != 0 ]]; then
        echo "Unable to apply custom patches - quilt is not installed" 
        return 1
     fi
     
     # Run 'quilt upgrade' to make sure things are up to date
-    quilt upgrade > /dev/null 2>&1
-    RC=$?
+    quilt upgrade > /dev/null 2>&1; RC=${PIPESTATUS[0]}
     if [[ $RC != 0 ]]; then
         echo "Failed to run 'quilt upgrade', exiting"
         return "$RC"
     fi
 
-    # First pop everything off (the image may or may not actually have the
-    # patches applied, so need to pop them first)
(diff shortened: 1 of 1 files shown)


bin/zenrun.d/patch.sh
```diff
@@ -27,15 +27,33 @@ applyPatches() {
         return "$RC"
     fi
 
-    # First pop everything off (the image may or may not actually have the
-    # patches applied, so need to pop them first)
-    quilt pop -a
+    # Assure for patch directory and series file exist.
```
ed-: VERY USE OF ENGLISH ED

bin/zenrun.d/patch.sh
```diff
@@ -27,15 +27,33 @@ applyPatches() {
...
     fi
-    
+
+    # Pop off old patches, if any
+    quilt top > /dev/null 2>&1
+    RC=$?
+    if [[ $RC == 0 ]]; then
+        quilt pop -a
+        RC=$?
+        if [[ $RC != ]];then
```
ed-: uhh

bin/zenrun.d/patch.sh
```diff
@@ -27,15 +27,33 @@ applyPatches() {
...
+    # Assure that patch directory and series file exist.
+    if [[ ! -d patches ]]; then
+        mkdir patches
+    fi
+    touch patches/series
+
+    # Check for new patches
+    quilt unapplied > /dev/null 2>&1
     RC=$?
     if [[ $RC != 0 ]]; then
```
alapidas: Technically this gives us the status if the redirection, not the ```quilt unapplied``` command.  This issue exists in other places in this script as well.

bin/zenrun.d/patch.sh
```diff
@@ -27,15 +27,33 @@ applyPatches() {
...
         return "$RC"
     fi
-    
+
+    # Pop off old patches, if any
+    quilt top > /dev/null 2>&1
+    RC=$?
+    if [[ $RC == 0 ]]; then
+        quilt pop -a
+        RC=$?
```
alapidas: You can just say ```if ! quilt pop -a;``` here

bin/zenrun.d/patch.sh
ed-: Of course it does. I'll take care of it.

PR #1098, opened 2015-10-01
Incorporate vSphere modeling optimization.

cluther:
This change incorporates the essence of the modeling optimization
currently being monkeypatched in by the vSphere ZenPack.

The optimization is first to wrap the application of all datamaps in a
single transaction instead of one transaction per datamap. This can
potentially reduce the number of invalidations generated by modeling a
large device.

The second part of the optimization is to wrap all datamap application
in a pausedAndOptimizedIndexing context manager to reduce roundtrips to
CatalogService. This can greatly reduce the time spent inside
ApplyDataMap when modeling a large device.

The final piece of this change is adding ProductClass to the list of
classes that are ignored by invalidation processing. This is mostly
unrelated to the optimization other than I saw a large number
HardwareClass and SoftwareClass objects being invalidated when modeling
UCS devices. These objects are global and only linked to devices through
their instances relationships, so there's no need for collectors to know
they've changed.

Fixes ZEN-19701.

### DIFF
--- a/Products/DataCollector/ApplyDataMap.py
+++ b/Products/DataCollector/ApplyDataMap.py
@@ -112,9 +112,20 @@ class ApplyDataMap(object):
         if deviceClass and device.getDeviceClassPath().startswith(CLASSIFIER_CLASS):
             device.changeDeviceClass(deviceClass)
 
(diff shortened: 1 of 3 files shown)


dougsyer:
Sweet

PR #1104, opened 2015-10-02
Update redis and collectorredis healthchecks

ed-:
Backgrounded use of redis-mon in healthchecks so it still ships
its metrics data, but also added the use of redis-cli ping to make
sure a response from redis is what's being tested for.

Closes ZEN-20103

### DIFF
--- a/bin/healthchecks/collectorredis/metrics
+++ b/bin/healthchecks/collectorredis/metrics
@@ -8,5 +8,6 @@
 #
 ##############################################################################
 
-/opt/zenoss/bin/redis-mon post
-/opt/zenoss/bin/redis-mon post-length --list metrics
+(timeout 10s /opt/zenoss/bin/redis-mon post) &
+(timeout 10s /opt/zenoss/bin/redis-mon post-length --list metrics) &
+redis-cli ping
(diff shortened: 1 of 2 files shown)


bin/healthchecks/collectorredis/metrics
```diff
@@ -8,5 +8,6 @@
 #
 ##############################################################################
 
-/opt/zenoss/bin/redis-mon post
-/opt/zenoss/bin/redis-mon post-length --list metrics
+(timeout 10s /opt/zenoss/bin/redis-mon post) &
+(timeout 10s /opt/zenoss/bin/redis-mon post-length --list metrics) &
+redis-cli ping
```
jplouis: how long is redis-mon taking to timeout?

ed-: In excess of the 15 seconds defined in the servicedefs, anyway.

jplouis: Won't this lead to having many active processes running in the container?

ed-: That could happen if the timeouts here are longer than the healthcheck interval. I haven't seen the process count get out of hand in my testing.

PR #1106, opened 2015-10-05
Feature/zen 20150

jplouis:


Products/ZenUtils/ZenDaemon.py
```diff
@@ -186,7 +187,7 @@ def setupLogging(self):
                 rootLog.addHandler(consoleHandler)
 
             for handler in (h for h in rootLog.handlers
-                    if isinstance(h, logging.StreamHandler)):
+                            if isinstance(h, logging.StreamHandler)):
```
smousa: was this extra tab intentional?

PR #1107, opened 2015-10-05
ZEN-20119 Zope upgrade CommitOnSuccess=True

ed-:
Adds a migration to set CommitOnSuccess to True by default for
Zope's "upgrade" command.

Fixes ZEN-20119

### DIFF
--- a/Products/ZenModel/migrate/enableCommitonUpgrade.py
+++ b/Products/ZenModel/migrate/enableCommitonUpgrade.py
@@ -1,0 +1,43 @@
+##############################################################################
+#
+# Copyright (C) Zenoss, Inc. 2015, all rights reserved.
+#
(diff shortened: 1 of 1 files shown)


Products/ZenModel/migrate/enableCommitonUpgrade.py
```diff
@@ -0,0 +1,41 @@
...
+log = logging.getLogger("zen.migrate")
+
+import Migrate
+import servicemigration as sm
+sm.require("1.0.0")
+
+
+class EnableCommitOnUpgrade(Migrate.Step):
+
+    version = Migrate.Version(5, 0, 70)
```
alapidas: This should be {{5.1.70}}

PR #1143, opened 2015-10-20
Two fixes for events on infrastructure page.

cluther:
See commit messages for more detail. The changes in the two commits are
related, but independent.

### DIFF
--- a/Products/Zuul/routers/zep.py
+++ b/Products/Zuul/routers/zep.py
@@ -323,12 +323,21 @@ class EventsRouter(DirectRouter):
             return self.queryArchive(limit=limit, start=start, sort=sort,
                                      dir=dir, params=params, exclusion_filter=exclusion_filter, keys=keys, uid=uid,
                                      detailFormat=detailFormat)
-        # special case for dmd/Devices in which case we want to show all events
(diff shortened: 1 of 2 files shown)


Products/Zuul/routers/zep.py
```diff
@@ -323,12 +323,21 @@ def query(self, limit=0, start=0, sort='lastTime', dir='desc', params=None, excl
...
+        # organizers children.
+        uids = {
+            '/zport/dmd/Devices': child_uids(self.context.dmd.Devices),
+            '/zport/dmd/Locations': child_uids(self.context.dmd.Locations),
+            '/zport/dmd/Groups': child_uids(self.context.dmd.Groups),
+            '/zport/dmd/Systems': child_uids(self.context.dmd.Systems),
+            }.get(uid, [uid])
 
         exclude_params = self._filterParser.findExclusionParams(params)
         if len(exclude_params) > 0:
```
dougsyer: Will this work okay if a user doesnt have a global role?

cluther: Yes, event permission filtering is still in effect. This change just extends the same thing we we doing for device classes to the other device organized types.

PR #1166, opened 2015-10-26
ZEN-19669: Update AddPrefixToTriggerRule migrate version

vikt0rs:
fast-adn-dirty fix to get correct Zenoss version after upgrade


vikt0rs:
Close, because the issue is invalid

PR #1168, opened 2015-10-26
ZEN-17544: Zenbatchdump fails on any 5.x deployment where there are c…

olst:
…ollectors


olst:
line 648: in case of hasattr(). What it would be if obj._isLocalHost == False?
line 652: I need default non-empty value here.

vikt0rs:
agreed, no objections

PR #1175, opened 2015-10-28
fixes ZEN-20740: False "Success" when adding Infrastructure.

itshane:


alapidas:
This has unnecessary variable renames - notably `snmpConf`

itshane:
It improves code readability, aligned with the style we have accross the module and doesn't cause any regressions.

PR #1185, opened 2015-10-30
ZEN-20099: Re-define propertyItems() to mask passwords

vikt0rs:


Products/ZenModel/Device.py
```diff
@@ -2235,4 +2236,22 @@ def getMacAddressCache(self):
...
+    security.declareProtected(access_contents_information, 'propertyItems')
+    def propertyItems(self):
+        """Return a list of (id, property) tuples. If property id related
+           to password, it will mask property value with '***'
+        """
+        result = []
+        for prop in self._properties:
+            prop_id = prop['id']
+            if prop['type'] == 'password':
+                prop_value = "***"
```
cluther: I'd recommend searching for the callers of `propertyItems`. It appears to be used in the ZenCollector collector framework to get properties needed for collection. So returning asterisks instead of the proper password seems like it would break all sorts of things.

vikt0rs: ok, will investigate this

vikt0rs: well, unfortunately I have a lack of ideas to fix it. The only suggestion is to add an optional parameter to the `propertyItems()` function to pass it, if we need password in function output and mask password otherwise.

vikt0rs:
should be fixed in another way

vikt0rs:
An another attemp to fis is https://github.com/zenoss/zenoss-prodbin/pull/1246 . @cluther could you please review it?

PR #1195, opened 2015-11-03
ZEN-19386 Heartbeat events showing up for devices that are not collec…

romanbakaleyko:
…tors.


romanbakaleyko:
port of https://github.com/zenoss/4.2.5-RPS/pull/607

PR #1198, opened 2015-11-03
ZEN-20795 Maintenance Events loses device context when using MultiRea…

romanbakaleyko:
…lmIP zenpack

### DIFF
--- a/Products/ZenEvents/Schedule.py
+++ b/Products/ZenEvents/Schedule.py
@@ -161,6 +161,7 @@ class Schedule:
                     maintenance_devices=devices,
                     device=self.monitor,
                     prodState=prodState,
+                    monitor=self.monitor
                 ))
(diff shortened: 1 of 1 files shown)


romanbakaleyko:
port https://github.com/Zenoss/4.2.5-RPS/commit/7d99995928109889d66ed3c7ee43eea52e1e3cc8

Products/ZenEvents/Schedule.py
```diff
@@ -161,6 +161,7 @@ def runEvents(self):
                     maintenance_devices=devices,
                     device=self.monitor,
                     prodState=prodState,
+                    monitor=self.monitor
```
dougsyer: Pretty sure vspheres need an ip fgield i used a fake ome because they dont have real manage ips unless they are stamd alone hosts

PR #1205, opened 2015-11-04
fixes ZEN-20519: Added hostname validation utility.

itshane:


alapidas:
Is there an immediate need for this Oleksandr?  If not, let's wait to merge something like this until it's needed.

itshane:
It will be used here: https://github.com/zenoss/ZenPacks.zenoss.CiscoUCS/pull/321/files

alapidas:
Closing per discussion

PR #1211, opened 2015-11-05
ZEN-20962 Remove empty graphs from zenmodeler

ed-:
Modifies MonitoringProfile of zenmodeler services to remove
a number of graphs which were empty anyway.

Closes ZEN-20962


tgjonestx:
Remember to cherry-pick to support/5.1.x

PR #1213, opened 2015-11-09
fixes ZEN-20859: ImportError: No module named servicemigration when d…

HIMix:
…oing a 5.0.8 upgrade


jpeacock-zenoss:
Almost.  Now this needs to be wrapped into a migration script.

PR #1234, opened 2015-11-16
ZEN-18802 Many popup dialogs reference absolute_url instead of absolu…

romanbakaleyko:
…te_url_path

### DIFF
--- a/Products/ZenModel/MibOrganizer.py
+++ b/Products/ZenModel/MibOrganizer.py
@@ -35,7 +35,7 @@ def manage_addMibOrganizer(context, id, REQUEST = None):
     context._setObject(id, sc)
     sc = context._getOb(id)
     if REQUEST is not None:
-        REQUEST['RESPONSE'].redirect(context.absolute_url() + '/manage_main')
+        REQUEST['RESPONSE'].redirect(context.absolute_url_path() + '/manage_main')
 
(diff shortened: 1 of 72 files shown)


Products/ZenRelations/PrimaryPathObjectManager.py
```diff
@@ -65,7 +65,7 @@ def getPrimaryId(self, fromNode=None):
     def getPrimaryUrlPath(self, full=False):
         """Return the primary path as an absolute url"""
         objaq = self.primaryAq()
-        if full: return objaq.absolute_url()
+        if full: return objaq.absolute_url_path()
         return objaq.absolute_url_path()
```
jpeacock-zenoss: I think, in this case, the existing code is fine.  The `absolute_url()` method is only called if the `full` parameter is passed a `True` argument.

Products/ZenModel/MibOrganizer.py
```diff
@@ -233,7 +233,7 @@ def handleUploadedFile(self, REQUEST):
         atomicWrite(savedMIBPath, mibs, raiseException=True, createDir=True)
 
         # create the job
-        mypath = self.absolute_url_path().replace('/zport/dmd/Mibs', '')
+        mypath = self.absolute_url_path_path().replace('/zport/dmd/Mibs', '')
```
jpeacock-zenoss: oops?

PR #1241, opened 2015-11-18
Adds unittests for zminionViaSupervisord migration

ed-:
As a result of testing, a fix to zminionViaSupervisord was found.


Products/ZenModel/migrate/Migrate.py
```diff
@@ -414,7 +414,7 @@ def orderedSteps(self):
     def list(self):
         print ' Ver      Name        Description'
         print '-----+---------------+-----------' + '-'*40
-        for s in self.allSteps:
+        for s in self.determineSteps():
```
tgjonestx: This looks suspect to me. 
If we are unit-testing a service migration, why do we need this change?

tgjonestx:
A couple of general comments:
 1. We're going to have lots of different service migrations (at least a dozen for RM 5.1.1), so can we have the unit-tests for each kind of service migration in a separate file to avoid lots of merge conflicts.
 1. For each kind of service migration, we need a unit-test to verify idempotency so we need to inject different input files for different test cases.

Products/ZenModel/migrate/Migrate.py
ed-: That's my fault, list() always returns the full list regardless of selected steps and that bothered me. I can remove that from this PR.

ed-:
Sure, I can add that in easy enough.

Products/ZenModel/tests/testXmlEdges.py
```diff
@@ -21,7 +21,7 @@ def testXMLEdges(self):
         """
         Test that checks correct convertion of edges to an XML file
         """
-        import pdb; pdb.set_trace()
+        #import pdb; pdb.set_trace()
```
tgjonestx: Is this related to zminion service migration?

ed-: It's not, but runtests won't finish without manually exiting from pdb with this line intact.
I'll take this change out of this PR for clarity.

Products/ZenModel/migrate/tests/common.py
```diff
@@ -0,0 +1,136 @@
...
+    migration_class_name = ''
+
+    def setUp(self):
+        pass
+
+    def tearDown(self):
+        pass
+
+    def test_cutover_correctness(self):
+        context = fakeContextFromFile(self.initial_servicedef)
```
tgjonestx: A little copy/paste code here which isn't so bad, but I think the differences btwn the two tests would be more clear if they each shared a common method. Something like
```
def test_cutover_correctness(self):
    verify_cutover(self.initial_servicedef, self.expected_servicedef)

 def test_cutover_idempotence(self):
     verify_cutover(self.expected_servicedef, self.expected_servicedef)

```

PR #1249, opened 2015-11-20
Solr Global Catalog

fraibacas:


Products/Zuul/catalog/events.py
```diff
@@ -135,6 +139,27 @@ def onTreeSpanningComponentBeforeDelete(ob, event):
             oldpaths = devicePathsFromComponent(component)
             catalog.unindex_object_from_paths(device, oldpaths)
 
+@adapter(ITreeSpanningComponent, IObjectRemovedEvent)
+def onTreeSpanningComponentAfterDelete(ob, event):
```
iancmcc: Should this be BeforeDelete?

Products/Zuul/catalog/model_catalog.py
```diff
@@ -0,0 +1,179 @@
...
+
+
+class ModelCatalogIndexable(object):
+
+    implements(IModelCatalog)
+
+    def __init__(self, wrapped):
+        #@TODO HACK I am getting issues with acqusition
+        try:
+            self.wrapped = wrapped.primaryAq()
```
iancmcc: This is actually quite expensive. We should trust the wrapped passed in, if at all possible. primaryAq() slows it way down.

Products/Zuul/catalog/model_catalog_tool.py
```diff
@@ -0,0 +1,317 @@
...
+        if types_query:
+            partial_queries.append(types_query)
+
+        # Build query for paths
+        if not paths:
+            paths = ('/'.join(self.context.getPhysicalPath()) + '*', )
+        elif isinstance(paths, basestring):
+            paths = (paths,)
+
+        """  OLD CODE. Why this instead of In?  What do we need depth for?
```
iancmcc: I'm not sure we currently use depth, although I can't remember. It will search at a specific depth in the path, e.g. "Get anything 3 levels below /zport/dmd/Devices named Server".

PR #1251, opened 2015-11-23
ZEN-11138: Cache getZenossRevision() function call

vikt0rs:


Products/ZenModel/ZenossInfo.py
```diff
@@ -279,6 +281,7 @@ def getZopeVersion(self):
         major, minor, micro, status, release = version.getZopeVersion()
         return Version(name, major, minor, micro)
 
+    @FunctionCache('getZenossRevision')
```
jplouis: 5.x doesn't run on SVN, this cmd will never work.  The method should either be removed or always return an empty string

vikt0rs: heh, that it.

@jplouis, how do you think - is it a sense to replace `svn info` to proper git command?

jplouis: It will never work on a customer site, even with git. I'm not sure it is worth the effort to print out the revision.

vikt0rs: It sounds like we should remove this method at all. @jplouis please confirm it and I'll create an another pull request

vikt0rs: btw, what about pkg_resources ?

jplouis: Removing the method is fine, you just have to find all uses of the method.

I'm not sure I understand the concern with pkg_resources

vikt0rs: oh, I mean that we can get Zenoss version using pkg_resources. For example

```
>>> import pkg_resources
>>>
>>> pkg_resources.get_distribution("Zenoss").version
'5.0.70'
>>>
>>> pkg_resources.get_distribution("Zenoss").parsed_version
('00000005', '00000000', '00000070', '*final')
>>>
```

vikt0rs: but anyway it looks reasonable for me to remove this method at all. Will make a new pull request.

vikt0rs:
Closed in favour of https://github.com/zenoss/zenoss-prodbin/pull/1258 @jplouis please review a new pull request

PR #1268, opened 2015-11-25
Added SNMP v3 args to the DataSource command execution

rphillipsz:
v3 arguments were missing from the snmp commands.  When snmp is configured for v3, the required arguments are added to the command (security name, level, security types).

### DIFF
--- a/Products/ZenModel/BasicDataSource.py
+++ b/Products/ZenModel/BasicDataSource.py
@@ -38,6 +38,39 @@ def checkOid(oid):
     return oid
 
 
+class SnmpCommand(object):
+    '''
+    Builds the command for SNMP.i  v3 has additional arguments
+    while v1/v2 just use the string template.
+    '''
+    def __init__(self, snmpinfo):
+        self.snmpinfo = snmpinfo
+
+    def getCommand(self):
+        command = snmptemplate % self.snmpinfo
+
+        if self.snmpinfo['zSnmpVer'] != 'v3':
+            return (command, command)
+
+        # v3 always requires the username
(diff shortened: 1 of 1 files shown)


Products/ZenModel/BasicDataSource.py
```diff
@@ -38,6 +38,39 @@ def checkOid(oid):
...
+        command += (" -u%(zSnmpSecurityName)s" % self.snmpinfo)
+        display = command
+
+        if self.snmpinfo['zSnmpPrivType'] and self.snmpinfo['zSnmpAuthType']:
+            display += (" -l authPriv -a %(zSnmpAuthType)s " % self.snmpinfo) + "-A ${zSnmpAuthPassword} " + \
+                ("-x %(zSnmpPrivType)s " % self.snmpinfo) + "-X ${zSnmpPrivPassword}"
+            command += (" -l authPriv -a %(zSnmpAuthType)s -A %(zSnmpAuthPassword)s "
+                "-x %(zSnmpPrivType)s -X %(zSnmpPrivPassword)s" % self.snmpinfo)
+        elif self.snmpinfo['zSnmpAuthType']:
+            display += (" -l authNoPriv -a %(zSnmpAuthType)s " % self.snmpinfo) + "-A ${zSnmpAuthPassword}"
```
jplouis: Should the passwords be show as "*****" instead of "${zSnmpAuthPassword}"

rphillipsz: I was replicating the output of the Command->smptwalk output (which is also
broken for v3), but can change it to ***** if that's better?  I'm
indifferent on which to use

PR #1279, opened 2015-11-30
MW behavior when editing them properly starts/ends the MW based on th…

rphillipsz:
…e adjusted values.

https://jira.zenoss.com/browse/ZEN-19258

Device now goes into/leaves maintenance as the dates are adjusted in and out of range (or enabled/disabled).

### DIFF
--- a/Products/ZenModel/MaintenanceWindow.py
+++ b/Products/ZenModel/MaintenanceWindow.py
@@ -283,8 +283,15 @@ class MaintenanceWindow(ZenModelRM):
             self.stopProductionState = stopProductionState
             self.skip = skip
             now = time.time()
-            if self.started and self.nextEvent(now) < now:
-                self.end()
+            if self.started:
+                if ((t + duration * 60) < now) or (t > now) or (not self.enabled):
+                    # We're running. If we should have already ended OR the start was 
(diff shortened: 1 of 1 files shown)


Products/ZenModel/MaintenanceWindow.py
```diff
@@ -283,8 +283,15 @@ def makeInt(v, fieldName, minv=None, maxv=None, acceptBlanks=True):
...
             self.skip = skip
             now = time.time()
-            if self.started and self.nextEvent(now) < now:
-                self.end()
+            if self.started:
+                if ((t + duration * 60) < now) or (t > now) or (not self.enabled):
+                    # We're running. If we should have already ended OR the start was 
+                    # moved into the future OR the MW is now disabled, end().
+                    self.end()
+            elif (t < now) and ((t + duration * 60) > now) and (self.enabled):
```
smousa: does this elif need to be indented?

rphillipsz: No, the indention is correct.  It's the case when the maintenance window
hasn't started but has been rescheduled to be active now.

rphillipsz:
Just realized I never did the backport to 5.1.x:
https://github.com/zenoss/zenoss-prodbin/pull/1311

PR #1284, opened 2015-12-02
Added test for zenmodelerWorkers

bwf93:


Products/ZenModel/migrate/tests/common.py
```diff
@@ -101,7 +109,7 @@ def _test_cutover(self, svcdef_before, svcdef_after):
         sm_context = '%s.sm.ServiceContext' % module_name
         migration = importlib.import_module(module_name)
         with mock.patch(sm_context, new=lambda: context):
-            getattr(migration, self.migration_class_name)().cutover(None)
+            getattr(migration, self.migration_class_name)().cutover(FakeDmd())
```
jafcrocker: We need to figure out a way to get a real dmd passed in for those migrations that may require them.

PR #1286, opened 2015-12-02
Hide/Disable UI elements for jobs/mibs if they don't have access.  Di…

rphillipsz:
…sable edit/add/delete on backend routers to prevent changes without permission.

### DIFF
--- a/Products/ZenUI3/browser/resources/js/zenoss/MIBs.js
+++ b/Products/ZenUI3/browser/resources/js/zenoss/MIBs.js
@@ -548,10 +548,10 @@ treesm = new Zenoss.TreeSelectionModel({
                 if (newnode.data.uid === '/zport/dmd/Mibs') {
                     isRoot = true;
                 }
-                Ext.getCmp('add-organizer-button').setDisabled(newnode.data.leaf);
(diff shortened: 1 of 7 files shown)


Products/ZenUI3/browser/resources/js/zenoss/MIBs.js
```diff
@@ -548,10 +548,10 @@ treesm = new Zenoss.TreeSelectionModel({
                 if (newnode.data.uid === '/zport/dmd/Mibs') {
                     isRoot = true;
                 }
-                Ext.getCmp('add-organizer-button').setDisabled(newnode.data.leaf);
-                Ext.getCmp('edit-mib-action').setDisabled(!newnode.data.leaf);
+                Ext.getCmp('add-organizer-button').setDisabled(newnode.data.leaf || Zenoss.Security.doesNotHavePermission('Manage DMD'));
+                Ext.getCmp('edit-mib-action').setDisabled(!newnode.data.leaf || Zenoss.Security.doesNotHavePermission('Manage DMD'));
```
jpeacock-zenoss: while `||` is available, canonical Python would use `or`.

PR #1316, opened 2015-12-07
Transactional model catalog

fraibacas:


Products/Zuul/catalog/indexable.py
```diff
@@ -0,0 +1,457 @@
...
+        if self.lanId() and self.lanId()!="None":
+            lan_id = self.lanId()
+        return lan_id
+
+    @indexed(StringFieldType(stored=True), attr_query_name="macaddress")
+    def idx_macaddress(self):
+        return self.macaddress
+
+
+class IpAddressIndexable():  # IpAddress inherits from this class
```
iancmcc: Add object as a base class, even for mixins. No old-style classes.

PR #1323, opened 2015-12-07
ZEN-21181 zendmd version is wrong

ed-:
Minor updates to ZenModel/data xml and zodb dump to fix zendmd version.
Previously it was reporting version 5.11.12; it should now report 5.1.1.


tgjonestx:
Where's the version number change in the diff? I couldn't find it.

ed-:
In the zodb.sql.gz

PR #1343, opened 2015-12-10
ZEN-20018 Run zeneventserver-create-db in upgrade.sh

ed-:


bin/zenrun.d/upgrade.sh
```diff
@@ -17,6 +17,16 @@ doUpgrade() {
...
+    zengc=$ZENHOME/bin/zenglobalconf
+    dbtype=$(${zengc} -p zodb-db-type)
+    host=$(${zengc} -p zodb-host)
+    port=$(${zengc} -p zodb-port)
+    user=$(${zengc} -p zodb-user)
+    userpass=$(${zengc} -p zodb-password)
+    /opt/zenoss/bin/zeneventserver-create-db \
+        --dbtype $dbtype --dbhost $host --dbport $port \
+        --dbuser $user --dbpass "${userpass}" \
+        --update_schema_only
```
alapidas: Can you please `|| return "$?"` here as well?

alapidas:
I _think_ we want to maybe run this for all databases, not just zep

alapidas:
Does `zodb_session` actually exist in 5.1?  I think @jplouis replaced that with a memcache-based solution

ed-:
It was in my develop environment; scratching and rebuilding to get a dead reckoning.

PR #1348, opened 2015-12-14
updated database dump with zen manager permissions

bwf93:


Products/ZenModel/data/menus.xml
```diff
@@ -4812,42 +4812,6 @@ False
...
-<property type="boolean" id="isdialog" mode="w" >
-True
-</property>
-<property type="float" id="ordering" mode="w" >
-90.5
-</property>
-</object>
-</tomanycont>
-</object>
 <object id='Templates' module='Products.ZenModel.ZenMenu' class='ZenMenu'>
```
jhuie-zenoss: why are these removed? Need to find out if this is correct.

PR #1351, opened 2015-12-15
fixes ZEN-20808: Fix for ZEN-17576 needs to be made configurable in d…

HIMix:
…aemon config files.

### DIFF
--- a/Products/ZenHub/PBDaemon.py
+++ b/Products/ZenHub/PBDaemon.py
@@ -1100,11 +1100,11 @@ class PBDaemon(ZenDaemon, pb.Referenceable):
                                type='int',
                                help='How often to ping zenhub')
 
-        self.parser.add_option('--ping-perspective',
+        self.parser.add_option('--disable-ping-perspective',
                                dest='pingPerspective',
                                help="Enable or disable ping perspective",
                                default=True,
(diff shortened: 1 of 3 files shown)


Products/ZenHub/PBDaemon.py
```diff
@@ -1100,6 +1100,12 @@ def buildOptions(self):
                                type='int',
                                help='How often to ping zenhub')
 
+        self.parser.add_option('--ping-perspective',
+                               dest='pingPerspective',
+                               help="Enable or disable ping perspective",
+                               default=True,
+                               action='store_true')
```
jpeacock-zenoss: This option configuration fails with an error if I try to specify `--ping-perspective=False` on the command line.

jplouis:
This will need a migrate for all collector daemons if you change the option.

jplouis:
There needs to be a migration script so that existing/migrated installs will have the correct options in the conf file.

alapidas:
The migrate script in this PR is incorrect.  It will change config files in the container.  The config files are actually defined in the service definition and then injected into each service.  You need to use the service migration SDK to make these changes on the service definition itself, not in the container.  There are examples that can be followed in the `ZenModel/migrate` folder.

PR #1353, opened 2015-12-15
fix ZEN-21338 saving event listed filtered with 'git status' as csv

rDr4g0n:
fixes https://jira.zenoss.com/browse/ZEN-21338

When the event query code is hit from the export button in the ui, the arguments were passed in as json which was not properly handled by the code that looks for "!!" and adds it to exclusion filter before performing the query.

The fix was to make sure that the `params` value passed into `findExclusionParams()` was a dict, and also to make sure the exclusion filter work was done before `_buildFilter()` was called on the `params` object.

### DIFF
--- a/Products/Zuul/routers/zep.py
+++ b/Products/Zuul/routers/zep.py
@@ -390,6 +390,10 @@ class EventsRouter(DirectRouter):
         @rtype:   generator
         @return:  Generator returning events.
         """
+
+        if isinstance(params, basestring):
+            params = loads(params)
+
         if not self._canViewEvents():
             return
         includeFilter, excludeFilter = self._buildRequestFilters(uid, params, evids, excludeIds)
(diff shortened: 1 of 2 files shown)


Products/Zuul/routers/zep.py
```diff
@@ -390,6 +390,10 @@ def queryGenerator(self, sort='lastTime', dir='desc', evids=None, excludeIds=Non
         @rtype:   generator
         @return:  Generator returning events.
         """
+
+        if isinstance(params, basestring):
+            params = loads(params)
+
```
rDr4g0n: if this code is being hit by the "export" button in the UI, params will be a json string, so parse it into a dict

Products/Zuul/routers/zep.py
```diff
@@ -626,8 +630,8 @@ def _buildRequestFilters(self, uid, params, evids, excludeIds):
             log.debug('Found specific event ids, adding to params.')
             includeUuids = evids
 
-        includeFilter = self._buildFilter([uid], params, specificEventUuids=includeUuids)
         exclude_params = self._filterParser.findExclusionParams(params)
+        includeFilter = self._buildFilter([uid], params, specificEventUuids=includeUuids)
```
rDr4g0n: `findExclusionParams()` will find and remove from `params` any search fields containing "!!", so it must be run before `_buildFilter()`

PR #1357, opened 2015-12-16
ZEN-21353: Migrate netcat changes in HBase svcs

alapidas:
This is necessary due to the fact that CC 1.0 had an hbase image that
had BSD-based `netcat` and 1.1+ has GNU `netcat`, and the options taken
by each are different


Products/ZenModel/migrate/updateHBaseForNetcat.py
```diff
@@ -0,0 +1,67 @@
...
+import servicemigration as sm
+sm.require("1.0.0")
+
+
+class UpdateHBaseForNetcat(Migrate.Step):
+    """ Recursively update all the HBase service healthchecks.  Necessary
+    due to the hbase image changing from an ubuntu base to a centos base,
+    which means BSD netcat to GNU netcat """
+
+    version = Migrate.Version(5,1,70)
```
tgjonestx: Looks like we're inconsistent on these versions across the service migration scripts. Most are 5,0,70, but some are 5,1,70. Not sure what the right number should be, but seems like all of the service migrations for 5.1.1 should have the same version

tgjonestx:
Looks good other than the 1 comment I had.

Products/ZenModel/migrate/updateHBaseForNetcat.py
alapidas: You're right - since this is going into 5.1 as well, I'll make this `5,0,70` and the port will match.

PR #1358, opened 2015-12-17
fixes ZEN-21357: OR and NOT event console filters not implemented for…

HIMix:
… some fields.


HIMix:
port from: https://github.com/zenoss/zenoss-prodbin/pull/1280

PR #1360, opened 2015-12-18
Several service-migration unittest fixes

ed-:
Refactored the migration scripts that use servicemigration
so they'll only call ctx.commit() if they made any changes
during the course of cutover()

- Added updateZeneventserverHealthCheck unittests
- Added commit_called count to fake context, but disabled for now
- Fixed useBeaker migration and tests
- Added conditional commit to updatezopethreadscount
- Check for localhost in opentsdbconfigs
- Fixed switchlogginglevel migration
- Fixed removeemptygraphs
- Fixed makeMemcachedUpdates
- Fixed most of convertRunsToCommands; it's not completely compliant with the unittest commit_called count yet.


ed-:
I'll revisit this later.

PR #1365, opened 2015-12-21
Foundation to start removing zcatalogs

fraibacas:
- Add method get_indexable_peers to ITreeSpanningComponent. The method returns the objects on the other tree of an ITreeSpanningComponent that must be indexed when the tree spanning component is updated
- Added interface IObjectEventsSubscriber. Objects that need to perform any actions when Zope object events happen will implement this interface instead of adding such actions to the index_object /unindex_object methods
- Fix IpAddresses path being indexed under the path relative to the ip inteface instead of the primary path
- Minor tweaks to model_catalog and model_catalog_tool
- Allow passing basic queries in dict format
- Remove layer3 and ipSearch ZCatalogs
- Remove layer2 ZCatalog
- Added methods to retrieve device components from solr. We will keep using the componentSearch ZCatalog due to better performance
- To store decimal ips as string they need to have the same number of characters for solr to properly make the comparison
- Do not index to global catalog and remove global catalog searches for zenoss core
- Updated data manager to add mid transaction consistency
- Improve Network Tree performance

### DIFF
--- a/Products/DataCollector/ApplyDataMap.py
+++ b/Products/DataCollector/ApplyDataMap.py
@@ -191,6 +191,10 @@ class ApplyDataMap(object):
 
         if changed:
             device.setLastChange()
+            # This is for IpInterfaces so the device get its paths updated.
+            # Should we think of doing this differently?
+            #
+            notify(IndexingEvent(device))
 
         log.debug(
             "_applyDataMap for Device %s will modify %d objects for %s",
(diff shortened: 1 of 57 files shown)


dougsyer:
Not sure if this is relevant but as an msp and as somone supporting multiple cloud services anything you can do that helps provide a consistent interface for accessing ips in a mulitirealm environment would be appreciated.

PR #1370, opened 2015-12-29
ZEN-14768: Delete pdb line

olst:


vikt0rs:
I hope, someday we will run unittests...

PR #1374, opened 2016-01-04
fixes ZEN-19385: Change "zSnmpMonitorIgnore" to False thereby the Pin…

itshane:
…g related device will be modeled and the device status will be updated.


jplouis:
Enabling SNMP modeling on the Ping device class by default is not desirable.

PR #1393, opened 2016-01-08
ZEN-19551 Editing a Product (Manufacturers Router) Fails Due to Bad A…

romanbakaleyko:
…udit Logging Input


jpeacock-zenoss:
Would you elaborate on why these changes fix the bad argument passed to `audit`?

PR #1394, opened 2016-01-08
fixes ZEN-21208: ZODB dump has been updated.

itshane:


itshane:
Based on Zenoss Core tempalte

PR #1399, opened 2016-01-12
fixes ZEN-21486: Central Query log config is set to WARN instead of INFO

HIMix:


Products/ZenModel/migrate/switchLoggingLevel.py
```diff
@@ -21,7 +21,7 @@
 
 class SwitchLoggingLevel(Migrate.Step):
 
-    version = Migrate.Version(5, 0, 70)
+    version = Migrate.Version(5, 2, 0)
```
wwwtyro: This should still be 5, 0, 70... it's a weird versioning convention we have.

wwwtyro:
Also there's a test associated with this [here](https://github.com/zenoss/zenoss-prodbin/blob/ZEN-21486/Products/ZenModel/migrate/tests/test_switchLoggingLevel.py). We need to make sure it runs correctly.

wwwtyro:
You can find details on running the tests [here](https://github.com/zenoss/zenoss-prodbin/blob/ZEN-21486/Products/ZenModel/migrate/README.md).

PR #1400, opened 2016-01-12
fixes ZEN-21487: Central Query log config is set to WARN instead of INFO

HIMix:


wwwtyro:
Is there a migration script for this?

wwwtyro:
This needs to be fixed in develop first, then ported here.

PR #1405, opened 2016-01-12
Feature/zen 21383.2

jpeacock-zenoss:


Products/ZenUtils/controlplane/application.py
```diff
@@ -13,20 +13,146 @@
...
+        """
+        since = int((time.time() - self._lastUpdate) * 1000)
+        # No refresh if no time has elapsed since the last update
+        if since == 0:
+            return
+        changes = self._client.getChangesSince(since)
+        if changes:
+            # Changes exist, so refresh the last update time.
+            self._lastUpdate = time.time()
+        for changed_svc in changes:
```
rDr4g0n: should this for...in be hit even if there are no changes?

jpeacock-zenoss: The `getChangesSince` call usually returns an empty list so the `for` won't execute any iterations.

jpeacock-zenoss: The `getChangesSince` call itself appears to be pretty quick.

rDr4g0n: Got it.

PR #1409, opened 2016-01-13
ZEN-21569 Add log.info to migrations

ed-:
At minimum, each cutover function logs that it's starting, and how many services are about to be affected.


wwwtyro:
Could this not be added to whatever calls cutover?

ed-:
Some of it, sure. I'll start in on that.

wwwtyro:
I guess the first line should be straightfoward. It'd be nice if we could enforce this logging in code instead of by convention. Not sure how tall an order that would be.

PR #1412, opened 2016-01-14
ZEN-21177: Zendisc adding a new device will not perform modeling

olst:


Products/DataCollector/zendisc.py
```diff
@@ -285,7 +286,7 @@ def findRemoteDeviceInfo(self, ip, devicePath, deviceSnmpCommunities=None):
             self.log.debug("Override acquired community strings")
             # Override the device class communities with the ones set on
             # this device, if they exist
-            communities = snmp_conf['zSnmpCommunities']
+            communities = snmp_conf['zSnmpCommunities'] or DEFAULT_COMMUNITIES
```
jpeacock-zenoss: `snmp_conf` is supposed to contain the device class configuration values, which, I believe, are captured with the awareness of the device class hierarchy.  So asking for the configuration properties of `/Server/SSH` will contain the values from `/Server` that `/Server/SSH` doesn't override.  With that said, I'm not sure adding `or DEFAULT_COMMUNITIES` will provide any benefit as `snmp_conf['zSnmpCommunities']` will have a value if someone set that property on the `/Server` device class.

It also appears that "public private" is the default zSnmpCommunities value at the `/Server` device class level.  Unless somebody removes that string, the `or` segment of this expression is never evaluated.

olst: Main goal of `or DEFAULT_COMMUNITIES` is to not allow to `community` variable contain blank value. Actually I don't care which value is set to `zSnmpCommunities` unless it's blank. ZEN-21177 issue exists because of `zSnmpCommunities == None`

PR #1414, opened 2016-01-14
make processes panels vertically resizable

rDr4g0n:


dougsyer:
Hurray.  Would be great id win services also could be resized and be able to multiselect drag into containers and multiselect delete.  Omg also if i could multiselect delete and drop in the mib import management would be sweet

PR #1420, opened 2016-01-15
ZEN-21583: Zendisc adding a new device will not perform modeling

olst:


jplouis:
So if we model with default communities if the customer has set zSnmpCommunities to blank, won't collection fail since, presumably, that won't have any community strings set?

olst:
If customer clear `zSnmpCommunities`, model will fail. If we use default value for `zSnmpCommunities` in case of its blank value, model won't fail.

PR #1429, opened 2016-01-19
fixes ZEN-21604: The event related to the device component does not h…

itshane:
…andled properly.


itshane:
UML Class Diagram - http://pix.my/mclC6YpT

PR #1438, opened 2016-01-22
ZEN-21460 make beaker changes directly to configFiles

ed-:


ed-:
The beaker changes made it to the config files, but the services still aren't properly answering.

PR #1439, opened 2016-01-22
ZEN-21704 Write beaker changes directly to zope.conf

ed-:
Port of #1438 to support/5.1.x


ed-:
The beaker changes made it to the config files, but the services still aren't properly answering.

PR #1443, opened 2016-01-27
fixes ZEN-3852: Loading Zenpack overrides user-configured settings.

itshane:


jplouis:
@cluther Can you comment on the efficacy of this solution.

cluther:
Looking.

cluther:
I'm going to need to build a develop environment to test this. Given the impact of a change like this, it's also going to need unit tests that test the various scenarios of what happens when XML is loaded with no existing objects, partial existing objects, and full existing objects.

We're also going to have to spread the word to groups that develop ZenPacks before merging something that changes behavior like this. It will have a significant impact on at least the solutions and services groups at Zenoss. It'll also be a big impact to our users and customers who develop ZenPacks.

dougsyer:
not sure if i have access to this ticket what is being solved?

PR #1451, opened 2016-01-28
Please review branch feature/ZEN-21784

cchristiezen:
Fixes for the following unit tests (ZEN-21784):
- testGets (Products.ZenModel.tests.testFileSystem.TestFileSystem)
- testGets (Products.ZenModel.tests.testHardDisk.TestHardDisk)
- testGets (Products.ZenModel.tests.testWinService.testWinService)


cchristiezen:
It was failing getting the parent url.  Since the parent url is the device url, it should check against that value.

PR #1457, opened 2016-02-02
ZEN-21460 MetricShipper service failing healthchecks

ed-:
Adds 'memcached' as an import to Zauth to fix the failing MetricShipper healthcheck.


ed-:
retest this

PR #1459, opened 2016-02-02
ZEN-21449 Maintenance Windows Use UTC

romanbakaleyko:


Products/ZenUI3/browser/resources/js/zenoss/devicemanagement/Administration.js
```diff
@@ -11,6 +11,8 @@
 
 Ext.ns('Zenoss', 'Zenoss.devicemanagement');
 
+    timeZone = "";
```
jpeacock-zenoss: 1. This should be declared with `var` so that it's only visible within this closure.
2. `timeZone` is *not* actually used anywhere that I can see.

Products/ZenUI3/browser/resources/js/zenoss/devicemanagement/Administration.js
```diff
@@ -257,7 +259,7 @@ Ext.ns('Zenoss', 'Zenoss.devicemanagement');
                                     items: [
                                         {
                                             xtype: 'label',
-                                            text: _t('Time:'),
+                                            text: Ext.String.format(_t('Time ({0})'), tz),
```
jpeacock-zenoss: I think `tz` was supposed to be `timeZone`?

Products/ZenUI3/browser/resources/js/zenoss/devicemanagement/Administration.js
```diff
@@ -1012,6 +1015,13 @@ Ext.define("Zenoss.devicemanagement.Administration", {
             }
             data = selected[0].data;
             maintDialog(this, data);
+        },
+        beforeRender: function() {
+            Zenoss.remote.DeviceManagementRouter.getTz({}, function(response){
+                if(response.success){
+                    tz = response.data;
```
jpeacock-zenoss: I think `tz` is supposed to be `timeZone`?

Products/ZenUI3/browser/resources/js/zenoss/devicemanagement/Administration.js
```diff
@@ -958,7 +960,8 @@ Ext.define("Zenoss.devicemanagement.Administration", {
                         sortable: true,
                         renderer: function(val){
                             var chunk = val.split(".");
-                            return chunk[0];
+                            var tz = val.split(" ")[2]
+                            return chunk[0]+" "+tz;
```
jpeacock-zenoss: What's going on with this code?

jpeacock-zenoss: OK. This is removing the microseconds portion of the original value, which looks like
> 2016/02/02 00:00:00.000 UTC

PR #1470, opened 2016-02-04
added sudo to serviced version command

bwf93:


alapidas:
Why is `sudo` now necessary here?

PR #1482, opened 2016-02-10
fixes ZEN-20808: Fix for ZEN-17576 needs to be made configurable in d…

HIMix:
…aemon files


Products/ZenModel/migrate/addDisablePingPerspectiveOption.py
```diff
@@ -0,0 +1,69 @@
+##############################################################################
+#
+# Copyright (C) Zenoss, Inc. 2015, all rights reserved.
```
jplouis: year is wrong

PR #1504, opened 2016-02-17
Component lists are sorted by a combination of alpha and numeric sort…

rphillipsz:
…s so they appear in the correct order


rphillipsz:
retest this, please

PR #1517, opened 2016-02-18
ZEN-22068 Multigraph reports show same data

carlcamera:
Multigraph reports when displaying identically-named components (think eth0 on several devices)
show only the data from the first component rather than multiple lines from multiple devices.
The fix is to uniquely identify the MetricServerGraphPoint by including device context value
into the line's id.

### DIFF
--- a/Products/Zuul/infos/metricserver.py
+++ b/Products/Zuul/infos/metricserver.py
@@ -136,6 +136,20 @@ class ColorMetricServiceGraphPoint(MetricServiceGraph):
         self._contextRPN = False
         super(ColorMetricServiceGraphPoint, self).__init__(graph, context)
 
+    @property
+    def id(self):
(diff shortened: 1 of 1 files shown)


Products/Zuul/infos/metricserver.py
```diff
@@ -136,6 +136,20 @@ def __init__(self, graph, context):
...
+        if self._multiContext:
+            dev = self._context.device()
+            if dev and dev.id != self._context.id:
+                return "%s %s %s" % (dev.id, self._context.id, self._object.id)
+            else:
+                return "%s %s" % (self._context.id, self._object.id)
+        return self._object.id
+
+    @property
+    def name(self):
```
jplouis: I'm concerned that you changed the value for the name property.  I believe this value gets used to calculate RPN expressions.

PR #1528, opened 2016-02-25
ZEN-15410: Cross Site Scripting from Exposed Helper Methods

serge30:
https://jira.zenoss.com/browse/ZEN-15410

Zope doesn't publish methods with no docstring set. Added the decorator to remove docstrings.

### DIFF
--- a/Products/ZenModel/DataRoot.py
+++ b/Products/ZenModel/DataRoot.py
@@ -37,7 +37,7 @@ import os
 import sys
 import string
 from Products.ZenMessaging.audit import audit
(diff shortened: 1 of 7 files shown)


Products/ZenModel/DataRoot.py
```diff
@@ -467,6 +467,7 @@ def reportError(self):
 
 
     #security.declareProtected('View', 'writeExportRows')
```
jpeacock-zenoss: Please remove the commented out code.

PR #1530, opened 2016-02-25
feature/ZEN-22160

morr-zenoss:
Cherry-picked #1524

Pull request created by zendev port


Products/DataCollector/SnmpClient.py
```diff
@@ -244,7 +244,10 @@ def clientFinished(self, result):
                 log.exception("Device %s had an error: %s",self.hostname,result)
         else:
             self._sendStatusEvent('SNMP agent up', eventKey='agent_down', severity=Event.Clear)
-        self.proxy.close()
+        try:
+            self.proxy.close()
+        except AttributeError:
+            log.info("Caught AttributeError closing SNMP connection.")
```
kwalkerz: Is there a reason you log this as Info rather than an Error? Is it common for a Session object to not have a 'sess' attribute?

morr-zenoss: I think I originally had that as a warning, and someone (JP?) had me change it to info.

PR #1539, opened 2016-03-01
ZEN-19894 Audit Log is filling with user=Unknown action=Failed messag…

romanbakaleyko:
…es by default


romanbakaleyko:
Not sure what migrate version should I use.

PR #1542, opened 2016-03-02
ZEN-19894 Audit Log is filling with user=Unknown action=Failed messag…

romanbakaleyko:
…es by default


Products/ZenModel/fixZauthHealthCheck.py
```diff
@@ -0,0 +1,43 @@
...
+import logging
+log = logging.getLogger("zen.migrate")
+
+import Migrate
+import servicemigration as sm
+sm.require("1.0.0")
+
+
+class FixZauthHealthCheck(Migrate.Step):
+    "Use different curl request to prevent `authentication failed` spam in audit.log"
```
jpeacock-zenoss: FYI, the idiomatic way to write documentation strings in Python is to use triple quotes, i.e.
```
    """Use different curl request to prevent 'authentication failed' span in audit.log"""
```

PR #1548, opened 2016-03-04
Apply prepId to service organizers.

rphillipsz:
Service names were sanitized by Products.ZenUtils.Utils.prepId, but
service organizer names were not.

### DIFF
--- a/Products/ZenModel/Organizer.py
+++ b/Products/ZenModel/Organizer.py
@@ -53,6 +53,7 @@ class Organizer(ZenModelRM, EventView):
         @type description: string
         @rtype: Organizer
         """
+        id = self.prepId(id)
         ZenModelRM.__init__(self, id)
         self.description = description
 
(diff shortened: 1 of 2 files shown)


Products/ZenModel/Organizer.py
```diff
@@ -355,8 +357,12 @@ def getOrganizer(self, path):
         >>> dmd.Events.Status.getOrganizer('/Events/Status/Snmp')
         <EventClass at /zport/dmd/Events/Status/Snmp>
         """
-        if path.startswith("/"): path = path[1:]
-        return self.getDmdRoot(self.dmdRootName).unrestrictedTraverse(path)
+        path = self.prepId(path)
```
iancmcc: This is probably the wrong place to do this. You want to do it down below, before the _getOb call.

Products/ZenModel/Organizer.py
```diff
@@ -183,6 +184,7 @@ def manage_addOrganizer(self, newPath, factory=None, REQUEST=None):
         if factory is None:
             factory = self.__class__
         if not newPath: return self.callZenScreen(REQUEST)
+        newPath = self.prepId(newPath)
```
iancmcc: This can't be done here. It needs to be below this. if newPath is a path with slashes, those slashes will be sanitized.

PR #1553, opened 2016-03-10
Feature/zen 15567

bwf93:


rDr4g0n:
Please make a few changes to the form to make it a bit more comfortable for the user:
1) single column is almost always preferable over multiple columns
2) add a sentence above the field to the give the user some context of what theyre filling (they can infer that its smtp related from the field names, but they shouldnt need to do that)
3) Add validation so that if a host is present, a port must also be present

bwf93:
Updated to address the requested changes.

PR #1567, opened 2016-03-15
ZEN-22385: Extend add_event() method signature

vikt0rs:
Added **kwargs param to add_event() to add more option to API call.

Cherry-Picked From: 55205328ea8aee945be9ab0c7f0a90326c536440


Products/Zuul/routers/zep.py
```diff
@@ -843,7 +843,8 @@ def updateEventSummaries(self, update, event_filter=None, exclusion_filter=None,
 
 
     @require(ZEN_MANAGE_EVENTS)
-    def add_event(self, summary, device, component, severity, evclasskey, evclass=None):
+    def add_event(self, summary, device, component, severity, evclasskey,
```
smousa: What else is calling this, such that you require **kwargs to be passed?

smousa: nevermind...I thought the bottom diff was on a different function

PR #1590, opened 2016-03-22
ZEN-22411: add tracking for metric prefixes, so RPN update works properly

morr-zenoss:
The fix for ZEN-22068 changed the way ids could be prefixed to make them unique. This broke multigraph reports which used RPNs, as the rpn update code expects the _context.id (only) to be prepended to the _object.id. This change adds a _prefix field and saves the prefix, whether it's  dev.id + _context.id or just _context.id. The _prefix field is then used to update the RPN, rather than assuming the metric will have _context.id (only) prepended. I have verified this on my dev system using test cases pertinent to both ZEN-22068 and ZEN-22411.


Products/Zuul/infos/metricserver.py
```diff
@@ -347,7 +350,7 @@ def _getGraphPoints(self, klass):
     def _updateRPNForMultiContext(self, infos, knownDatapointNames):
         for info in infos:
             if getattr(info._object, "rpn", False):
-                newRPN = mutateRPN(info._context.id, knownDatapointNames, info._object.rpn)
+                newRPN = mutateRPN(info._prefix, knownDatapointNames, info._object.rpn)
```
jplouis: Can you make _prefix a method call instead?  Don't like the side effect of having to have called the "id" or "name" method first in order for this to work..

PR #1593, opened 2016-03-23
Component Graphs filter will be showing only contextually related gra…

itshane:
…phs.

Fixes ZEN-22620.


itshane:
Original solution has been made by Chet for LinuxMonitor ZP https://github.com/zenoss/ZenPacks.zenoss.LinuxMonitor/pull/60/commits/3e94441cb702b7b2c3282637d568e9850c03c1e8#diff-7284140a91f7dc12cac56ac67c3c5ff9R49

PR #1594, opened 2016-03-23
ZEN-22626: Update page tips for Dashboard

kwalkerz:


kwalkerz:
Here's what it looks ilke now:
result

PR #1596, opened 2016-03-24
Change flapping threshold to be an integer instead of a severity

jhuie-zenoss:


fraibacas:
good to merge once the build tests have finished

PR #1603, opened 2016-03-31
Please review branch feature/ZEN-22695.port

wwwtyro:
Cherry-picked #1600

Pull request created by zendev port

### DIFF
--- a/Products/ZenUtils/extdirect/zope/router.py
+++ b/Products/ZenUtils/extdirect/zope/router.py
@@ -18,8 +18,8 @@ class ZopeDirectRouter(DirectRouter):
     def __call__(self):
         # Allow only requests with application/json content type as text/plain
         # content type is used for CSRF attacks.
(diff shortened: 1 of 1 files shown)


Products/ZenUtils/extdirect/zope/router.py
```diff
@@ -18,8 +18,8 @@ def __init__(self, context, request=None):
     def __call__(self):
         # Allow only requests with application/json content type as text/plain
         # content type is used for CSRF attacks.
-        content_type = self.request.get('CONTENT_TYPE')
-        if not isinstance(content_type, basestring) or content_type.lower() != 'application/json':
+        content_type = self.request.get_header('content-type')
```
tgjonestx: HTTP header names are case-insensitive - pls make sure that that get_header makes a case-insensitive.
Otherwise, clients who made a valid request with a value like `Content-Type` will be denied incorrectly.

PR #1606, opened 2016-04-05
ZEN-20284 RM 5.x Commands issues against a device via the UI (ping, t…

romanbakaleyko:
…raceroute, snmpwalk) are run on master, not on collector host

### DIFF
--- a/Products/ZenModel/Commandable.py
+++ b/Products/ZenModel/Commandable.py
@@ -195,7 +195,7 @@ class Commandable:
         '''
         command = cmd.command
         # make sure we are targeting the right collector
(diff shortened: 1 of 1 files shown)


Products/ZenModel/Commandable.py
```diff
@@ -195,11 +195,9 @@ def compile(self, cmd, target):
         '''
         command = cmd.command
         # make sure we are targeting the right collector
-        if not command.startswith("dcsh") and hasattr(target, "getPerformanceServerName"):
+        if hasattr(target, "getPerformanceServerName"):
```
jpeacock-zenoss: I think removing the check for `dcsh` may break existing custom commands.

PR #1607, opened 2016-04-05
ifindex could be missing, use row index

dbouchillon:
Fixes ZEN-22732

https://jira.zenoss.com/browse/ZEN-22732
Happens on certain firmware versions. Reproduced using snmposter
Tested using attached snmpwalks from JIRA


dbouchillon:
bump

PR #1616, opened 2016-04-13
temp_for_review

romanbakaleyko:


romanbakaleyko:
do not merge

Products/ZenUtils/patches/pasmonkey.py
```diff
@@ -331,6 +350,27 @@ def updateCredentials(self, request, response, login, new_password):
         request.SESSION.set('__ac_logged_as', user_id)
         request.SESSION.set('__ac_logged_info', info)
 
+        if self.attempt.get(user_id):
```
serge30: `user_id in self.attempt`

Products/ZenUtils/patches/pasmonkey.py
```diff
@@ -310,7 +317,19 @@ def updateCredentials(self, request, response, login, new_password):
...
+    current = self.attempt.get(login, {})
+    if current.get('counter', 1) >= ATTEMPT_LIMIT:
+        lastAttempt = current.get('lastAttempt')
+        if datetime.now() - lastAttempt < LOCKED_TIMEOUT:
+            isBlocked = True
+        else:
+            current['counter'] = 0
+            isBlocked = False
+    else:
+        isBlocked = False
```
serge30: `isBlocked` is already initialized as False, thus this lines may be removed

Products/ZenUtils/patches/pasmonkey.py
```diff
@@ -331,6 +350,27 @@ def updateCredentials(self, request, response, login, new_password):
         request.SESSION.set('__ac_logged_as', user_id)
         request.SESSION.set('__ac_logged_info', info)
 
+        if self.attempt.get(user_id):
+            del self.attempt[user_id]
+    else:
+        if current:
+            now = datetime.now()
```
serge30: for timestamps better to use `time.time()`

Products/ZenUtils/patches/pasmonkey.py
```diff
@@ -331,6 +350,27 @@ def updateCredentials(self, request, response, login, new_password):
...
+            last = current['lastAttempt']
+            if isBlocked:
+                remaining = LOCKED_TIMEOUT - (now - last)
+                msg = "Account [%s] is locked due to numerous bad attempts, try again in %s seconds or contact administrator" % (login, remaining.seconds )
+                log.warning(msg)
```
serge30: Also it would be great to log such cases to audit log

Products/ZenUtils/patches/pasmonkey.py
```diff
@@ -331,6 +350,27 @@ def updateCredentials(self, request, response, login, new_password):
...
+                request.SESSION.set('__blocked', msg)
+            else:
+                current['counter'] += 1
+                current['lastAttempt'] = now
+                self.attempt[login] = current
+        else:
+            structure = {'lastAttempt' : None, 'counter' : None}
+            structure['lastAttempt'] = datetime.now()
+            structure['counter'] = 1
+            self.attempt[login] = structure
```
serge30: `structure = {'lastAttempt' : datetime.now(), 'counter' : 1}`

Products/ZenUtils/patches/pasmonkey.py
```diff
@@ -288,6 +291,10 @@ def updateCredentials(self, request, response, login, new_password):
     # PAS sends to this methos all credentials provided by user without
     # checking.  So they need to be validate before session update.
 
+    LOCKED_TIMEOUT = timedelta(seconds=300)
+    ATTEMPT_LIMIT = 5
```
serge30: Constants is better define outside the function

Products/ZenModel/UserSettings.py
```diff
@@ -449,6 +449,28 @@ def manage_deleteUsers(self, userids=(), REQUEST=None):
...
+    def manage_unlockUsers(self, userids=(), REQUEST=None):
+        """Unlocked a list of zenoss users from the system which are locked
+            due to numerous bad login attempts.
+        """
+
+        for user_id in userids:
+            try:
+                del self.acl_users.sessionAuthHelper.attempt[user_id]
+            except KeyError:
+                pass
```
jpeacock-zenoss: Prefer to use exceptions for 'big' errors and I think exceptions are expensive, even in Python.  So, there are two cleaner alternative ways to code this:
```
for user_id in userids:
    self.acl_users.sessionAuthHelper.attempt.pop(user_id, None)
```
or
```
helper = self.acl_users.sessionAuthHelper
for user_id in (uid for uid in userids if uid in helper.attempt):
    del helper.attempt[user_id]  # or helper.attempt.pop(user_id)
```

Products/ZenModel/UserSettings.py
```diff
@@ -449,6 +449,28 @@ def manage_deleteUsers(self, userids=(), REQUEST=None):
                 audit('UI.User.Delete', username=userid)
             return self.callZenScreen(REQUEST)
 
+    security.declareProtected(ZEN_MANAGE_DMD, 'manage_unlockUsers')
+    @validate_csrf_token
+    def manage_unlockUsers(self, userids=(), REQUEST=None):
+        """Unlocked a list of zenoss users from the system which are locked
+            due to numerous bad login attempts.
```
jpeacock-zenoss: No need to indent this line (the line starting with 'due').  It should be lined up with the first '"', e.g.
```
    """Unlocked a list of zenoss users from the system which are locked
    due to numerous bad login attempts.
    """
```

Products/ZenModel/migrate/addAccountLocker.py
```diff
@@ -0,0 +1,34 @@
...
+    version = Migrate.Version(5,1,2)
+
+    def cutover(self, dmd):
+        
+        # Adding attribute if it is not in the place.
+        if not hasattr(dmd.zport.acl_users.sessionAuthHelper, 'attempt'):
+            dmd.zport.acl_users.attempt = OOBTree()
+
+        if not hasattr(dmd.zenMenus.User_list.zenMenuItems, 'unlockUser'):
+            dmd.zenMenus.User_list.manage_addZenMenuItem(id="unlockUser", description='Unlock Users...', action='dialog_unlockUsers', permissions=('Manage DMD',), isdialog=True)
```
jpeacock-zenoss: The ellipses ('...') in the description should be removed.  Ellipses are used for interjecting incomplete thoughts (and sentences) into text.  In this case, "Unlock Users", _is_ a complete thought (i.e. verb noun), so '...' doesn't belong.

Products/ZenUtils/patches/pasmonkey.py
```diff
@@ -310,7 +318,17 @@ def updateCredentials(self, request, response, login, new_password):
...
-    if user_id is None:
+    request.SESSION.set('__blocked', None)
+    current = self.attempt.get(login, {})
+    if current.get('counter', 1) >= ATTEMPT_LIMIT:
+        lastAttempt = current.get('lastAttempt')
+        if time.time() - lastAttempt < LOCKED_TIMEOUT:
+            isBlocked = True
+        else:
+            current['counter'] = 0
+            isBlocked = False
```
jpeacock-zenoss: `isBlocked` is already `False` at this point, isn't it?

Products/ZenUtils/patches/pasmonkey.py
```diff
@@ -310,7 +318,17 @@ def updateCredentials(self, request, response, login, new_password):
     # user and return immediately if authenticated
     user_id, info = acl_users._tryEmergencyUserAuthentication(credentials)
 
-    if user_id is None:
+    request.SESSION.set('__blocked', None)
+    current = self.attempt.get(login, {})
```
jpeacock-zenoss: if you use `self.attempt.setdefault(login, {})` instead, you won't need to assign `current` back to `attempt[user_id]` further down in the code (line 367).

jpeacock-zenoss:
The only potential drawback I see in this implementation is that, I think, the `__blocked` attribute, at some point, is sent across the wire between the server and browser.  I've seen the `__ac_*` attributes on the wire, so I assume `__blocked` will show up as well.  And if it does appear, then it can be hacked.  If there is a way to store session variables in the session database _without_ having them appear on the wire, that would be the preferred way.

If my thinking is wrong here, please show me.

PR #1619, opened 2016-04-13
use daemon commands in zenhome/bin

jplouis:


jplouis:
retest this

smousa:
why not change the value of daemonpath?

jplouis:
Because it is documented as being the absolute path and didn't want to change the meaning of the value in case it is being used by a zenpack.

PR #1627, opened 2016-04-19
ZEN-22458 Auto-close login failed events after successful login

carlcamera:
This closes out (in many but not all cases) the failed-login event for a server when
a successful SSH login occurs. See ZEN-22458 for details.


carlcamera:
test again

carlcamera:
retest this

jplouis:
I read the comments in the ticket and do not understand why the "zCommandUsername" event cannot be cleared by zenmodeler.

carlcamera:
Because we cannot spoof the context under which re run?  The zCommandUsername event is created under the context of zencommand. The modeler is running under zenmodeler. You were the one who helped me understand that we cannot "spoof" context.

PR #1631, opened 2016-04-21
ZEN-22801 Update health checks for possibility that the db passwords …

romanbakaleyko:
…are customized


Products/ZenModel/migrate/fixMariadbHealthCheck.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def cutover(self, dmd):
+
+        try:
+            ctx = sm.ServiceContext()
+        except sm.ServiceMigrationError:
+            log.info("Couldn't generate service context, skipping.")
+            return
+
+        mariadb-model = filter(lambda s: s.name == "mariadb-model", ctx.services)
```
jplouis: This won't migrate core.  The name of the service is different and core.

romanbakaleyko:
retest this

PR #1632, opened 2016-04-21
ZEN-22812 portlet chart renders incorrectly

carlcamera:
fixes problem: chart is squished and overlapped by top text.


carlcamera:
retest this

PR #1636, opened 2016-04-25
ZEN-21879: 'SVGElement.offsetParent' is deprecated and will be removed in M50, around April 2016

Deer-WarLord:


rDr4g0n:
Please leave a comment on this code identifying why this is a workaround and that it is copy/pasted from extjs 4.1.3 (or whatever the appropriate version). This way in future upgrades we will be able to determine if it can be removed.

Products/ZenUI3/browser/resources/js/zenoss/zenoss.js
```diff
@@ -68,6 +68,80 @@ Ext.namespace('Zenoss.env');
 
 Ext.QuickTips.init();
 
+Ext.enableGarbageCollector = false;
+Zenoss.env.enableGarbageCollector = true;
+
```
Deer-WarLord: This workaround is the full copy of current ExtJs GC function in Zenoss except lines with deprecated function. The reason I did it is because in new versions of ExtJs I didn't find fix of deprecated calls in GC function

PR #1641, opened 2016-04-27
tool to be able to pack systems that have a huge number of stale objects

fraibacas:


fraibacas:
Please do not merge. Just for review purposes.

PR #1649, opened 2016-05-03
protobuf strings are utf-8 encoded instead of ascii encoded

bwf93:


Products/ZenMessaging/queuemessaging/adapters.py
```diff
@@ -112,12 +112,12 @@ def fill(self, proto):
...
-    decoded safely with 7-bit ASCII.  If any specialized encoding is desired, it
+    decoded safely with UTF-8.  If any specialized encoding is desired, it
     is the responsibility of the caller/sender to take care of it.
     """
     if isinstance(s, str):
         try:
-            unicode(s, 'ascii')
+            s = unicode(s, 'utf_8')
         except UnicodeDecodeError:
             s = str(s.decode('ascii','ignore'))
```
johnlhamilton: Should this be changed to utf8 as well?

johnlhamilton:
Would it make sense to add a unit test for this?

Products/ZenMessaging/queuemessaging/adapters.py
bwf93: It just failed to encode as unicode so trying to do that here should fail again.  This line is a worst case scenario that is going to try to force it to ascii so at least some of the string gets encoded.

PR #1660, opened 2016-05-06
ZEN-22856 We should add an ID to identify the instance when using mul…

romanbakaleyko:
…tiple hub instances

### DIFF
--- a/Products/ZenHub/PBDaemon.py
+++ b/Products/ZenHub/PBDaemon.py
@@ -643,6 +643,22 @@ class PBDaemon(ZenDaemon, pb.Referenceable):
         """
         self.log.info("Attempting to connect to zenhub")
 
+    def getZenhubInstanceId(self):
+        """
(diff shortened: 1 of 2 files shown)


Products/ZenHub/PBDaemon.py
```diff
@@ -643,14 +643,30 @@ def connecting(self):
...
     def gotPerspective(self, perspective):
         """
         This gets called every time we reconnect.
 
         @parameter perspective: Twisted perspective object
         @type perspective: Twisted perspective object
         """
-        self.log.info("Connected to ZenHub")
+        self.getZenhubInstanceId()
         self.perspective = perspective
```
jpeacock-zenoss: The `getZenhubInstanceId` call should be placed after setting the `perspective` attribute because `getZenhubInstanceId` depends on the `perspective` existing and having a valid object.

PR #1665, opened 2016-05-10
ZEN-23055: snmpwalk user command doesn't pass in correct snmpv3 argum…

olst:
…ents

### DIFF
--- a/Products/ZenModel/data/devices.xml
+++ b/Products/ZenModel/data/devices.xml
@@ -1,190 +1,190 @@
 <?xml version="1.0" encoding="ISO-8859-1" ?>
 
 <!--
-    Zenoss RelationshipManager export completed on 2016-03-11 11:26:54.655148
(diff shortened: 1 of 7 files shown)


jpeacock-zenoss:
Many of these changes in the xml file are "real" changes; only the order of the XML properties as they appear in the tags has changed, which is generally an irrelevant in XML.

The only real changes I do see are the formulas.  It's curious that there's a difference here.  I suppose there's a migration script that made these changes but the XML export (and possibly the SQL dump) was never updated.

Products/ZenModel/data/devices.xml
```diff
@@ -11271,7 +11271,7 @@ True
 False
 </property>
 <property type="string" id="maxval" mode="w" >
-long(here.speed) / 8 * .75
+here.speed / 8 * .75
```
jpeacock-zenoss: I surprised to see these formula changes.  I suppose there's a migration script that wasn't accompanied by updates to the XML and SQL dump files.

PR #1666, opened 2016-05-12
fixes ZEN-14671: ValueChange Threshold alerts on change from None to 0

HIMix:


Products/ZenModel/ValueChangeThreshold.py
```diff
@@ -51,30 +52,31 @@ class ValueChangeThresholdInstance(MetricThresholdInstance):
     Threshold that emits an event when a value changes from its previous value. Does not send clear events.
     """
 
-    def __init__(self, id, context, dpNames, eventClass, severity):
-        MetricThresholdInstance.__init__(self, id, context, dpNames, eventClass, severity)
-        self._lastValues = {}
+    lastValues = Map.Locked(Map.Timed({}, 60*60*24)) # 24-hour timeout
```
jpeacock-zenoss: Please explain why this is now a class level variable.  It's not obvious why having an instance variable was not sufficient.

PR #1696, opened 2016-06-01
ZEN-10028: Allow to run datasource test via SSH

vikt0rs:


Products/ZenRRD/zentestcommand.py
```diff
@@ -16,26 +16,21 @@
...
+log = logging.getLogger("zen.zentestcommand")
 snmptemplate = ("snmpwalk -c%(zSnmpCommunity)s "
                 "-%(zSnmpVer)s %(manageIp)s %(oid)s")
 
 
-class TestRunner(ZenScriptBase):
+class TestRunner(ZenScriptBase.ZenScriptBase):
 
     def __init__(self):
         ZenScriptBase.__init__(self, connect=True)
```
glujan: So this piece of code is referring to module `ZenScriptBase` instead of the class with very same name? :confused:

vikt0rs: Agreed, It's a mistake

Products/ZenRRD/zentestcommand.py
```diff
@@ -16,29 +16,24 @@
...
-log = logging.getLogger("zen.zentestcommand")
+import logging
+import sys
 
-import Globals
-from Products.ZenUtils.ZenScriptBase import ZenScriptBase
-from Products.DataCollector.SshClient import SshClient
+import Globals # noqa
+from Products.ZenUtils import Utils
+from Products.ZenUtils import ZenScriptBase
```
jpeacock-zenoss: This isn't typical.
Why import the module, i.e.
```
from Products.ZenUtils import ZenScriptBase
```
instead of the class, i.e.
```
from Products.ZenUtils.ZenScriptBase import ZenScriptBase
```
?

Importing the class avoids all the `ZenScriptBase.ZenScriptBase` code you have to write later.

vikt0rs: hm, I see both cases in Zenoss code and PEP8 allows both cases too.
Personally, I prefer to use `import module`, not `from module import func` but I understand that this question may be a part of holywar and the difference is mainly subjective so I'll agree with any your suggestion.

vikt0rs:
fixed style of imports, as we discussed with @jpeacock-zenoss

PR #1706, opened 2016-06-06
ZEN-23208: Zope 'answering' health check retries on failure

glujan:
Port of #1682 and #1683


romanbakaleyko:
wrong version in migrate script

PR #1708, opened 2016-06-07
Make component grid page size configurable. Current hardcoded value i…

fraibacas:
…s too high for such a small panel


dougsyer:
In the past when the grid size changed the setting allowing you to unmonitor x interfaces at a time was mismatched not sure id that is now matchjng

PR #1718, opened 2016-06-13
Fixes ZEN-23173.

itshane:
Valid mask attribute for IpRouteEntry is routemask.

### DIFF
--- a/Products/DataCollector/plugins/zenoss/cmd/linux/netstat_rn.py
+++ b/Products/DataCollector/plugins/zenoss/cmd/linux/netstat_rn.py
@@ -83,7 +83,7 @@ class IpRoutesParser(BaseParser):
             if isgate:
                 # default via 10.111.23.1 dev eth0
                 route.id = "0.0.0.0_0"
-                route.mask = '0'
+                route.routemask = '0'
                 route.setTarget = '0.0.0.0/0'  # rfc1519
                 route.setNextHopIp = isgate.group(1)
                 route.setInterfaceName = isgate.group(2)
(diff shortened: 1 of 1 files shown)


itshane:
retest this

Products/DataCollector/plugins/zenoss/cmd/linux/netstat_rn.py
```diff
@@ -98,8 +98,8 @@ def parse(self, results):
             if isroute:
                 # 192.168.99.0/24 dev eth0  scope link
                 route.id = isroute.group(1).replace('/', '_')
-                route.mask = isroute.group(1).split('/', 1)[1]
-                if route.mask == 32:
+                route.routemask = isroute.group(1).split('/', 1)[1]
+                if route.routemask == 32:
```
cluther: It would appear that route.routemask could never == 32 here since it was just parsed out of a string without conversion to int. IpRouteEntry.routemask is defined as type string. So probably best to keep working with it as a string.

Products/DataCollector/plugins/zenoss/cmd/linux/netstat_rn.py
```diff
@@ -110,8 +110,8 @@ def parse(self, results):
             isroute = IP_ROUTE_RECORD_SHORT.search(line)
             if isroute:
                 route.id = isroute.group(1).replace('/', '_')
-                route.mask = isroute.group(1).split('/', 1)[1]
-                if route.mask == 32:
+                route.routemask = isroute.group(1).split('/', 1)[1]
+                if route.routemask == 32:
```
cluther: Same as comment above: route.routemask is a string and should be treated as such.

PR #1733, opened 2016-07-04
ZEN-23195 raise ValueError in case of duplicate mapping

ssOlexBaiko:
This is the port of zenoss/4.2.5-RPS#754


glujan:
Retest this

PR #1752, opened 2016-07-08
ZEN-22899 XSS in MIBs

romanbakaleyko:


Products/Zuul/routers/mibs.py
```diff
@@ -234,8 +235,23 @@ def setInfo(self, **data):
         audit('UI.Mib.Edit', uid, data_=data)
         return DirectResponse.succeed(data=Zuul.marshal(info))
 
+    def _validateOid(self, oid):
+        """
+        Check for illegal characters in OID
+        """
+        oidRegex= re.compile('^[-.0-9]+$')
```
jpeacock-zenoss: `-` is a valid character for SNMP OIDs?

PR #1755, opened 2016-07-08
ZEN-23774 Force cookies to be secure, httponly

ed-:
Service migration to add a rewrite rule to zproxy to force
any cookies with the path "/", including beaker.session, to
be secure and HTTPOnly.

Port of #1754 to support/5.1.x


ed-:
retest this

PR #1767, opened 2016-07-15
Add javascript build

tgjonestx:
https://jira.zenoss.com/browse/ZEN-23963

### DIFF
--- a/makefile
+++ b/makefile
@@ -1,8 +1,17 @@
-VERSION ?= 5.2.0
-BRANCH ?= develop
+VERSION  := 5.2.0
+BRANCH   := develop
+ARTIFACT := prodbin-$(VERSION)-$(BRANCH).tar.gz
 
-.PHONY: all
-all: tar
+.PHONY: all clean build javascript
 
-tar:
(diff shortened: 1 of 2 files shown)


makefile
```diff
@@ -1,8 +1,17 @@
...
+
+build: javascript
+	tar cvfz $(ARTIFACT) Products bin
+
+javascript:
+	make -f javascript.mk build
+
+clean:
+	rm -f $(ARTIFACT)
+	-make -f javascript.mk clean
```
jplouis: is that "-" intentional?

rDr4g0n: I'm guessing that if you try to clean and the files to be cleaned are absent it will exit non-zero, so the dash says to ignore that.

rDr4g0n:
I'm good with this if JP is.

makefile
tgjonestx: Yes - the dash says - don't exit with an error on this step if it fails.  On Jenkins builds, we always want to execute `make clean build`, though some build slaves may not have any files to clean

PR #1776, opened 2016-07-21
ZEN-24095 Change memcached startup and max request

glujan:
Port of #1773


Products/ZenModel/migrate/changeMemcachedStartup.py
```diff
@@ -0,0 +1,58 @@
...
+
+import Migrate
+import servicemigration as sm
+sm.require("1.0.0")
+
+
+class ChangeMemcachedStartup(Migrate.Step):
+    "Change memcached startup to respect config file and update config file"
+
+    version = Migrate.Version(5, 1, 7)
```
jpeacock-zenoss: Is this the right version to use?  The next release is 5.1.5, not 5.1.7.

glujan: You're right, should be 5.1.5

PR #1777, opened 2016-07-22
ZEN-24044 Navigating to paged component grid fails to select targeted…

carlcamera:
… component

Added code for paged component grid to load correct page and select targeted component


dougsyer:
Thanks for this

PR #1790, opened 2016-08-02
ZEN-23871: event export to csv information overload

olst:


Products/ZenUI3/browser/eventconsole/export.py
```diff
@@ -75,11 +77,18 @@ def csv(self, response, archive, **params):
         response.setHeader('Content-Disposition', 'attachment; filename=events.csv')
         from csv import writer
         writer = writer(response)
-
+        events = [dict((k, v) for k, v in evt.iteritems() if v) for _, evt in
+                  self._query(archive, **params)]
```
jpeacock-zenoss: An alternate way to create the dictionary is to use a dictionary comprehension, like so:
```
events = [
    {k: v for k, v in evt.iteritems() if v}
    for _, evt in self._query(archive, **params)
]
```

Products/ZenUI3/browser/eventconsole/export.py
```diff
@@ -75,11 +77,19 @@ def csv(self, response, archive, **params):
...
+        keys = [x.keys() for x in events]
```
jpeacock-zenoss: This can be a generator of iterators, i.e.
```
keys = (x.iterkeys() for x in events)
```
and should still work with the next line.

Products/ZenUI3/browser/eventconsole/export.py
```diff
@@ -75,11 +77,19 @@ def csv(self, response, archive, **params):
...
+        fields = {item for sublist in keys for item in sublist}
```
jpeacock-zenoss: I'm not sure whether this is any faster (or slower), but is probably a little more readable:
```
fields = {item for item in itertools.chain.from_iteratable(keys)}
```
Just remember to `import itertools`

Products/ZenUI3/browser/eventconsole/export.py
```diff
@@ -75,11 +78,19 @@ def csv(self, response, archive, **params):
         response.setHeader('Content-Disposition', 'attachment; filename=events.csv')
         from csv import writer
         writer = writer(response)
-
+        events = [dict((k, v) for k, v in evt.iteritems() if v) for _, evt in
+                  self._query(archive, **params)]
+        keys = (x.iterkeys() for x in events)
+        fields = {item for item in itertools.chain.from_iterable(keys)}
```
jpeacock-zenoss: I'm thinking that `fields` should ultimately be a list rather than a set, e.g.
```
fields = list({item for item in itertools.chain.from_iterable(keys)})
```
It's still created from a set to get the necessary union of the fields.

I say this because it's converted to list when the header is written and also when the number of fields exceeds the `CSV_MAX_COLUMNS` value.  Just do the conversion once.

PR #1794, opened 2016-08-02
Update iframe src with newer Kibana

youngklee:


jafcrocker:
I think this is not acceptable because it is not backwards compatible.  We need to be able to upgrade CC without upgrading RM.  Please check with an architect before merging.

johnlhamilton:
Per discussion with Young it seems that this does not break compatibility between newer CC (Hoth) and and older RM (5.1.x)

PR #1798, opened 2016-08-03
New location for conf files needed during zenoss initialization. Add …

jplouis:
…sitecustomize.py script to tar file

### DIFF
--- a/etc/zope.conf
+++ b/etc/zope.conf
@@ -1,0 +1,1135 @@
+# This is the proposed version for SC / Zope 2.12.1#
+
+###############################################################################
+# Welcome to Zope 2.
+###############################################################################
+#
+# This is the Zope configuration file.  The Zope configuration file
+# shows what the default configuration directives are, and show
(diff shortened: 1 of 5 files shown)


etc/zope.conf
```diff
@@ -0,0 +1,1135 @@
...
+#     use as the initial CGI environment variables. This is useful
+#     when you want to proxy requests from another web server to Zserver,
+#     and would like Zserver's CGI environment to reflect the CGI
+#     environment of the other web server.
+#
+# Default: unset
+#
+# Example:
+#
+# <cgi-environment>
```
rphillipsz: Does this need HTTPS ON?  This is how we've instructed people to turn off HTTPS (setting to OFF) if needed.

jplouis: This is the default file in the image. When running in control center it gets overwritten by the service definition.  This is used during install and if you want to run zope in the image outside of control center.

PR #1816, opened 2016-08-15
ZEN-24100: Removing ZEP metrics container log

svambati:


Products/ZenModel/migrate/updateZepLogbackConfig.py
```diff
@@ -0,0 +1,90 @@
...
+            ctx = sm.ServiceContext()
+        except sm.ServiceMigrationError:
+            log.info("Couldn't generate service context, skipping.")
+            return
+    
+        isUpdated = False
+        service = filter(lambda s: s.name == "zeneventserver", ctx.services)[0]
+       
+        # Update logback.xml
+        cfg = filter(lambda f: f.name == "/opt/zenoss/etc/zeneventserver/logback.xml", service.originalConfigs)[0]
```
jplouis: For the xml we wan't to modify it even if the user has changed it.  There are 2 config file variables to change. See https://github.com/zenoss/zenoss-prodbin/blob/af381284e31b502e5e24e8159000521a04b12e63/Products/ZenModel/migrate/useBeakerInZope.py#L55

PR #1823, opened 2016-08-23
fixed ZEN-20463: RHEL/CentOS 7 fails with > 1 zSshConcurrentSessions …

HIMix:
…SSH Timeout


jpeacock-zenoss:
I recall hearing there is a plan for the Metis release to upgrade Twisted to the latest version which apparently _does_ support the new keys.  If so, this change isn't required for the develop branch.

PR #1830, opened 2016-08-23
feature/zen 24365: exclude Host and ENV data from CallHome report in 5.X or later

corntoole:


corntoole:
You can test this change by doing the following:

1. attaching to a service like zope and running callhome with `python /opt/zenoss/Products/ZenCallHome/callhome.py -M` to verify that it doesn't include host or ENV data within the callhome report
2. running `zendev test -d -- -m Products.ZenCallHome` to make sure the relevant tests pass.

PR #1834, opened 2016-08-25
ZEN-24535: convert event time in email to timezone from user property

rderkach:
${evt/firstTime} and ${evt/lastTime} in notifications are sent in
UTC.
Take a timezone from user propery and convert time in notification
email to that timezone, and group target list of emails by timezones.

[Jira and demo](https://jira.zenoss.com/browse/ZEN-24535)

### DIFF
--- a/Products/ZenModel/actions.py
+++ b/Products/ZenModel/actions.py
@@ -50,6 +50,7 @@ from Products.ZenUtils.IpUtil import getHostByName, isip
 from Products.ZenUtils.guid.guid import GUIDManager
 from Products.ZenUtils.ProcessQueue import ProcessQueue
 from Products.ZenUtils.ZenTales import talEval, InvalidTalesException
+from Products.ZenUtils.Time import convertTimestampToTimeZone, isoToTimestamp
 from zenoss.protocols.protobufs.zep_pb2 import (
     SEVERITY_CLEAR, SEVERITY_INFO, SEVERITY_DEBUG,
     SEVERITY_WARNING, SEVERITY_ERROR, SEVERITY_CRITICAL,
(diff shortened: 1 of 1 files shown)


Products/ZenModel/actions.py
```diff
@@ -394,6 +396,37 @@ def _encodeBody(self, body):
             plain_body = MIMEText(body.decode('ascii', 'ignore'))
         return plain_body
 
+    def _evtTimeToUserTz(self, body, zone_name):
```
jpeacock-zenoss: This function locates a single timestamp within the content.  How can we be sure there won't be additional timestamps?

Products/ZenModel/actions.py
```diff
@@ -394,6 +396,37 @@ def _encodeBody(self, body):
...
+        """
+        Convert event time from email body to timezone from use property.
+        """
+        evt_date = ''
+        evt_date_with_tz = ''
+        try:
+            evt_date = re.search(
+                '(\d+/\d+/\d+ \d+:\d+:\d+.\d+)', body).group()
```
jpeacock-zenoss: Does the millisecond (or nanosecond) portion of the timestamp _always_ present?  If it doesn't, this won't match.

Products/ZenModel/actions.py
rderkach: I was referring to a standard notification template, that is why it locates a single times, and yes it will only change one timestamp, which will be a problem.

Products/ZenModel/actions.py
rderkach: I saw them in every events that came for notification. I can take them from there, since in the result I'm not setting them anyway.

Products/ZenModel/actions.py
rderkach: Removed that function, and now convert the time in event object.

rderkach:
Removed _evtTimeToUserTz function, and now convert the time in event object
Greb a firstTime and lastTime timestamps from event object convert that timestamp to appropriate timezone.

Products/ZenModel/actions.py
```diff
@@ -394,76 +395,104 @@ def _encodeBody(self, body):
...
+        for target_timezone in targets_by_timezones:
+            # Convert timestamp to user timezone
+            signal.event.last_seen_time = long(
+                isoToTimestamp(convertTimestampToTimeZone(
+                    (original_lst / 1000),
+                    target_timezone, '%Y-%m-%d %H:%M:%S'))) * 1000
+            signal.event.first_seen_time = long(
+                isoToTimestamp(convertTimestampToTimeZone(
+                    (original_fst / 1000),
+                    target_timezone, '%Y-%m-%d %H:%M:%S'))) * 1000
```
jpeacock-zenoss: I find this conversion code difficult to read and understand.

Are `original_fst` and `original_lst` just integer values that represent milliseconds?

Perhaps a lambda or helper function would help here, .e.g.
```
def adjustToTimezone(millis, timezone):
    return long(isoToTimestamp(convertTimestampToTimezone(
        millis / 1000, timezone, "%Y-%m-%d %H:%M:%S"
    ))) * 1000
```
and then later
```
        signal.event.first_seen_time = adjustToTimezone(
            signal.event.first_seen_time, target_timezone
        )
        signal.event.last_seen_time = adjustToTimezone(
            signal.event.last_seen_time, target_timezone
        )
```

Products/ZenModel/actions.py
```diff
@@ -394,76 +395,104 @@ def _encodeBody(self, body):
...
+            targets = targets_by_timezones[target_timezone]
```
jpeacock-zenoss: This line is not necessary.  This information can be extracted in the `for` statement itself, e.g.

```
    for target_timezone, targets in targets_by_timezones.iteritems():
```

Products/ZenModel/actions.py
```diff
@@ -394,76 +395,104 @@ def _encodeBody(self, body):
...
+
+            data = self._signalToContextDict(signal, notification)
+
+            # Check for email recipients in the event
+            details = data['evt'].details
+            mail_targets = details.get('recipients', '')
+            mail_targets = [x.strip() for x in mail_targets.split(',') if x.strip()]
+            if len(mail_targets) > 0:
+                log.debug("Adding recipients defined in the event %s", mail_targets)
+                targets |= set(mail_targets)
```
jpeacock-zenoss: Does this line, `targets |= set(mail_targets)`, work?

According to the `_targetsByTz` method, it looks like `targets` is a list and there is no `|=` operator defined for `list` and `set`.

You should add a test where there's an event having details containing a 'recipients' field.

Products/ZenModel/actions.py
```diff
@@ -394,76 +395,104 @@ def _encodeBody(self, body):
...
+
+            if signal.clear:
+                log.debug('This is a clearing signal.')
```
jpeacock-zenoss: This debug log isn't going to be useful. What event is a clearing signal and why does that matter?
Have a look at https://sites.google.com/a/zenoss.com/engineering/home/faq/logging-guidelines for how to write log messages.

Products/ZenModel/actions.py
rderkach: Tested it, and it will raise an exeption, now using `set` in `_targetsByTz` method.

Products/ZenModel/actions.py
rderkach: Yes, right.

Products/ZenModel/actions.py
```diff
@@ -394,76 +395,110 @@ def _encodeBody(self, body):
...
+        original_fst = signal.event.first_seen_time
+        original_sct = signal.event.status_change_time
+        for target_timezone, targets in targets_by_timezones.iteritems():
+            # Convert timestamp to user timezone
+            signal.event.last_seen_time = self._adjustToTimezone(
+                original_lst, target_timezone)
+            signal.event.first_seen_time = self._adjustToTimezone(
+                original_fst, target_timezone)
+            signal.event.status_change_time =  self._adjustToTimezone(
+                original_sct, target_timezone)
```
jpeacock-zenoss: Is there any use for `original_lst`, `original_fst`, and `original_sct` beyond this point?  If not, there's no point in creating these variables.  The `*_time` attributes of the event are being replaced with timezone adjusted values so it's perfectly fine to have this:
```
    signal.event.last_seen_time = self._adjustToTimezone(
        signal.event.last_seen_time, target_timezone
    )
```
It's also fewer variables to think about (thus lower cognitive load) when reading the code.

rderkach: The use of those `original_lst`, `original_fst`, `and original_sct` is that after first iteration in `for` I will convert the `signal.event.last_seen_time` so it will be a timestamp in new timezone, in the next iteration I will work with it again, the one in a timezone, so if I will move from it I will get a bad time for other timezones.
For converting to different timezones timestamp has to be in UTC.

PR #1843, opened 2016-08-29
ZEN-24265: Having Community/Custom ZenPacks break upgrade

ssOlexBaiko:
The reason the issue described in  ZEN-24265 exist is that a Zenoss upgrade replaces the zenoss docker image. The source and binaries for all ZenPacks exist in /opt/zenoss/ZenPacks which is in that zenoss docker image. Many persistent objects stored in ZODB are instances of classes described in the ZenPack source code. Without the source code available, those persistent objects are broken in all sorts of ways.
The fix is laying the ZenPack files down in the image first, then doing the full pack installs.


ssOlexBaiko:
retest this

ssOlexBaiko:
I 'd like to discuss this PR

Products/ZenUtils/zenpack.py
```diff
@@ -481,6 +481,10 @@ def doRestore(pack):
         fixedSomething = len(sortedPacks) > 0
         while len(sortedPacks) > 0:
             packListLen = len(sortedPacks)
+            # Load zenpacks files before normal installation to avoid
+            # upgrade breaking (ZEN-24274).
+            for packs in sortedPacks:
+                self._loadFiles(packs, zpsToRestore[packs][0])
```
jpeacock-zenoss: I was confused a bit by `packs`.  It's actually a `pack` (singular).  So a variable rename is in order here.

ssOlexBaiko: I`ll change it

jpeacock-zenoss: This `for` loop and the outer `while` loop are both iterating over `sortedPacks`.  I think this will result in many repetitive files-only installs of ZenPacks.  Perhaps the files-only install of `sortedPacks` should occur before the `while` loop.

Products/ZenUtils/zenpack.py
```diff
@@ -528,10 +532,20 @@ def _linkedRestore(self, zp):
...
+    def _loadFiles(self, zenpackID, zenpackVersion):
+        """
+        Install all zenpacks with '--files-only' flag for preventing
+        DistributionNotFound Error.
+        """
+        zenpack = self._findEggs(zenpackID, zenpackVersion)[0]
+        if zenpack and zenpack.lower().endswith(".egg"):
+            self.log.info("Loading files for %s", zenpack)
+            cmd = ["zenpack", "--files-only", "--install", zenpack]
+            subprocess.check_call(cmd)
```
jpeacock-zenoss: Since no database work is required here, is it possible to just directly call the code that performs the files-only installs rather than running the zenpack command in subshell?

Products/ZenUtils/zenpack.py
```diff
@@ -479,6 +479,17 @@ def doRestore(pack):
...
         fixedSomething = len(sortedPacks) > 0
+        # Installing all zenpacks with '--files-only' flag
+        # for preventing DistributionNotFound Error
+        for pack in sortedPacks:
+            candidate = self._findEggs(pack, zpsToRestore[pack][0])
+            try:
+                EggPackCmd.InstallEggAndZenPack(
+                    self.dmd,
+                    candidate[0],
+                    filesOnly=True)
```
jpeacock-zenoss: It's OK to have all three arguments on the same line.  It should still be readable.  E.g.
```
    EggPackCmd.InstallEggAndZenPack(
        self.dmd, candidate[0], filesOnly=True)
```

Products/ZenUtils/zenpack.py
```diff
@@ -479,6 +479,17 @@ def doRestore(pack):
...
+            except (OSError,) as e:
```
jpeacock-zenoss: Wrapping the exception type is unnecessary here.
```except OSError as e:``` is sufficient.

PR #1853, opened 2016-09-06
ZEN-24692 Cannot use ssh keys for connections

romanbakaleyko:


romanbakaleyko:
demo https://jira.zenoss.com/secure/attachment/34518/zenmodeler_output

jplouis:
SSHUserAuthClient no longer supports handling string representation of keys. Deprecated in twisted in 9.x and functionality to dropped in 16.x.

PR #1862, opened 2016-09-08
add zport isup endpoint

rDr4g0n:


rDr4g0n:
ed is super mean and he made me cry, so im redoing this.

rDr4g0n:
the above is DEFINITELY the reason why

PR #1867, opened 2016-09-14
Adds Parent Device to Component Details View

alam-zenoss:


alam-zenoss:
Yes, you're quite right. Performance would certainly affected when requiring the _entire_ device. It'll probably be better if I were to create a new attribute in infos called 'deviceName' and just return a `<object>.device().getId()` or maybe get a direct link to the overview of the Device like we do for Resource for events. I'll make the change.

Products/Zuul/infos/component/__init__.py
```diff
@@ -28,6 +28,10 @@ def device(self):
         return self._object.device()
 
     @property
+    def deviceName(self):
+        return self._object.device().getId()
```
iancmcc: If you legitimately want this to be the name, you should do `self._object.device().titleOrId()`

PR #1868, opened 2016-09-15
ZEN-23593

ntzen:
Added event creation for when the snmpclient runs into exceptions to better alert customers to situations where they might not be monitoring their devices as they expected to be and it isn't clear enough on the device detail page that modelling failed


jplouis:
A different fix is being considered.

PR #1869, opened 2016-09-15
ZEN-22313 Maria DB Pool allocation update

mayathegreat:
Added migration script to zenmodel migration directory
Added test for the above under the migration/tests directory


Products/ZenModel/migrate/updateMariaDbPoolAllocations.py
```diff
@@ -0,0 +1,43 @@
...
+import logging
+import Migrate
+import os
+import servicemigration as sm
+
+log = logging.getLogger("zen.migrate")
+sm.require("1.0.0")
+
+class UpdateMariaDBPoolAlloc(Migrate.Step):
+    version = Migrate.Version(5, 1, 0)
```
iancmcc: This should be `(5, 2, 0)` if this is a Metis ticket

Products/ZenModel/migrate/updateMariaDbPoolAllocations.py
```diff
@@ -0,0 +1,43 @@
...
+        log.info("Found {0} services with 'mariadb' in their service path".format(len(mariaservices)))
+        for m in mariaservices:
+            log.info("{0}".format(m))
+            cfs = filter(lambda f: f.name == "/etc/my.cnf", m.originalConfigs + m.configFiles)
+            for cf in cfs:
+                lines = cf.content.split('\n')
+                for i, line in enumerate(lines):
+                    if line.startswith("innodb_buffer_pool_size"):
+                        lines[i] = line.replace("512M", "{{percentScale .RAMCommitment 0.8}}")
+                        log.info("Changed MariaDB innodb_buffer_pool_size from 512M (default) to '{{percentScale .RAMCommitment 0.8}}' on line {0}".format(i))
```
iancmcc: Before logging this, check to see if you actually did change the line, maybe? Or better yet, check if `"512M" in line` first. This is perfectly safe, of course, but no point logging if it didn't actually make a replacement.

Products/ZenModel/migrate/updateMariaDbPoolAllocations.py
mayathegreat: Ah, yes it's for Metis. I'll make the change.

Products/ZenModel/migrate/updateMariaDbPoolAllocations.py
mayathegreat: good point, I'll make the change :)

PR #1870, opened 2016-09-16
Feature/zen 21558

romanbakaleyko:


romanbakaleyko:
demo screens in the JIRA https://jira.zenoss.com/browse/ZEN-21558

Products/Zuul/routers/manufacturers.py
```diff
@@ -51,8 +51,8 @@ def editProduct(self, params=None):
         """
         Edit a product
         """
-        oldData = self.getProductData(params['uid'], params['oldname'])
         facade = self._getFacade()
+        oldData = facade.getProductData(params['uid'], params['oldname'])[0]
```
glujan: I don't understand this change. Is it because you want to avoid marshalling data inside `self.getProductData`)?

Products/Zuul/facades/manufacturersfacade.py
```diff
@@ -194,7 +195,7 @@ def editManufacturer(self, params):
         """
         edit the information for a given manufacturer
         """
-        manufacturer = self._getObject("/zport/dmd/Manufacturers/"+params['name'])
+        manufacturer = self._getObject("/zport/dmd/Manufacturers/"+params['oldname'])
```
glujan: Won't it break if name contains some naughty symbol, like `/`?

romanbakaleyko: UI form doesn't allow '/' symbol.

Products/Zuul/routers/manufacturers.py
romanbakaleyko: it returns DirectResponse( data ) what I don't need.

PR #1877, opened 2016-09-23
migration for fix/ZEN-24578 ensures Pagespeed mod for nginx is enabled

corntoole:


Products/ZenModel/migrate/tests/test_enableNginxPagespeed.py
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env python
```
rphillipsz: This looks like a duplicate of test_EnableNginxPagespeed.py

corntoole: The dupe has been removed.

PR #1883, opened 2016-09-28
ZEN-22313 Remove uintToInt from Migration script

carlcamera:
Update migration script removing reference to evil uintToInt macro.  PS: not really evil - it's just that we cannot guarranty its presence after migration.


carlcamera:
retest this.

PR #1885, opened 2016-09-29
ZEN-24813: pull build number from build system and surface in the UI.

ciphero:


corntoole:
Looks fine. Just update your branch.

makefile
```diff
@@ -1,4 +1,5 @@
 VERSION  := 5.2.0
```
tgjonestx: I'm a little rusty on make-isms, but I think this needs to be  `VERSION ?= 5.2.0` so that `VERSION` can be overridden with command line arguments

ciphero: Ok, will do

corntoole: http://www.gnu.org/software/make/manual/make.html#Flavors

PR #1894, opened 2016-10-05
ZEN-23767 Maintenance Window UI does not recognize custom Production …

romanbakaleyko:
…States


romanbakaleyko:
demo in attached screenshot in the https://jira.zenoss.com/browse/ZEN-23767

PR #1897, opened 2016-10-06
ZEN-25294 zenpack create fails in multihost

ed-:
Serviced zenpack create was failing in with the error 'easy-install.pth:
Resource temporarily unavailable'. This commit adds a touch immediately
before the call to setup.py develop to wake up the NFS so the file is
available in time to modify the file.


ed-:
Not ready just yet

PR #1900, opened 2016-10-07
ZEN-23882 Reflected Cross-Site Scripting (XSS)

romanbakaleyko:


romanbakaleyko:
I am not 100% sure about this change, could be that in same cases we don't want to escape html in flares. If so, we can do this optionally.

PR #1902, opened 2016-10-07
Feature/ZEN-25376

spopezen:
https://jira.zenoss.com/browse/ZEN-25376

Removal of the /services/:serviceid/status endpoint caused the queryServiceStatus to break when using CC 1.2.

This PR fixes that by using the new endpoint from the CC V2 api when necessary, while maintaining backwards compatibility with CC 1.1.x.


Products/ZenUtils/controlplane/client.py
```diff
@@ -222,12 +226,57 @@ def queryServiceInstances(self, serviceId):
...
+        """
+        if not self._useV2:
+            try:
+                response = self._dorequest("/services/%s/status" % serviceId)
+                body = ''.join(response.readlines())
+                response.close()
+                decoded = ServiceStatusJsonDecoder().decode(body)
+            except urllib2.HTTPError as ex:
+                if ex.getcode() == 404:
+                    self._useV2 = True
```
fraibacas: Let's set this in the constructor since we will need it for other missing/modified endpoints

PR #1904, opened 2016-10-10
Convert snmpV1SpecificTrap to string when concatenating.

jplouis:


glujan:
I don't understand why this change is needed

PR #1906, opened 2016-10-11
ZEN-24535: convert event time in email to timezone from user property

rderkach:
This is port from : https://github.com/zenoss/zenoss-prodbin/pull/1834

[Jira](https://jira.zenoss.com/browse/ZEN-20623)

### DIFF
--- a/Products/ZenModel/actions.py
+++ b/Products/ZenModel/actions.py
@@ -50,6 +50,7 @@ from Products.ZenUtils.IpUtil import getHostByName, isip
 from Products.ZenUtils.guid.guid import GUIDManager
 from Products.ZenUtils.ProcessQueue import ProcessQueue
 from Products.ZenUtils.ZenTales import talEval, InvalidTalesException
+from Products.ZenUtils.Time import convertTimestampToTimeZone, isoToTimestamp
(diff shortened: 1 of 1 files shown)


Products/ZenModel/actions.py
```diff
@@ -392,77 +393,111 @@ def _encodeBody(self, body):
...
+        Take timezone from user property to convert a event time in
+        notification and also group targets emails by those timezones.
+        """
+        targets_by_timezones = {}
+        for user in dmd.ZenUsers.getAllUserSettings():
+            if user.email in targets:
+                if targets_by_timezones.get(user.timezone):
+                    targets_by_timezones[user.timezone].add(user.email)
+                else:
+                    targets_by_timezones[user.timezone] = set([user.email])
```
glujan: Instead of this inner `if ... else` you could use: `targets_by_timezones.setdefault(user.timezone, set()).add(user.email)`

glujan: Also, `targets_by_timezones` is quite a longish name, I would stick with `tz_targets` or something like that

rderkach: Thanks. I will keep in mind.

PR #1910, opened 2016-10-12
ZEN-23950: Fix static file resources to be nicer with PageSpeed.

iancmcc:
This also provides a view that zproxy can use to map PageSpeed static file loads directly to the filesystem, freeing up threads in ZServer that would ordinarily be spent serving static files.


iancmcc:
Tests pass on my box. Confirmed with JP before merging. The build system is having significant issues at the moment.

PR #1918, opened 2016-10-21
added migrate to add kibana endpoint

sockterrier:


Products/ZenModel/migrate/addKibanaRouteInZproxyConf.py
```diff
@@ -0,0 +1,78 @@
...
+    def _update_kibana_config(self, zproxy_config, old_route_string):
+        """
+        Find the old kibana route and replace it with the new one
+        @param old_route_string: old route string found in the config file
+        """
+        new_content = ""
+        start = zproxy_config.content.find(old_route_string)
+        end = zproxy_config.content.find("}", start)
+        new_content = zproxy_config.content[:end+1] + \
+                      self.KIBANA_NEW_ROUTE_CONFIG + \
```
fraibacas: This is removing the 1.1.x route for kibana. We need to add the 1.2 route only if it has not already been added

PR #1920, opened 2016-10-24
ZEN-21868 Clicking on graphs bar are not honoring the selected filter…

romanbakaleyko:
… fields in the event console tab opened

### DIFF
--- a/Products/ZenUI3/browser/resources/js/zenoss/EventPanels.js
+++ b/Products/ZenUI3/browser/resources/js/zenoss/EventPanels.js
@@ -1490,15 +1490,22 @@
             this.refresh();
         },
         restoreURLState: function() {
-            var qs = window.location.search.replace(/^\?/, ''),
-            state = Ext.urlDecode(qs).state;
-            if (state) {
+            var qs = window.location.search.replace(/^\?/, ''), state;
+            var decoded = Ext.urlDecode(qs);
+
+            if (decoded.state) {
                 try {
-                    state = Ext.decode(Zenoss.util.base64.decode(decodeURIComponent(state)));
-                    this.applyState(state);
-                    this.filterRow.storeSearch();
+                    state = Ext.decode(Zenoss.util.base64.decode(decodeURIComponent(decoded.state)));
                 } catch(e) { }
+            //in case parameters are not encoded
+            } else {
+                state = {"filters": decoded};
             }
+
+            this.clearFilters();
+            this.applyState(state);
+            this.filterRow.storeSearch();
+            
         },
         clearURLState: function() {
             var qs = Ext.urlDecode(window.location.search.replace(/^\?/, ''));


Products/ZenUI3/browser/resources/js/zenoss/EventPanels.js
```diff
@@ -1498,6 +1498,11 @@
                     this.applyState(state);
                     this.filterRow.storeSearch();
                 } catch(e) { }
+            // for cases with not encoded parameters
+            } else if (qs) {
+                state = {"filters": Ext.urlDecode(qs)};
```
rDr4g0n: This looks like an already existing issue, but `state` is never declared. This means by default it is being declared and defined as a global variable, which can introduce issues if some other code touches a `state` global variable as well.

rDr4g0n: `Ext.urlDecode(qs)` is called here and also just above. it looks like this code could be made more readable by first decoding and saving the result, then checking for existence of `state` on the decoded object.

Products/ZenUI3/browser/resources/js/zenoss/EventPanels.js
```diff
@@ -1498,6 +1498,11 @@
...
+                this.applyState(state);
+                this.filterRow.storeSearch();
```
rDr4g0n: lines 1504 and 1505 are identical to 1498 and 1499. This code can be modified so that the if/else stuff is only for the purpose of getting and validating the `state` variable (and handling exceptions in that validation), then the code flows through to apply the state and update the store.

Products/ZenUI3/browser/resources/js/zenoss/EventPanels.js
```diff
@@ -1490,11 +1490,20 @@
             this.refresh();
         },
         restoreURLState: function() {
-            var qs = window.location.search.replace(/^\?/, ''),
-            state = Ext.urlDecode(qs).state;
-            if (state) {
+            var qs = window.location.search.replace(/^\?/, ''), state, decState;
+            
+            if (qs) {
                 try {
```
rDr4g0n: This `try` statement is now very large. it should be as minimal as possible. the reason it was present initially was to catch errors from `Zenoss.util.base64.decode`, so it needs to be minimized to that. This big `try` block could be especially frustrating since the `catch` block ignores the error, so whoever is debugging this would have a very hard time finding the problem.

Products/ZenUI3/browser/resources/js/zenoss/EventPanels.js
```diff
@@ -1490,11 +1490,20 @@
...
-                    state = Ext.decode(Zenoss.util.base64.decode(decodeURIComponent(state)));
+                    if (Ext.urlDecode(qs).state) {
```
rDr4g0n: `Ext.urlDecode(qs)` is being called 3 times now. It makes more sense to call it once with something like `var decoded = Ext.urlDecode(qs);` *then* check for presence of the `state` field with something like `if(decoded.state){ ... }`.

Products/ZenUI3/browser/resources/js/zenoss/EventPanels.js
```diff
@@ -1490,11 +1490,20 @@
...
+                        decState = Ext.urlDecode(qs).state;
+                        state = Ext.decode(Zenoss.util.base64.decode(decodeURIComponent(decState)));
```
rDr4g0n: wrap just this line with `try...catch`

Products/ZenUI3/browser/resources/js/zenoss/EventPanels.js
```diff
@@ -1490,11 +1490,20 @@
...
+                    //in case parameters are not encoded
+                     } else {
```
rDr4g0n: theres an extra space on the indent for this line

Products/ZenUI3/browser/resources/js/zenoss/EventPanels.js
```diff
@@ -1490,15 +1490,22 @@
...
+            var decoded = Ext.urlDecode(qs);
+
+            if (decoded.state) {
                 try {
-                    state = Ext.decode(Zenoss.util.base64.decode(decodeURIComponent(state)));
-                    this.applyState(state);
-                    this.filterRow.storeSearch();
+                    state = Ext.decode(Zenoss.util.base64.decode(decodeURIComponent(decoded.state)));
                 } catch(e) { }
+                //in case parameters are not encoded
```
jpeacock-zenoss: Isn't this comment is related to the `else` clause?  Presently it looks like it's commenting code missing from the 'then' clause.

PR #1924, opened 2016-10-27
ZEN-25463 Collector config cycle refresh results in loss of ping status

romanbakaleyko:


romanbakaleyko:
The problem appears after updating configuration for a particular device/devices by adding changes to the configuration by hand (e.g. creating/removing data source) or by scheduled updating configuration according to `Config Cycle Interval`. During configuration update daemon removes old tasks and create new according to a new configuration (at that moment we loose task's status) so until new maintenance performs https://github.com/zenoss/zenoss-prodbin/blob/develop/Products/ZenCollector/daemon.py#L705 daemon tries to collect even if device is down.

Products/ZenCollector/daemon.py
```diff
@@ -584,6 +584,10 @@ def _updateConfig(self, cfg):
             # all pending tasks have completed
             if not self.options.cycle:
                 self._pendingTasks.append(taskName)
+        # put tasks on pause after configuration update to prevent unnecessary collections ZEN-25463
+        for devId in self._unresponsiveDevices:
```
jplouis: This seems excessive. This method is for updating one config and the for loop is modifying every task that is paused.

PR #1925, opened 2016-10-28
ZEN-25579: add migration for new HMaster and RegionServer graphs

rderkach:
This is migration script and tests for https://github.com/zenoss/zenoss-service/pull/327 and https://github.com/zenoss/zenoss-service/pull/281

[Jira](https://jira.zenoss.com/browse/ZEN-25579)

### DIFF
--- a/Products/ZenModel/migrate/AddHMasterRegionServerGraphConfigs.py
+++ b/Products/ZenModel/migrate/AddHMasterRegionServerGraphConfigs.py
@@ -1,0 +1,490 @@
+##############################################################################
+#
+# Copyright (C) Zenoss, Inc. 2015, all rights reserved.
+#
+# This content is made available according to terms specified in
+# License.zenoss under the directory where your Zenoss product is installed.
+#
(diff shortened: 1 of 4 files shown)


Products/ZenModel/migrate/AddHMasterRegionServerGraphConfigs.py
```diff
@@ -0,0 +1,490 @@
...
+                            )
+                        ]
+                    )
+                )
+            else:
+                log.info("opcounts graph found; skipping.")
+
+            if not filter(lambda x: x.ID == 'RegionServer', mcs):
+                log.info("No RegionServer internal metrics found; creating.")
+                commit = True
```
ssOlexBaiko: I believe that `commit = True` should be after append() method, move it to the 481 line
similar in other parts.

rderkach: Yep, it will look better.

PR #1928, opened 2016-10-28
ZEN-25688: Failed to gather Support Bundle (Metis)

svambati:


Products/ZenUtils/elastic/client.py
```diff
@@ -38,8 +38,27 @@ def __init__(self):
...
         """
         self.session = None
+        self.cc_version = ""
+        if self._checkHothOrNewer():
+            self.ccURL = 'http://127.0.0.1:443'
+        else:
+            self.ccURL = 'https://127.0.0.1'
         self.login()
 
+    def _checkHothOrNewer(self):
```
fraibacas: We should make this method "public" in controlplane/client.py and reuse it

Products/ZenUtils/elastic/client.py
```diff
@@ -38,8 +38,27 @@ def __init__(self):
         used across all requests that the client makes.
         """
         self.session = None
+        self.cc_version = ""
+        if self._checkHothOrNewer():
```
fraibacas: We should probably create a method in controlplane.py that returns the CC url depending on the cc version

PR #1930, opened 2016-10-29
Feature/zen 25688 1

svambati:


Products/ZenUtils/elastic/client.py
```diff
@@ -38,6 +38,12 @@ def __init__(self):
         used across all requests that the client makes.
         """
         self.session = None
+        self.cc_version = ""
```
fraibacas: What is cc version used for?

Products/ZenUtils/controlplane/client.py
```diff
@@ -71,24 +81,10 @@ def __init__(self, user, password, host=None, port=None):
         self._creds = {"username": user, "password": password}
         self._netloc = "%(host)s:%(port)s" % self._server
         self.cc_version = ""
```
fraibacas: we need self.cc_version bc it is injected in the UI

Products/ZenUtils/controlplane/client.py
```diff
@@ -30,7 +30,18 @@
 
 
 SERVICED_VERSION_ENV = "SERVICED_VERSION"
-
+    
+def checkHothOrNewer():
```
fraibacas: This method should return the cc version extracted from the env variable SERVICED_VERSION_ENV or "1.1.x" if the variable is not found

Products/ZenUtils/controlplane/client.py
fraibacas: call the method that returns the cc version and assign it to self.cc_version

Products/ZenUtils/controlplane/client.py
```diff
@@ -71,24 +81,10 @@ def __init__(self, user, password, host=None, port=None):
         self._creds = {"username": user, "password": password}
         self._netloc = "%(host)s:%(port)s" % self._server
         self.cc_version = ""
-        self._hothOrNewer = self._checkHothOrNewer()
+        self._hothOrNewer = checkHothOrNewer()
```
fraibacas: _hothOrNewer would be true if self.cc_version != "1.1.X"

Products/ZenUtils/elastic/client.py
```diff
@@ -38,6 +38,12 @@ def __init__(self):
...
+        self._hothOrNewer = checkHothOrNewer()
```
fraibacas: With your latest change, checkHothOrNewer returns a tuple which will make self._hothOrNewer evaluate to true since it will not be empty

PR #1936, opened 2016-11-01
ZEN-25845 Change ZenModel migration numbering scheme

ed-:
Renumbering all migrations at least as young as 5.1.1 because
some releases and appliances had db versions ahead of the migrations
shipped.

Switching to a single-number scheme to reduce confusion and to break
the relation between releases of Zenoss and migrations of ZenModel--
they need not stay in lockstep.

The new scheme is based on 100 to make sure releases as early as 5.1.1
will run all of the necessary migrations, as the db version could be as
far forward as 5.1.70 in some systems.


ed-:
Going to resubmit this one under a new ticket number.

PR #1938, opened 2016-11-02
port of fix from RPS

sockterrier:


sockterrier:
this is a port of this change that didn't go into develop first:

https://github.com/zenoss/4.2.5-RPS/pull/705/files

PR #1943, opened 2016-11-03
ZEN-25506 Use user local time in date pickers

glujan:
[Jira](https://github.com/zenoss/zenoss-prodbin/compare/ZEN-25506)

We must display user's local time in date pickers, not UTC.

### DIFF
--- a/Products/ZenUI3/browser/resources/js/zenoss/form/graphPanel.js
+++ b/Products/ZenUI3/browser/resources/js/zenoss/form/graphPanel.js
@@ -710,6 +710,10 @@
                 return;
             }
 
+            if (isAdjusted === undefined) {
+                isAdjusted = true;
+            }
+
             var adjustedTime;
 
             // if someone is using a date object, get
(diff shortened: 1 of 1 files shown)


Products/ZenUI3/browser/resources/js/zenoss/form/graphPanel.js
```diff
@@ -709,6 +713,7 @@
             if(!ms){
                 return;
             }
+            isAdjusted = isAdjusted || true;
```
glujan: Uhm. Unintentionally yes. Code works because argument is not being passed at all in some usages, so `isAdjusted` will be `undefined` and default to _true_.

Anyway, this is a bug, let me fix it.

glujan:
@achow-zenoss I've fixed bug found by you. Thanks!

I would like to wait with merge for @rDr4g0n comment as we had a talk about this PR in private.

rDr4g0n:
Sorry for the delay, I've had a hard time getting zenoss running with symlinked `Products` directory so that I can verify this code by editing/debugging.

PR #1945, opened 2016-11-04
ZEN-24535: add a timezone to event time in email

rderkach:
There was no timezone in in event time which may be confusing for
a user.
Fetch all time string in email body and add timezone to the end of
it.

### DIFF
--- a/Products/ZenModel/actions.py
+++ b/Products/ZenModel/actions.py
@@ -400,14 +400,11 @@ class EmailAction(IActionBase, TargetableAction):
         Take timezone from user property to convert a event time in
         notification and also group targets emails by those timezones.
         """
-        targets_by_timezones = {}
+        tz_targets = {}
         for user in dmd.ZenUsers.getAllUserSettings():
             if user.email in targets:
-                if targets_by_timezones.get(user.timezone):
-                    targets_by_timezones[user.timezone].add(user.email)
-                else:
-                    targets_by_timezones[user.timezone] = set([user.email])
-        return targets_by_timezones
+                tz_targets.setdefault(user.timezone, set()).add(user.email)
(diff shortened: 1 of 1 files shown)


Products/ZenModel/actions.py
```diff
@@ -400,14 +400,11 @@ def _targetsByTz(self, dmd, targets):
...
-        targets_by_timezones = {}
+        tz_targets = {}
         for user in dmd.ZenUsers.getAllUserSettings():
             if user.email in targets:
-                if targets_by_timezones.get(user.timezone):
-                    targets_by_timezones[user.timezone].add(user.email)
-                else:
-                    targets_by_timezones[user.timezone] = set([user.email])
-        return targets_by_timezones
+                tz_targets.setdefault(user.timezone, set([user.email])).add(user.email)
```
glujan: `tz_targets.setdefault(user.timezone, set()).add(user.email) ` would be better. You don't need to create set with user.email if you are going to try adding it in next instructions anyway.

Products/ZenModel/actions.py
```diff
@@ -453,6 +450,11 @@ def executeBatch(self, notification, signal, targets):
 
             log.debug('Sending this subject: %s', subject)
 
+            # Find all time strings in body and add timezone to it
+            times = set(re.findall('(\d+/\d+/\d+ \d+:\d+:\d+.\d+)', body))
```
glujan: Please use raw string for patterns.

Let's discuss this regex in private.

Products/ZenModel/actions.py
rderkach: Yes, missed that.

glujan:
You should squash comments, I believe

PR #1946, opened 2016-11-04
ZEN-23871: event export to csv information overload

fraibacas:
- Reverted changes that unsuccessfully tried to fix this ticket (https://github.com/zenoss/zenoss-prodbin/pull/1790). 
- The issue is that, when exporting to cvs, each event detail gets its own column. Events with lots of details (like trap events) generate cvs with so many columns that the file was impossible to be read. To address this, all the event details that are returned in the "details" key by zep are put in one column.
- Also, now columns are arranged differently. First, the columns displayed in the event console, then all the remaining fields in alphabetic order and last but not least the details.

Example in the ticket: https://jira.zenoss.com/browse/ZEN-23871

### DIFF
--- a/Products/ZenUI3/browser/eventconsole/export.py
+++ b/Products/ZenUI3/browser/eventconsole/export.py
@@ -14,7 +14,6 @@ Given a list of events to export, format them
 appropriately and then return back a string
 """
 
-import itertools
 import json
 import logging
 from datetime import datetime
@@ -28,7 +27,7 @@ from interfaces import IEventManagerProxy
 
 log = logging.getLogger('zen.eventexport')
 
-CSV_MAX_COLUMNS = 16000
+DETAILS_KEY = "details"
 
 class EventsExporter(BrowserView):
     def __call__(self):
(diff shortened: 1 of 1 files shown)


fraibacas:
The way I see it: there is a bug and an enhancement request. This PR fixes the bug. If we want to change the way we export events, we can create an enhancement ticket and do it which should not be a p1.

fraibacas:
Enhancement ticket: https://jira.zenoss.com/browse/ZEN-23871

dougsyer:
In my opinion just have an export option thay includes detsiks and one that doesnt.  Sometimes i want the details as columns but usually not.  Id make the default only include details that you can see in the default event consoke colums plus any indexed fields added by zenpacks..  


The defauly should be less detaiks because i want it to be easy for my end users

dougsyer:
By default i mean all the selectable fields.  Its not important imo to just export columns that are added or subttracted by users just add all selectable fields including indexed fields added by zps.

99% of the time i throw it in a pivot table anyways.

And it would be nice to clean up sone of the fields that export out o have a cleaned example if you want.  

Also id delimit with something other than commas probably pipes are bestbut now im being picky. Inse alot of commas in event messages but not many pipes but again thats picky

PR #1957, opened 2016-11-10
ZEN-22199: Test datasource UI functionality should test on device's actual collector

olst:


olst:
In fact, if device is on remote collector I use zminion to execute ssh command with /opt/zenoss/bin/runRemoteSsh.py script.

https://github.com/zenoss/zenoss-prodbin/pull/1957/files#diff-4ad762b992ef65f2f56ded6d457918e8R254

-> command = self.compile(RemoteSshCommand(self.command, device), device)
(Pdb) n
> /opt/zenoss/Products/ZenModel/BasicDataSource.py(256)testDataSourceAgainstDevice()
-> executeStreamCommand(command, write)
(Pdb) command
'zminion --minion-name zminion_remote run -- "/opt/zenoss/bin/runRemoteSsh.py 
--cmd=\'/usr/bin/env uptime\' --device_name=ip-10-111-23-84.zenoss.loc --host=10.111.23.84 
--username=emVubnk=   --password=dG9ybmFkbzIy --port=22 --login_tries=1 --search_path=[] 
--existance_test=\'test -f %s\'  --login_timeout=10.0 --command_timeout=15.0 
--key_path=~/.ssh/id_dsa --concurrent_sessions=5"'
(Pdb) 
We use executeSshCommand method to run ssh commands:
https://github.com/zenoss/zenoss-prodbin/pull/1957/files#diff-4ad762b992ef65f2f56ded6d457918e8L237

I use runRemoteSsh.py script because I can't use executeSshCommand method (or any script that uses that method) on remote collector. It's not possible to run zminion on remote collectors with commands that need zope, as collectors don't have zodb access.

olst:
I made changes for encrypting credentials

iancmcc:
More to the point, an XOR cipher is inherently insecure unless you're using a random key every time.

jplouis:
Some notes were added in the ticket.  If the proposed solution in the ticket isn't possible than a please let comment in the ticket.

PR #1958, opened 2016-11-10
ZEN-22878 Escape names in search results

glujan:
[Jira](https://jira.zenoss.com/browse/ZEN-22878)

I had to reindex catalog to make it work.

### DIFF
--- a/Products/Zuul/catalog/global_catalog.py
+++ b/Products/Zuul/catalog/global_catalog.py
@@ -8,6 +8,7 @@
 ##############################################################################
 
 
+from cgi import escape
 from itertools import ifilterfalse, chain
 
 import zExceptions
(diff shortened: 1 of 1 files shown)


romanbakaleyko:
I would escape it in the UI side, reindexing can be very painfull for some customers.

glujan:
Whole HTML snippet is send to the client side, so it is not possible to escape it in the UI.

We may, of course, try to escape some _dangerous_ tags on the UI but I don't think it's an elegant solution.

rDr4g0n:
Reindexing may be necessary in the case that some unescaped values have already made it in, but this PR's purpose is to patch the hole rather than retroactively fix it.

glujan:
@rDr4g0n Would you mind giving a review to this PR? I don't mind any type of review 😉

PR #1962, opened 2016-11-11
ZEN-23195: In case of a sequence conflict, resequence to avoid dupes.

iancmcc:


dougsyer:
im a 4 user but to be clear what is dangerous for us is when we are forced to delete a whole catalog and have to wait especially during an upgrade for it ti rebuild..especially component search.  It only takes ine unexpected issue sometimes to pop up and if it takes you 3 hours to reindex and it fails at 2:30 you are in big big trouble and it can happen even adter running toolbox tools.

But reindexing while we are up and running in general is no big deal.  And in 4 the lucenne deletion is no big deal because status is done quick ..

But again deleting a catalog entirely and basically being down during a reinddex is terrifying.

May not be relevant here but wanted to weigh in on a subtle but important distinction for us at least.

PR #1990, opened 2016-11-18
add 1s to first_seen events filter to make it's interval (TODATE, FROMDATE]

corntoole:


Products/Zuul/routers/zep.py
```diff
@@ -518,9 +518,11 @@ def _buildFilter(self, uids, params, specificEventUuids=None, includeContextInUi
                 'message': params.get('message'),
             }
             parsed_params = self._filterParser.parseParams(params)
-            # ZEN23418: Add 1 sec to last_seen range filter,
-            # so that it includes {events: event['last_seen'] <= params['lastTime']}
+            # ZEN23418: Add 1 sec to first_seen and last_seen range filters,
+            # so that it includes {events: event['{first|last}_seen'] <= params['{firstTime|lastTime}']}
             filter_params.update(parsed_params)
+            if filter_params['first_seen'] is not None and len(filter_params['first_seen']) == 2:
+                filter_params['first_seen'][1] = filter_params['first_seen'][1]+1000
```
cchristiezen: Would you not need to subtract a second from the first_seen to get the events >= first_seen.

Also, rather than add a second to do the inclusion is there a comparison that could be changed instead?

corntoole: I haven't been able to find where the filters are applied, but it appears that it inclusive on the first_seen end of the range.

PR #1996, opened 2016-12-01
fixes ZEN-26256: User and Time Stamp for Event Log Entries Span Multi…

HIMix:
…ple Lines


glujan:
I can see that solution was described by reporter in Jjira (and he gave screenshots already) but still, you've missed a ticket number

PR #1997, opened 2016-12-01
fixes ZEN-25962: Enter Key does not Add User Log Entries

HIMix:


romanbakaleyko:
This solution doesn't work for me.

PR #1999, opened 2016-12-05
ZEN-25826 Add date format to user settings

glujan:
[Jira](https://jira.zenoss.com/browse/ZEN-25826)

In this PR I've added a new setting (on an user level) - user can now specify a date format which will be displayed on graphs.

Also, I've changed order of loading scripts on page - I need to have `ZenossData` JS snippet before `visualization.js` for zenoss/query#264

### DIFF
--- a/Products/ZenModel/skins/zenmodel/editUserSettings.pt
+++ b/Products/ZenModel/skins/zenmodel/editUserSettings.pt
@@ -218,6 +218,29 @@
             </script>
           </td>
         </tr>
+        <tr>
+          <td class="tableheader" align="left">Date format</td>
+          <td class="tablevalues">
+            <select id="user_settings_dateformat" name="dateFormat"></select>
+            <script tal:content='string:
+                 Zenoss.env.CURRENT_USER_DATE_FORMAT = "${context/dateFormat}" || Zenoss.USER_DATE_FORMAT;
+             '></script>
+            <script>
+              // build a select list with the current date format selected.
+              // Zenoss.data.timeZones comes from moment-timezone-data.js and
+              Ext.onReady(function() {
+                  var select = Ext.get('user_settings_dateformat').dom, formats = ['MM/DD/YY', 'DD/MM/YY', 'YY/MM/DD'], format, i;
+                  for (i = 0; i < formats.length; i++) {
+                      format = formats[i];
+                      select.options[i] = new Option(format, format);
+                      if (format === Zenoss.env.CURRENT_USER_DATE_FORMAT) {
+                          select.options[i].selected = true;
+                      }
+                  }
+             });
+            </script>
+          </td>
+        </tr>
 
         <tr>
             <td style="height:1px;border-bottom:dashed 1px #bbb;" class="tableheader">
--- a/Products/ZenUI3/browser/configure.zcml
+++ b/Products/ZenUI3/browser/configure.zcml
@@ -398,7 +398,8 @@
 
 <browser:viewlet
     name="zenoss_data"
-    manager=".interfaces.IMainSnippetManager"
+    weight="3"
+    manager=".interfaces.IJavaScriptSrcManager"
     class=".javascript.ZenossData"
     permission="zope2.Public"
     />
--- a/Products/ZenUI3/browser/javascript.py
+++ b/Products/ZenUI3/browser/javascript.py
@@ -251,13 +251,19 @@ class ZenossData(JavaScriptSnippet):
         # 3. the timezone of the browser
         user = self.context.dmd.ZenUsers.getUserSettings()
         timezone = user.timezone
+        date_fmt = user.dateFormat
         snippet = """
+          (function(){
+            Ext.namespace('Zenoss.env');
+
             Zenoss.env.COLLECTORS = %r;
             Zenoss.env.priorities = %r;
             Zenoss.env.productionStates = %r;
             Zenoss.USER_TIMEZONE = "%s" || jstz.determine().name();
-        """ % ( collectors, priorities, productionStates, timezone )
-        return snippet
+            Zenoss.USER_DATE_FORMAT = "%s" || "MM/DD/YY";
+          })();
+        """ % ( collectors, priorities, productionStates, timezone, date_fmt )
+        return SCRIPT_TAG_TEMPLATE % snippet
 
 class BrowserState(JavaScriptSnippet):
     """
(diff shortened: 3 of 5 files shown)


Products/ZenModel/skins/zenmodel/editUserSettings.pt
```diff
@@ -218,6 +218,29 @@
...
+            <select id="user_settings_dateformat" name="dateFormat"></select>
+            <script tal:content='string:
+                 Zenoss.env.CURRENT_USER_DATE_FORMAT = "${context/dateFormat}" || Zenoss.USER_DATE_FORMAT;
+             '></script>
+            <script>
+              // build a select list with the current date format selected.
+              // Zenoss.data.timeZones comes from moment-timezone-data.js and
+              Ext.onReady(function() {
+                  var select = Ext.get('user_settings_dateformat').dom, formats = ['MM/DD/YY', 'DD/MM/YY', 'YY/MM/DD'], format, i=0;
+                  for (format of formats) {
```
rDr4g0n: `for...of` is a brand new js feature that will not work on many browsers. Also, you need a counter `i` here, so it makes more sense to use something that provides that. You probably want `Array.forEach` which provides each element and its index in the array.

Products/ZenUI3/browser/javascript.py
```diff
@@ -251,13 +251,18 @@ def snippet(self):
...
         snippet = """
-            Zenoss.env.COLLECTORS = %r;
-            Zenoss.env.priorities = %r;
-            Zenoss.env.productionStates = %r;
-            Zenoss.USER_TIMEZONE = "%s" || jstz.determine().name();
-        """ % ( collectors, priorities, productionStates, timezone )
-        return snippet
+            var z = Ext.ns('Zenoss'),
+                env = Ext.ns('Zenoss.env');
+
```
rDr4g0n: Why are you explicitly loading the Ext `Zenoss` namespace var here instead of using the global `Zenoss` variable as it was before?

Products/ZenModel/skins/zenmodel/editUserSettings.pt
glujan: Got it, thanks

Products/ZenUI3/browser/javascript.py
glujan: Because at this moment global variable does not exist yet - script which is creating it is loading later than `visualization.js` which depends on `Zenoss.USER_DATE_FORMAT` variable.

I didn't want to mess up with script ordering too much.

rDr4g0n: Ah, ok. sounds good.

rDr4g0n: After chatting about it, I think the global `Zenoss` object should be created here as it was before instead of using Ext namespaces.

glujan: And I'm unnecessarily creating global variables `z` and `env`.

Products/ZenUI3/browser/javascript.py
```diff
@@ -251,13 +251,19 @@ def snippet(self):
...
+            Ext.namespace('Zenoss.env');
+
             Zenoss.env.COLLECTORS = %r;
             Zenoss.env.priorities = %r;
             Zenoss.env.productionStates = %r;
             Zenoss.USER_TIMEZONE = "%s" || jstz.determine().name();
-        """ % ( collectors, priorities, productionStates, timezone )
-        return snippet
+            Zenoss.USER_DATE_FORMAT = "%s" || "MM/DD/YY";
+          })();
```
glujan: @rDr4g0n I've changed this snippet to be in a local scope. It's working exactly as in [zenoss.js](../../tree/develop/Products/ZenUI3/browser/resources/js/zenoss/zenoss.js#L8) (and many other places). I believe it's done properly this time.

glujan:
Fix from this PR will work on upgraded systems but won't on fresh deployments (at least it is like that in zendev environment)

Products/ZenUI3/browser/javascript.py
```diff
@@ -251,13 +251,19 @@ def snippet(self):
...
+        """ % ( collectors, priorities, productionStates, timezone, date_fmt )
+        return SCRIPT_TAG_TEMPLATE % snippet
```
rDr4g0n: This change is probably not what is expected. This function used to return a snippet of javascript, but this change to insert the snippet into `SCRIPT_TAG_TEMPLATE` makes this function return a full `<script>` tag with the javascript snippet inside. I would avoid changing how it works unless you're sure that it makes sense.

Products/ZenUI3/browser/configure.zcml
```diff
@@ -398,7 +398,8 @@
 
 <browser:viewlet
     name="zenoss_data"
-    manager=".interfaces.IMainSnippetManager"
+    weight="3"
+    manager=".interfaces.IJavaScriptSrcManager"
```
rDr4g0n: Why is `manager` changed here?

rDr4g0n: Why is `manager` changed here?

glujan: I had to change manager because `zenoss_data` in no longer a JavaScript snippet, but a whole `<script>` element, as you spotted in your other comment. Here I'll try explain why I made such a change.

In case you don't remember, a difference between managers is that `IMainSnippetManager` collects JS snippets and packs it into one `<script>` node, whereas `IJavaScriptSrcManager` is expecting a `<script>` node. Code grouped together by `IMainSnippetManager` is included after all `<script>` tags provided by `IJavaScriptSrcManager`.

The problem is that `zenoss-data` was originally included after `visualization-library` (see Products/ZenUI3/browser/javascript.zcml file) but now `visualization-library` depends on `zenoss-data`.

I couldn't move whole node which was rendered by `IMainSnippetManager` before `visualization-library` because it has its own dependencies (which are loaded by `IJavaScriptSrcManager`...). Thus I had to extract `zenoss-data` from `IJavaScriptSrcManager` and transform it into independent node, wrapping it in it's own `<script>` tag.

Products/ZenUI3/browser/javascript.py
glujan: Please read my other comment on `configure.zcml` file, I've addressed there your concerns.

Products/ZenUI3/browser/configure.zcml
glujan: I also had to introduce `weight` attribute to make it render before `visualization-library`. To make this attribute work, they need to have the same managers.

PR #2001, opened 2016-12-06
ZEN-26257: MetricShipper Can Fail Silently When Connection to MetricC…

olst:
…onsumer Fails

### DIFF
--- a/Products/ZenModel/migrate/addWebSocketHealthCheck.py
+++ b/Products/ZenModel/migrate/addWebSocketHealthCheck.py
@@ -1,0 +1,48 @@
+##############################################################################
+#
+# Copyright (C) Zenoss, Inc. 2016, all rights reserved.
+#
+# This content is made available according to terms specified in
+# License.zenoss under the directory where your Zenoss product is installed.
+#
+##############################################################################
+
+import os
+import re
+import logging
+log = logging.getLogger("zen.migrate")
+
+import Migrate
+import servicemigration as sm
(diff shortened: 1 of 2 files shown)


bin/healthchecks/MetricShipper/websocket_opened
```diff
@@ -0,0 +1,17 @@
...
+##############################################################################
+#
+# Copyright (C) Zenoss, Inc. 2016, all rights reserved.
+#
+# This content is made available according to terms specified in
+# License.zenoss under the directory where your Zenoss product is installed.
+#
+##############################################################################
+
+if [ -e /opt/zenoss/var/websocket_error ]
```
iancmcc: This file can be simplified to just `test -e /opt/zenoss/var/websocket_error`

Products/ZenModel/migrate/addWebSocketHealthCheck.py
```diff
@@ -0,0 +1,48 @@
...
+import Migrate
+import servicemigration as sm
+from servicemigration import HealthCheck
+sm.require("1.0.0")
+
+
+class AddWebSocketHealthCheck(Migrate.Step):
+    """Add `websocket_opened` healthcheck to MetricShipper service"""
+
+    version = Migrate.Version(5,2,20)
```
iancmcc: This version is not correct. Please look at more recent migrate scripts. I think the version is 107 right now.

PR #2002, opened 2016-12-07
fixes ZEN-26259: Event Details Sections are Expanded by Default.

HIMix:


romanbakaleyko:
this solution doesn't preserve settings between sessions, probably we want to store settings in the _browsers_state for each user. 
https://github.com/zenoss/zenoss-prodbin/blob/develop/Products/ZenUI3/browser/resources/js/zenoss/zenoss.js#L221

I would discuss this with @rDr4g0n

PR #2010, opened 2016-12-16
Bump version to 5.2.1

tgjonestx:


makefile
```diff
@@ -1,8 +1,12 @@
-VERSION  ?= 5.2.0
+VERSION  ?= 5.2.1
 BUILD_NUMBER ?= DEV
-BRANCH   ?= develop
-REVISION ?= 3
-ARTIFACT := prodbin-$(VERSION)-$(REVISION)-$(BRANCH).tar.gz
+BRANCH   ?= support/5.2.x
```
mozart27: I don't think this will work for the filepath.  The `/` will cause problems.

PR #2015, opened 2016-12-30
ZEN-26520: add netmask 64 to the ipv6 value from snmp

rderkach:
After remodeling snmp device with ipv6 addresses, they are
not presented  in interface component.
Add netmask 64 to ipv6 addresses to prevent overwriting ipv6
objects when nothing changed in configuration of the device,
since setter of ipv6 objects set them with netmask 64.

[JIRA](https://jira.zenoss.com/browse/ZEN-26520)


williamdean:
How does this behave if the interface isn't a /64? For example, many routers/servers will use a /127 for point-to-point links.

A more broader question is why aren't we using IPv6-MIB to fetch the IPv6 addresses?  Eg. ipv6AddrAddress & ipv6AddrPfxLength. http://www.net-snmp.org/docs/mibs/ipv6MIB.html#ipv6AddrPrefixTable

Products/DataCollector/plugins/zenoss/snmp/InterfaceMap.py
```diff
@@ -156,7 +156,7 @@ def process(self, device, results, log):
 
                 elif len(ip_parts) == 16:
                     addr_type = IpUtil.IPV6_ADDR_TYPE
-                    ip = IpUtil.bytesToCanonIp(ip_parts)
+                    ip = IpUtil.bytesToCanonIp(ip_parts) + '/64'
```
williamdean: Appears that we're just picking a frequently used subnet and assuming that it is the correct subnet. We should fetch the IPv6 subnet using the IPv6-MIB. This will likely require some significant changes to the InterfaceMap plugin.

rderkach:
Got your point. Changed that behavior and add new OID **ipAddressPrefix** `.1.3.6.1.2.1.4.34.1.5.2.6` in IP-MIB, which returns netmask.

jpeacock-zenoss:
As for IPv6-MIB, that particular MIB was superseded in 2006 by an updated IP-MIB.  From the IP-MIB spec:

```
revision "2006-02-02" {
     description
       "The IP version neutral revision with added IPv6 objects for
        ND, default routers, and router advertisements.  As well as
        being the successor to RFC 2011, this MIB is also the
        successor to RFCs 2465 and 2466.  Published as RFC 4293.";
   }
```

Products/DataCollector/plugins/zenoss/snmp/InterfaceMap.py
```diff
@@ -104,7 +107,10 @@ def process(self, device, results, log):
 
         # Add in IPv6 info
         ipv6table = tabledata.get("ipAddressIfIndex")
-        if ipv6table:
+        ipv6netmasktable = tabledata.get("ipAddressPrefix")
+        if ipv6table and ipv6netmasktable:
+            for key in ipv6table.keys():
+                ipv6table[key].update(ipv6netmasktable[key])
```
jpeacock-zenoss: This presumes that data for both IPv6 addresses _and_ netmasks must be available from the device if we want to model an IPv6 address.  What are the implications if we still used the retrieved IPv6 addresses regardless whether the netmasks could be retrieved?

PR #2022, opened 2017-01-10
ZEN-26580: add netmask 64 to the ipv6 value from snmp

rderkach:
This is port from https://github.com/zenoss/zenoss-prodbin/pull/2015

[JIRA](https://jira.zenoss.com/browse/ZEN-26580)


rderkach:
Do not merge, should discuss whether it would be in the next release.

PR #2042, opened 2017-01-19
fixes ZEN-26620: XSS on the device's details page

HIMix:
[Jira](https://jira.zenoss.com/browse/ZEN-26620)


HIMix:
Will squash commits using github "squash and merge" functionality.

PR #2047, opened 2017-01-20
ZEN-26635: Tracebacks on upgrade from 5.1.10 to 5.2.1

olst:
This one:
https://github.com/zenoss/zenoss-prodbin/blob/support/5.1.x/Products/ZenModel/migrate/removeEmptyGraphs.py
has switched 5.1.x dmd.version to 107, because of that a lot of 107 migrations from 5.2.x are not run during 5.1.x -> 5.2.1 upgrade and because of this we have an issue described in ZEN-26635.


olst:
agree, I wanted to get an advise, then I should update all migrate scripts versions in 5.2.x.

PR #2049, opened 2017-01-23
ZEN-22461  Improper Restriction of Excessive Authentication Attempts.

romanbakaleyko:
[1] made a zodb-dump to include the plugin to the initial db
[2] change chmod of the bash script.


romanbakaleyko:
Added installation of Account locker plugin in to the zenbuild.py to make it work for new installation from xml.

PR #2051, opened 2017-01-23
ZEN-17882: check for existing OIDs when installing zenpack

rderkach:
When installing zenpack, it ignors what OIDs are already
in the system and just create duplicates in its MIBs modules.
When zenpack install  OIDs using object.xml check whether OID
is presented in the system, and if yes skip creating new OID.

[JIRA&DEMO](https://jira.zenoss.com/browse/ZEN-17882)


rderkach:
Reverting regarding comment in related ticket.

PR #2053, opened 2017-01-24
ZEN-25282 Event Details Sections are Expanded by Default

romanbakaleyko:


romanbakaleyko:
port https://github.com/zenoss/zenoss-prodbin/pull/2050

PR #2064, opened 2017-02-01
ZEN-26631: auto cleared events have transient severity

olst:
port:
https://github.com/zenoss/zenoss-prodbin/pull/2038


StevePC:
doesn't this change stop any text fields from being displayed?

olst:
please see this image:
https://jira.zenoss.com/secure/attachment/38320/38320_23336-fix.png

PR #2066, opened 2017-02-01
ZEN-25764: Alert on missed collection runs

olst:
https://jira.zenoss.com/secure/attachment/38941/events.png
https://jira.zenoss.com/secure/attachment/38940/cleared.png

### DIFF
--- a/Products/ZenCollector/scheduler.py
+++ b/Products/ZenCollector/scheduler.py
@@ -19,7 +19,9 @@ single device or other monitored object.
 
 import logging
 import math
+import os
 import signal
+import sys
 import time
 from StringIO import StringIO
 
@@ -30,6 +32,7 @@ from twisted.python.failure import Failure
 
 from Products.ZenCollector.interfaces import IScheduler, IScheduledTask, IPausingScheduledTask
 from Products.ZenCollector.tasks import TaskStates
+from Products.ZenEvents import Event
 from Products.ZenUtils.Utils import dumpCallbacks
 from Products.ZenUtils.keyedset import KeyedSet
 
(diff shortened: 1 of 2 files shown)


Products/ZenCollector/scheduler.py
```diff
@@ -154,7 +157,14 @@ def late(self):
...
         """
-        # TODO: report an event
+        try:
+            # send event only for missed runs on devices.
+            self.task._eventService.sendEvent(
+                STATUS_EVENT, device=self.task._devId,
+                summary="Missed run on {}".format(self.task._manageIp),
+                component="zencommand", severity=Event.Warning)
+        except Exception:
+            pass
```
mozart27: Is it worth adding a logging line to the exception block?

mozart27: doesn't seem worth it after talking with @olst

Products/ZenCollector/scheduler.py
```diff
@@ -154,7 +157,14 @@ def late(self):
         """
         Called whenever this task is late and missed its scheduled run time.
         """
-        # TODO: report an event
+        try:
+            # send event only for missed runs on devices.
+            self.task._eventService.sendEvent(
+                STATUS_EVENT, device=self.task._devId,
+                summary="Missed run on {}".format(self.task._manageIp),
+                component="zencommand", severity=Event.Warning)
```
jplouis: The event isn't specific enough because a device can have many tasks, especially in collectors like zenpython.  Something to specifically identify this task needs to be added to the event, probably added to the eventkey. An event class should be specified as well.

A corresponding clear event is not sent when the task is back on time.

olst: this is the place where clear events are sent:
https://github.com/zenoss/zenoss-prodbin/blob/develop/Products/ZenRRD/zencommand.py#L631

jplouis: That has nothing to do with the new events you are sending.  Those events are only for zencommand and are specific events sent by the zencommand task.  The code you added will send events for all collectors.  Additionally your have since added the event class of Cmd_Fail which is probably not the correct class to use.

olst: thanks, got it.

PR #2069, opened 2017-02-01
Feature/bld 2

youngklee:
Add the Jenkins build script to the repo.

### DIFF
--- a/jenkins_build.sh
+++ b/jenkins_build.sh
@@ -20,41 +20,40 @@ check_var() {
     fi
 }
 
-# REPO_NAME: the name of the repo this job will checkout
-REPO_NAME=zenoss-prodbin
-# BRANCH: the branch of the repo this job will use
-BRANCH=develop
-ZENDEV_REPO=git@github.com:zenoss/zendev.git
-# ZENDEV_BRANCH: the zendev branch this job will use
-ZENDEV_BRANCH=zendev2
-ZENDEV_VER=0.2.0
-GO_VER=1.7.4
+# Check if all required parameters are defined.
+echo Checking WORKSPACE;check_var $WORKSPACE;echo OK
+echo Checking JOB_BASE_NAME;check_var $JOB_BASE_NAME;echo OK
 
+# The name of the repo this job will checkout
+REPO_NAME=zenoss-prodbin
+# The branch of the repo this job will use
+if [ -z "${BRANCH}" ]; then BRANCH=develop; fi
+ZENDEV_REPO=git@github.com:zenoss/zendev.git
+# The zendev branch this job will use
+if [ -z "${ZENDEV_BRANCH}" ]; then ZENDEV_BRANCH=zendev2; fi
+if [ -z "${ZENDEV_VER}" ]; then ZENDEV_VER=0.2.0; fi
+if [ -z "${GO_VER}" ]; then GO_VER=1.7.4; fi
 # The name of the zendev environment that will be created.
 # This name will be used as the devimg tag as well.
 ZENDEV_ENV=${REPO_NAME}-${JOB_BASE_NAME}
-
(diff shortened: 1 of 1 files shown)


jenkins_build.sh
```diff
@@ -0,0 +1,47 @@
+set -e
```
tgjonestx: Please add `#!/bin/bash` to the first line just to avoid any potential confusion.

Also, add `-x` so in case something really weird goes wrong in the future, the job log in Jenkins will have a more complete record of what the script did.

jenkins_build.sh
```diff
@@ -0,0 +1,47 @@
...
+
```
tgjonestx: Please add some comments to document which environment variables must be set by the caller and what each means.

jenkins_build.sh
```diff
@@ -0,0 +1,47 @@
...
+REPO_NAME=zenoss-prodbin
```
tgjonestx: Please add checks for required variables which are empty/undefined. This helps people who are unfamiliar with the script catch operator errors more quickly.  For example, see https://github.com/zenoss/zenoss-deploy/blob/develop/jenkins/import-ova-ami.sh#L20

jenkins_build.sh
```diff
@@ -0,0 +1,47 @@
...
+VENV_DIR=venv
+GO_VER=1.7.4
+
+echo Creating a virtual env...
+virtualenv ${VENV_DIR}
+source ${VENV_DIR}/bin/activate
+git clone ${ZENDEV_REPO} ${ZENDEV_DIR}
+cd ${ZENDEV_DIR};git checkout ${ZENDEV_BRANCH}
+pip install -e .
+zendev version | grep ${ZENDEV_VER} # Check zendev version
```
tgjonestx: Consider how this might look to someone reviewing the log file who doesn't know what to expect. 
When it works, the script will output "0.2.0" without explanation.  If the actual version of zendev changes, the script will output nothing.  I suggest something like:
```
ZENDEV_VERSION=`zendev version`
echo "zendev version = ${ZENDEV_VERSION}"
```

jenkins_build.sh
youngklee: Added.

jenkins_build.sh
youngklee: Added some comments.

jenkins_build.sh
youngklee: Added checks for undefined variables.

jenkins_build.sh
youngklee: Revised as suggested

jenkins_build.sh
```diff
@@ -1,36 +1,65 @@
...
-virtualenv ${VENV_DIR}
-source ${VENV_DIR}/bin/activate
-git clone ${ZENDEV_REPO} ${ZENDEV_DIR}
-cd ${ZENDEV_DIR};git checkout ${ZENDEV_BRANCH}
+virtualenv venv
+source venv/bin/activate
+git clone ${ZENDEV_REPO} zendev
+cd zendev;git checkout ${ZENDEV_BRANCH}
 pip install -e .
 zendev version | grep ${ZENDEV_VER} # Check zendev version
```
tgjonestx: i think you need to change this line so that the echo on the next line will report the version we ARE using as opposed to version we expect to use. Better yet, check the actual version against the expected version and exit if they are not the same
```
ZENDEV_VER_ACTUAL=`zendev version`
if [ "${ZENDEV_VER_ACTUAL}" != "${ZENDEV_VER}" ]; then
    echo "ERROR: expected zendev version ${ZENDEV_VER}, but found zendev version ${ZENDEV_VER_ACTUAL} "
    exit 1
fi
```

PR #2070, opened 2017-02-01
Fix/ZEN-26671 - zenhub stats kill zenhubworker processes

corntoole:
This adds hook for ZenHub to wait some amount of time before `kill -SIGUSR2` a new hubworker process. Zenhubworker instances are configured to ignore SIGUSR2 until its initialization is complete.

### DIFF
--- a/Products/ZenHub/zenhub.py
+++ b/Products/ZenHub/zenhub.py
@@ -967,8 +967,7 @@ class ZenHub(ZCmdBase):
         self.log.debug("Starting %s", ' '.join(args))
         prot = WorkerRunningProtocol(self)
         proc = reactor.spawnProcess(prot, exe, args, os.environ)
-        spawn_time = time.time()
-        proc.spawn_time = spawn_time
+        proc.spawn_time = time.time()
         self.workerprocessmap[proc.pid] = proc
         self.worker_processes.add(prot)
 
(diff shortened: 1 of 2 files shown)


Products/ZenHub/zenhub.py
```diff
@@ -959,6 +967,8 @@ def processEnded(self, reason):
         self.log.debug("Starting %s", ' '.join(args))
         prot = WorkerRunningProtocol(self)
         proc = reactor.spawnProcess(prot, exe, args, os.environ)
+        spawn_time = time.time()
```
mozart27: Why not combine these two lines? to be `proc.spawn_time = time.time()` ?

Products/ZenHub/zenhubworker.py
```diff
@@ -54,8 +54,8 @@ class zenhubworker(ZCmdBase, pb.Referenceable):
     "Execute ZenHub requests in separate process"
 
     def __init__(self):
+        signal.signal(signal.SIGUSR2, signal.SIG_IGN)
```
jplouis: What happens to SIGUSR2 now that is ignored? Will stats not be printed?

corntoole: The subsequent call to register the original SIGUSR2 handler is still there. SIGUSR2 is only ignored until the __init__() methods of all it's super classes along with the pre-requisite setup for zenhubworker.sighandler_USR2 is run.

PR #2073, opened 2017-02-03
Feature/ZEN-26752 zenhub enhancements

corntoole:
This PR adds more process info and the ability to run ZenHub and its zenhubworkers with profiling on.

To run zenhub with profiling on, add this line below to zenhub.conf
```profiling True```
To get current profile data from the profiler, attach to the zenhub container and the run the following:
```zenhub debug```

If profiling is turned on for zenhub, each invocation of the `zenhub debug` command will cause zenhub and the currently running zenhubworker to dump their individual profile stats to files within the standard OS temporary directory (most likely `/tmp`). Initially, zenhubworkers will inherit their profiling config from their hub, but you can drop into the container and turn profiling for the hubworker if you like. Debug statement are logged each time stats are dumped with the location of the stats that just created.

### DIFF
--- a/Products/ZenHub/zenhub.py
+++ b/Products/ZenHub/zenhub.py
@@ -89,6 +89,9 @@ from Products.ZenHub import PB_PORT
 from Products.ZenHub import OPTION_STATE
 from Products.ZenHub import CONNECT_TIMEOUT
 
+
+from Products.ZenUtils.debugtools import ContinuousProfiler
+
 HubWorklistItem = collections.namedtuple('HubWorklistItem', 'priority recvtime deferred servicename instance method args')
 WorkerStats = collections.namedtuple('WorkerStats', 'status description lastupdate previdle')
 LastCallReturnValue = collections.namedtuple('LastCallReturnValue', 'returnvalue')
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: Products/ZenHub/zenhub.py [hunk 1]
```diff
@@ -195,7 +198,7 @@ def perspective_getService(self,
                 service.addListener(listener, options)
             return service
 
-    def perspective_reportingForWork(self, worker):
+    def perspective_reportingForWork(self, worker, pid=None):
```
corntoole: A `ZenHub` instance has an attribute, `workers`, that is a collection of references to `zenhubworkers`, but the references are remote references and do not include any of the attributes of the referenced `zenhubworker` instance.

Products/ZenHub/zenhub.py
```diff
@@ -204,6 +207,8 @@ def perspective_reportingForWork(self, worker):
         @return None
         """
         worker.busy = False
+        if pid is not None:
```
corntoole: When this code is reached, `worker` is a [remote reference](http://twistedmatrix.com/documents/8.2.0/api/twisted.spread.pb.RemoteReference.html), and so it won't have the pid, unless the actual `zenhubworker` sends it its `ZenHub` when it reports for work, or I'm able to associate the pid of newly spawned `zenhubworker` proc, with the corresponding worker in `ZenHub.workers`. `ZenHub` has three collections of objects referring to `zenhubworkers` (see [here](https://github.com/zenoss/zenoss-prodbin/blob/develop/Products/ZenHub/zenhub.py#L414-L423)). This was the clearest way to associate the correct hubworker pid with the worker, when the Hub is reporting statistics on that worker.

Products/ZenHub/zenhub.py
mozart27: ok.

Products/ZenHub/zenhub.py
mozart27: ok.  thanks!

PR #2075, opened 2017-02-03
ZEN-23553 Object id value type of SNMP trap is now supported.

karolzen:
https://jira.zenoss.com/browse/ZEN-23553


Products/ZenEvents/zentrap.py
```diff
@@ -620,12 +619,12 @@ def decodeSnmpv1(self, addr, pdu):
         # Look for the standard trap types and decode them without
         # relying on any MIBs being loaded.
         eventType = {
-            0: 'snmp_coldStart',
-            1: 'snmp_warmStart',
-            2: 'snmp_linkDown',
-            3: 'snmp_linkUp',
-            4: 'snmp_authenticationFailure',
-            5: 'snmp_egpNeighorLoss',
```
karolzen: I deleted snmp_ prefix becouse it's redudant for me. Is this change ok?
Event's summary (before change):
- snmpv1:  snmp trap snmp_OIDNAME,
- snmpv2: snmp trap OIDNAME.

Products/ZenEvents/zentrap.py
```diff
@@ -651,15 +650,13 @@ def decodeSnmpv2(self, addr, pdu):
...
-                result["oid"] = '.'.join(map(str, vb_value))
+                result["oid"] = vb_value
                 eventType = self.oid2name(
                         vb_value, exactMatch=False, strip=False)
             else:
                 self._add_varbind_detail(vb_result, vb_oid, vb_value)
         result.update({name:','.join(vals) for name, vals in vb_result.iteritems()})
 
-        if eventType in ["linkUp", "linkDown"]:
-            eventType = "snmp_" + eventType
```
karolzen: Same like above.

Products/ZenEvents/zentrap.py
```diff
@@ -545,17 +545,21 @@ def _value_from_dateandtime(self, value):
...
-            value.decode('utf8')
-            return value
-        except UnicodeDecodeError:
-            # Try converting to a date
-            decoded = self._value_from_dateandtime(value)
-            if not decoded:
-                decoded = 'BASE64:' + base64.b64encode(value)
-            return decoded
+        # Check if values is OID (SNMP object id)
+        if isinstance(value, tuple) and len(value) > 2:
```
jpeacock-zenoss: It's only an OID if it has at least 3 elements in the tuple?

What's the scenario for 1 and 2 element tuples?

Does it have to be a tuple?  Could it be any kind of sequence?

karolzen: > It's only an OID if it has at least 3 elements in the tuple?
> Does it have to be a tuple? Could it be any kind of sequence?
Only OIDs are stored in tuples.

> What's the scenario for 1 and 2 element tuples?
Will not happen. OID minimum length is 3 (start with 1.3.). Can we call this 'right way'? This code will be working with `if isinstance(value, tuple)` too.

lundybernard: This change reverts the current _convert_value() function from the one we are using in 5.x to the older version from the 4.2.5-RPS repo.  Are we reverting intentionally?

karolzen: Yes, this was added by me to 4.2.5 before this PR.

Products/ZenEvents/zentrap.py
lundybernard: this looks reasonable, but why not get rid of the whole event name map, and just set eventType to name if the oid2name() call returns a valid name?

karolzen: Hmm, eventType is based on generic value in SNMPv1. So we can't search it in MIB's base.

PR #2031, opened 2017-01-12
ZEN-22461 Improper Restriction of Excessive Authentication Attempts

romanbakaleyko:

PR #2079, opened 2017-02-03
ZEN-22244 Improper Restriction of Excessive Authentication Attempts

romanbakaleyko:


romanbakaleyko:
port https://github.com/zenoss/zenoss-prodbin/pull/2031
https://github.com/zenoss/zenoss-prodbin/pull/2056

PR #2085, opened 2017-02-08
Add jenkins build script

youngklee:


jenkins_build.sh
```diff
@@ -0,0 +1,97 @@
...
+}
+
+# Check if all required parameters are defined.
+echo Checking WORKSPACE;check_var $WORKSPACE;echo OK
+echo Checking JOB_BASE_NAME;check_var $JOB_BASE_NAME;echo OK
+
+# The name of the repo this job will checkout
+REPO_NAME=zenoss-prodbin
+# The branch of the repo this job will use
+if [ -z "${BRANCH}" ]; then BRANCH=support/5.2.x; fi
```
tgjonestx: If this is the only change relative to the version on the `develop` branch, I would change the code on the develop branch to something like:
```
echo Checking BRANCH;check_var $BRANCH;echo OK
```
And I would add a Jenkins job parameter to each of the jenkins jobs, and specify a default value for that parameter as `develop` or `support/5.2.x` depending on which branch the respective job builds.

Then I would cherry-pick the script from the `develop` branch to the `support/5.2.x` branch.

The advantages of doing it this way are:
* The build logic for branches remains as close as possible
* When we need to create a new branch in the future (e.g. support/5.3.x), we don't have to change the build script; we only have to change the default arguments on the new Jenkins jobs.

youngklee: Revised as suggested

PR #2091, opened 2017-02-13
ZEN-23845: A single user can crash the UI running reports

olst:


Products/ZenUI3/browser/resources/js/zenoss/BackCompatPanel.js
```diff
@@ -19,6 +19,7 @@ Ext.define("Zenoss.IFramePanel", {
         config = Ext.applyIf(config || {}, {
             timeout: 5000, // Wait 5s for iframe to initialize before failing
             pollInterval: 50,
+            processed: true,
```
glujan: Why do you default to `true`?

Products/ZenUI3/browser/resources/js/zenoss/Reports.js
```diff
@@ -26,6 +26,7 @@ report_panel = Ext.create('Ext.panel.Panel', {
     }, {
         xtype: 'panel',
         ref: 'graph_reports',
+        graphs_added: true,
```
glujan: Same question about defaulting to `true`

olst: it should be in `false` only when a node was clicked and a report start loading.

glujan: Okay, so _while being added_ this is `false`. When graphs are already loaded or are not going to be loaded at all, it will be `true`. And analogically to my other comment regarding `processed: true`.

Did I got it right?

olst: right,
when we have them in `false` user can click other nodes but those reports will not start loading. (see lines 197-198)

Products/ZenUI3/browser/resources/js/zenoss/Reports.js
```diff
@@ -192,15 +194,19 @@ Ext.define('Zenoss.ReportTreePanel', {
...
             }
             var attrs = newnode[0].data;
             if (attrs.isGraphReport) {
+                report_panel.graph_reports.graphs_added = false;
                 report_panel.renderGraphReport(attrs);
             } else if (attrs.isMultiGraphReport) {
+                report_panel.graph_reports.graphs_added = false;
                 report_panel.renderMultiGraphReport(attrs);
             } else {
+                report_panel.backcompat.processed = false;
```
glujan: But where are you setting  `backcompat.processed` again to true? In [this](https://github.com/zenoss/zenoss-prodbin/pull/2091/files#diff-bb0add8e6b973593ece583f3eb114a48R148) load handler?

Products/ZenUI3/browser/resources/js/zenoss/Reports.js
glujan: Understood. I still have 1 more question, though.

Products/ZenUI3/browser/resources/js/zenoss/Reports.js
olst: correct

olst:
retest this

PR #2093, opened 2017-02-13
Allow components to have unique graph definitions

spopezen:
https://jira.zenoss.com/browse/ZEN-26519

These changes fix ZEN-26519 by checking the graph definition for each component instead of assuming they all have the same graph definition.  If a component does not have a graph definition, the behavior remains the same.


spopezen:
Retest this

PR #2094, opened 2017-02-14
ZEN-22461 Improper Restriction of Excessive Authentication Attempts

romanbakaleyko:
add checker to verify if plugin is already installed.


romanbakaleyko:
retest

PR #2098, opened 2017-02-14
Adding wrapper to monitor runzope memory size

achow-zenoss:
This is the first stage of the story.
This part has the scripts ready
The second part actually changes the zope container service def to call the scripts

### DIFF
--- a/bin/zen_restart_runzope
+++ b/bin/zen_restart_runzope
@@ -1,0 +1,33 @@
+#!/usr/bin/env bash
+#
+# zen_restart_runzope  shell script restarts runzope when its memory leaks too much.
+#
+##############################################################################
+#
+# Copyright (C) Zenoss, Inc. 2007, all rights reserved.
+#
+# This content is made available according to terms specified in
+# License.zenoss under the directory where your Zenoss product is installed.
+#
+##############################################################################
+
+# set the max to 2G
+(( max_m=2*1024*1024 ))
+
+# 10 minutes
+poll_t=600
+
+# runzope pid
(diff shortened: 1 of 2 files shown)


bin/zen_restart_runzope
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
+#
+# zen_restart_runzope  shell script restarts runzope when its memory leaks too much.
+#
+##############################################################################
+#
+# Copyright (C) Zenoss, Inc. 2007, all rights reserved.
```
tgjonestx: Should be 2017

bin/zen_restart_runzope
```diff
@@ -0,0 +1,33 @@
...
+# This content is made available according to terms specified in
+# License.zenoss under the directory where your Zenoss product is installed.
+#
+##############################################################################
+
+# set the max to 2G
+(( max_m=2*1024*1024 ))
+
+# 10 minutes
+poll_t=600
```
tgjonestx: It would be good if we could find a way to allow overrides from the CC for `max_m` and `poll_t`.

bin/zenrunzope
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+#
+# zenrunzope        This shell script takes care of starting and monitoring runzope.
+#
+##############################################################################
+#
+# Copyright (C) Zenoss, Inc. 2007, all rights reserved.
```
tgjonestx: should be 2017

PR #2109, opened 2017-02-17
ZEN-25763 Unable to selectively adjust zFileSystemSizeOffset based on…

romanbakaleyko:
… FS type


Products/ZenModel/FileSystem.py
```diff
@@ -110,11 +110,23 @@ class FileSystem(OSComponent):
         )
 
 
+    def getNoneRootTotalBlocks(self):
```
jpeacock-zenoss: It's not clear what a `noneRootTotalBlocks` property is.  Should this be a separate method?

Products/ZenModel/FileSystem.py
```diff
@@ -110,11 +110,23 @@ class FileSystem(OSComponent):
...
+        usedBlocks = self.usedBlocks()
+        if availBlocks and usedBlocks:
+            nonRootTotalBlocks = availBlocks + usedBlocks
+            return nonRootTotalBlocks
+
+
     def getTotalBlocks(self):
+        noneRootTotalBlocks = self.getNoneRootTotalBlocks()
+        if noneRootTotalBlocks:
+            return noneRootTotalBlocks
```
jpeacock-zenoss: From reading the result of your investigation of how to calculate the total block count, it seems Zenoss has to support two different algorithms.  It also seems like both algorithms have to exist behind the `getTotalBlocks` API.

The two algorithms are _sum of parts_ and _percent of total_, with _sum of parts_ being the preferred algorithm if the required parts are available.

No critique here; I'm just summarizing my understanding of what the fix is.

Products/ZenReports/plugins/filesystems.py
```diff
@@ -21,7 +21,7 @@ def getColumns(self):
                  Column('device_url', PythonColumnHandler( 'device.getDeviceUrl()' )),
                  Column( 'mount', PythonColumnHandler( 'component.mount' ) ),
                  Column( 'usedBytes', RRDColumnHandler( 'usedFilesystemSpace__bytes' ) ),
-                 Column( 'totalBytes', PythonColumnHandler( 'component.totalBytes()' ) ) ]
+                 Column( 'totalBytes', PythonColumnHandler( 'component.totalBlocks * component.blockSize ' ) ) ]
```
jpeacock-zenoss: Why is the result of the `totalBytes` method not the desired value here?

Products/ZenModel/FileSystem.py
jpeacock-zenoss: If kept as a separate method, it needs to be renamed and _not_ be public because it's an implementation detail of the `getTotalBlocks` method.  I would suggest `_getSumOfPartsTotalBlocks` as a name that better describes what the value returns (with a `_` prefix to make it not public).

PR #2112, opened 2017-02-17
Feature/zen 26864 cdk improvements

mayathegreat:
added logic for enabling / disabling controls and other features to improve graph useability
small syntax/typo fixes


mayathegreat:
screen-before
screen-after
See screenshots above -- note that the toolbar is grayed out when we are refreshing the graphs. Not quite visible but the gear icon is also spinning to indicate it is loading in

PR #2124, opened 2017-03-01
ZEN-26529 DERIVE Data Points Exhibiting COUNTER Behavior in Graphs

yuri1984:
demo for the ticket:
https://jira.zenoss.com/browse/ZEN-26529?focusedCommentId=103363&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-103363

### DIFF
--- a/Products/Zuul/infos/template.py
+++ b/Products/Zuul/infos/template.py
@@ -488,6 +488,9 @@ class DataPointInfo(InfoBase):
             if self._object.isCounter():
                 rateOptions['counter'] = True
                 rateOptions["resetThreshold"] = 1
+            if self._object.rrdtype == "DERIVE" and str(self.rrdmin) != '0':
+                # to show down spikes on DERIVE
+                rateOptions['counter'] = False
             if self._object.rrdmax is not None:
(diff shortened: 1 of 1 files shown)


Products/Zuul/infos/template.py
```diff
@@ -488,6 +488,9 @@ def getRateOptions(self):
             if self._object.isCounter():
                 rateOptions['counter'] = True
                 rateOptions["resetThreshold"] = 1
+            if self._object.rrdtype == "DERIVE" and str(self.rrdmin) != '0':
```
yuri1984: str(self.rrdmin) take here because of this comparison:
https://github.com/zenoss/zenoss-prodbin/blob/develop/Products/ZenModel/RRDDataPoint.py#L200-200

jplouis: Why not just change the isCounter() method?

yuri1984: Would be more logical. However it is required to post `counter = False` into JSON request to OpenTSDB to fix the issue. Other request types does not require this value to be set.

PR #2130, opened 2017-03-02
Fix ZEN-26637: raise log level message for start of model run, log if no collection happened

corntoole:
I did the following to provide more accurate timing info for modeling runs

  - add flag to check whether any devices were processed during
`ZenModeler` collector loop
  - raise level of log message denoting start of collector loop for
proper modeling run timings

### DIFF
--- a/Products/DataCollector/zenmodeler.py
+++ b/Products/DataCollector/zenmodeler.py
@@ -113,6 +113,8 @@ class ZenModeler(PBDaemon):
         self.devicegen = None
         self.counters = collections.Counter()
         self.configFilter = None
+        # ZEN-26637 - did we collect during collector loop?
+        self.didCollect = False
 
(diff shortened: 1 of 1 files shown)


Products/DataCollector/zenmodeler.py
```diff
@@ -1110,10 +1115,10 @@ def mainLoop(self, driver):
         if self.clients:
             self.log.error("Modeling cycle taking too long")
             return
-
+        self.didCollect = False
```
mozart27: Is it necessary to set this if it's already set in the class instantiation where you declare `self.didCollect = False` (line 117)?

corntoole: the flag needs to be cleared every time a modeling run starts. It may be that the assignment should be removed from the `ZenModeler.__init__()`

PR #2138, opened 2017-03-06
ZEN-26741 All on Same Graph Legends

carlcamera:
Identify metrics in all-in-same-graph graph by the same name when in individual charts.


carlcamera:
Code appears first on RM 5.2.x branch build #337. 
http://platform-jenkins.zenoss.eng/job/product-assembly/job/support-5.2.x/job/core-pipeline/327/

PR #2140, opened 2017-03-07
ZEN-25763 Unable to selectively adjust zFileSystemSizeOffset based on…

romanbakaleyko:
… FS type

### DIFF
--- a/Products/ZenModel/FileSystem.py
+++ b/Products/ZenModel/FileSystem.py
@@ -110,11 +110,23 @@ class FileSystem(OSComponent):
         )
 
 
+    def _getTotalBlocksAsSum(self):
+        availBlocks = self.cacheRRDValue('availBlocks', None)
+        usedBlocks = self.usedBlocks()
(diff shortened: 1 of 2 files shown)


Products/ZenModel/FileSystem.py
```diff
@@ -110,11 +110,23 @@ class FileSystem(OSComponent):
         )
 
 
+    def _getTotalBlocksAsSum(self):
+        availBlocks = self.cacheRRDValue('availBlocks', None)
+        usedBlocks = self.usedBlocks()
+        if availBlocks and usedBlocks:
```
jpeacock-zenoss: This will evaluate to False if either `availBlocks` or `usedBlocks` has the value zero.  Is this what you intended? Or did you mean to say that both values _have to exist_ which means you should be checking for `None` explicitly.

romanbakaleyko: correct

PR #2141, opened 2017-03-07
ZEN-22199: Test datasource UI functionality should test on device's a…

olst:
…ctual collector


bin/zentestds
```diff
@@ -0,0 +1,118 @@
...
+                plugins=plugins,
+                device=device,
+                datacollector=self)
+            connection.run()
+        except:
+            self.log.exception("Error opening connection.")
+
+    def main(self, driver):
+        yield self.config().callRemote('getDeviceConfig', [self.options.device])
+        n = driver.next()
```
jpeacock-zenoss: Any reason why this code couldn't be written using `inlineCallbacks` instead of the driver object?  Driver was this thing that was written before `inlineCallbacks` came along.  Using `inlineCallbacks` should be the preferred approach for new asynchronous code.

olst:
previous commit: https://github.com/Zenoss/zenoss-prodbin/commit/348cb93dc205266a3017de44774928ae47d7b6ce

olst:
retest this

PR #2144, opened 2017-03-07
ZEN-26981: add tool to save/restore dashboards' states

fraibacas:


fraibacas:
Tool raises an exception in Zenoss core bc .dashboards() relationship is not available

PR #2148, opened 2017-03-08
ZEN-26990

ntzen:
port from https://github.com/zenoss/4.2.5-RPS/pull/735
This is the same patch put in place for ZEN-22711 but was never moved into the 5x code


olst:
retest this

PR #2150, opened 2017-03-08
Fix ZEN-26637 pt2 adds more logging info, fixes race between mainLoop & checkStop

corntoole:
- added a collector loop iteration counter
- add more logging to `fillCollectionSlots()`,
`processClientFinished()` & `checkStop()` to follow progress of modeling
run
- The root cause of the defect seems to be that `checkStop` signals
the end of a collector loop before `mainLoop` can receive and set the
modelers deviceList on which to collect. To remedy, we set a flag to let
`checkStop` know if the `deviceList` was set for the current modeling
run. If this fix doesn't work, the additional logging information will
still provide the user with more info to determine what work is being
done during a given modeling run.


corntoole:
Below is a log excerpt showing a reproduction of the defect

```2017-03-08 17:02:57,938 INFO zen.ZenModeler: Scan time: 293.76 seconds for collector loop #2
2017-03-08 17:02:57,938 INFO zen.ZenModeler: Scanned 28 of 28 devices during collector loop #2
2017-03-08 17:03:04,182 INFO zen.ZenModeler: Starting collector loop #3...
2017-03-08 17:03:04,183 INFO zen.ZenModeler: Collecting for path /Devices
2017-03-08 17:03:04,939 INFO zen.ZenModeler: ZenModeler.checkStop() called from fillCollectionSlots
2017-03-08 17:03:04,939 INFO zen.ZenModeler: Did not collect during collector loop
2017-03-08 17:03:04,939 INFO zen.ZenModeler: Scan time: 0.76 seconds for collector loop #3
2017-03-08 17:03:04,939 INFO zen.ZenModeler: Scanned 28 of 28 devices during collector loop #3
2017-03-08 17:04:06,040 INFO zen.ZenModeler: Got 59 devices to be scanned during collector loop #3
2017-03-08 17:04:06,381 INFO zen.ZenModeler: Filled collection slots for 1 of 59 devices during collector loop #3
```

As can be seen in the logs, when `mainLoop` kicks off collector loop no.3, a chain of deferred callbacks eventually end up in a call to `checkStop` that determines that collector loop no.3 is done (see [zenmodeler.py#L830-L831](https://github.com/zenoss/zenoss-prodbin/blob/support/5.2.x/Products/DataCollector/zenmodeler.py#L830-L831)) before `mainLoop` receives a `deviceList` to begin filling collection slots (see [zenmodeler.py#L1122](https://github.com/zenoss/zenoss-prodbin/blob/support/5.2.x/Products/DataCollector/zenmodeler.py#L1122)). The work to be done during collector loop no.3 will carry on asynchronously, but its post-collector-loop reporting will already have been completed and as a result inaccurate.

PR #2157, opened 2017-03-09
Feature/zen 26522 ipv6 ignore local host services

mozart27:


Products/DataCollector/plugins/zenoss/snmp/IpV6ServiceMap.py
```diff
@@ -40,6 +40,12 @@ class IpV6ServiceMap(SnmpPlugin):
                     {'.1':'v4', '.2':'v6'}),
     )
 
+    def ignoredAddr(self, addr):
+        # ignore localhost designations
+        # 127.0.0.0/8
+        # ::1
+        return addr in ['::1'] or addr.startswith('127.0.0')
```
williamdean: A 127.0.0.0/8 is actually 24 bits left for the host portion, so it's 127.x.x.x. so you'd need to do `addr.startswith('127.')`

PR #2159, opened 2017-03-09
ZEN-26850 Missed Pings count fix

mayathegreat:
Added a method that automatically runs after the configs are updated which clears out the current ping "Missed Runs" count


mayathegreat:
screenshot from 2017-03-09 14-16-50

PR #2163, opened 2017-03-10
ZEN-27012 Graph Title Breadcrumb

ssOlexBaiko:
[Jira with demo](https://jira.zenoss.com/browse/ZEN-27012)


jpeacock-zenoss:
I think this feature should be implemented as a global behavior rather than a per graph definition option.  For a user to go to every graph definition to check (or uncheck) an option for a presentation preference is _not_ a great usability experience.

I can think two options that should achieve the goal of the ticket but with good usability:
1. Make the option to toggle between full device path and device only available directly in the graph display area.  Although, I'm not sure where to save a preference for which to display by default.
2. Make the display dynamic.  If displaying the full device path is too long, crop it down to fit it in.  For example, if `/Server/SSH/Linux/IACApps/SharedServices/Prod/Reporting/Mindsparkfeeds/lvprdfeedbe1` is too long, then cropping it to `/Server/.../lvprdfeedbe1` will allow it to fit.

I believe option 2 provides the best experience because it requires no action on the user one way or the other and should work for users who may want it one way or the other.  Some users want to see the organizer but will understand that cropping the name to fit the space.  ZEN-27012 is about users who don't care one way or the other but just want a presentation that doesn't wrap or otherwise mess up the display.

To expand on the option 2 example, here are the possible variations that are possible to fit the name:
/Server/SSH/Linux/IACApps/SharedServices/Prod/Reporting/Mindsparkfeeds/lvprdfeedbe1
/Server/SSH/Linux/IACApps/SharedServices/Prod/Reporting/.../lvprdfeedbe1
/Server/SSH/Linux/IACApps/SharedServices/Prod/.../lvprdfeedbe1
/Server/SSH/Linux/IACApps/SharedServices/.../lvprdfeedbe1
/Server/SSH/Linux/IACApps/.../lvprdfeedbe1
/Server/SSH/Linux/.../lvprdfeedbe1
/Server/SSH/.../lvprdfeedbe1
/Server/.../lvprdfeedbe1
lvprdfeedbe1

PR #2176, opened 2017-03-14
Explicitly set next expected call in case task was never executed or …

itshane:
…scheduled.

Fixes ZPS-960


Products/ZenCollector/scheduler.py
```diff
@@ -505,12 +508,23 @@ def getTasksForConfig(self, configId):
...
         loopingCall = self._loopingCalls.get(taskName, None)
         if loopingCall:
             return loopingCall._expectNextCallAt
 
+    def setNextExpectedRun(self, taskName, taskInterval):
+        """
+        Set the next expected execution time for given task.
+        """
+        loopingCall = self._loopingCalls[taskName]
+        loopingCall._expectNextCallAt = time.time() + taskInterval
```
achow-zenoss: Could loopingCall nil? as check in line #514?

itshane: It shouldn't be the case according to the execution order, however, I've updated the branch, please recheck.

itshane:
retest this

itshane:
retest this

PR #2178, opened 2017-03-15
ZEN-26170 Edit page for multi-graph reports doesn't show more than 39…

karolzen:
https://jira.zenoss.com/browse/ZEN-26170


jpeacock-zenoss:
You've replaced the spaces with tabs.  Please restore the use of spaces for indentation.

PR #2181, opened 2017-03-15
service migration and tests for new opentsdb setting

ecifreo:


Products/ZenModel/migrate/updateOpenTSDBSkipUnresolvedTags.py
```diff
@@ -0,0 +1,68 @@
...
+
+sm.require("1.0.0")
+log = logging.getLogger("zen.migrate")
+
+
+class UpdateOpenTSDBSkipUnresolvedTags(Migrate.Step):
+    """ Enable OpenTSDB option to skip unresolved tags.
+
+        This migration step iterates through services that have opentsdb.conf as an original
+        configuration file.  For each of these services both original and latest config fles
```
mozart27: nit - `fles` should prolly be `files`

Products/ZenModel/migrate/updateOpenTSDBSkipUnresolvedTags.py
```diff
@@ -0,0 +1,68 @@
...
+        """ This method is called by the migration process and overrides the method
+            located in the parent class Migrate.Step.
+        """
+        try:
+            ctx = sm.ServiceContext()
+        except sm.ServiceMigrationError:
+            log.info("Couldn't generate service context, skipping.")
+            return
+
+        tsdbs = filter(lambda s: "opentsdb" in ctx.getServicePath(s), ctx.services)
```
mozart27: Is this lambda to get `tsdbs` just informational?  It appears it's overwritten in the next few lines.

ecifreo: It does get overwritten, but the first time called it gets all configs plus all "original configs", so yes, informational. Next time only gets "original configs".

Products/ZenModel/migrate/updateOpenTSDBSkipUnresolvedTags.py
ecifreo: Copy pasta from another migration but I will fix, don't like typos.

PR #2190, opened 2017-03-17
Adds explicit description of snmp decode failure

lundybernard:
Fixes ZEN-24658

SNMP Trap values of "BASE64:..." for values we failed to decode
were causing confusion for users who believed the original trap
was sent in base64.

* new value reads "Failed to decode: converted to BASE64:"<base64 encoded value>"
* adds unit tests for zentrap

### DIFF
--- a/Products/ZenEvents/tests/test_zentrap.py
+++ b/Products/ZenEvents/tests/test_zentrap.py
@@ -2,6 +2,8 @@ from Products.ZenTestCase.BaseTestCase import BaseTestCase
 from mock import patch
 from Products.ZenEvents.zentrap import TrapTask
 
+from struct import pack
+
 
 class TrapTaskUnitTest(BaseTestCase):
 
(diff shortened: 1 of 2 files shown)


lundybernard:
rerun checks

Products/ZenEvents/tests/test_zentrap.py
```diff
@@ -0,0 +1,59 @@
...
+
+    @patch("Products.ZenEvents.zentrap.TrapTask.__init__")
+    def test_convert_value_handles_invalid_chars(self, mock_traptask_init):
+        mock_traptask_init.return_value = None
+        trap_task = TrapTask()
+
+        value = '\xde\xad\xbe\xef\xfe\xed\xfa\xce'
+        self.assertEqual(
+            trap_task._convert_value(value),
+            'Failed to decode: converted to BASE64:"3q2+7/7t+s4="'
```
jpeacock-zenoss: I don't think this will succeed.  The string starts with `BASE64` not `Failed to decode`.

PR #2197, opened 2017-03-20
ZEN-26380: manage dashboards tool: handle exception retrieving user g…

fraibacas:
…roups

### DIFF
--- a/Products/ZenUtils/manage_dashboard_state.py
+++ b/Products/ZenUtils/manage_dashboard_state.py
@@ -112,6 +112,7 @@ def set_up_logger():
     log.setLevel(logging.INFO)
     log.propagate = False
     return log
+
 log = set_up_logger()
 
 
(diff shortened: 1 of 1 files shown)


Products/ZenUtils/manage_dashboard_state.py
```diff
@@ -142,13 +143,17 @@ def dashboard_users(self):
...
+        group_names = []
+        try:
+            user_id = user.getId()
+            ap = self._dmd.zport.acl_users.manage_getUserRolesAndPermissions(user_id)
+            if ZEN_MANAGE_DMD in ap.get('allowed_permissions', {}):
+                group_names = user.getAllGroupSettingsNames()
+            else:
+                group_names = user.getUserGroupSettingsNames()
+        except Exception as e:
+            log.warn("Could not retrieve user {}'s groups: {}".format(user, e))
```
mozart27: Should this be an `error` instead of a `warn`?

fraibacas: It is a cli tool and it is not a big deal if we cant retrieve the user's groups

PR #2209, opened 2017-03-22
ZEN-27069 When I click on File systems component, it throws a flare e…

romanbakaleyko:
…rror message


romanbakaleyko:
see comment in https://jira.zenoss.com/browse/ZEN-27069

PR #2211, opened 2017-03-22
Drop a zendev env first if the name is already used

youngklee:


jenkins_build.sh
```diff
@@ -77,6 +77,7 @@ source $(zendev bootstrap)
 
 echo Creating a zendev environment...
 cd $WORKSPACE
+zendev drop ${ZENDEV_ENV}
```
tgjonestx: If `${ZENDEV_ENV}` does not exist, will zendev throw an error and cause the entire script to stop?

youngklee: No. It still returns code 0.

PR #2213, opened 2017-03-22
Fix collection check logic to check collage delay even if previous mo…

ecifreo:
…deling runs succeeded.


ecifreo:
Thanks @corntoole , but JP has some changes he wants me to make. I'll post another commit soon after testing.

PR #2215, opened 2017-03-23
ZEN-27112 Flare error message appears when click on filesystem compon…

romanbakaleyko:
…ent of windows device


romanbakaleyko:
port https://github.com/zenoss/zenoss-prodbin/pull/2209

PR #2225, opened 2017-03-24
Feature/zen 26935 update guides

mayathegreat:
removed old docs
updated with new docs


mayathegreat:
oh, you're right! Missed those.

PR #2232, opened 2017-03-28
ZEN-27116: add flag to daemon to indicate it is loading configs from …

fraibacas:
…cache


Products/ZenCollector/daemon.py
```diff
@@ -245,10 +245,13 @@ def __init__(self, preferences, taskSplitter,
         self.preferences.postStartup()
         self.addedPostStartupTasks = False
 
-        # Used in resmgr. True once we have finished loading the configs
-        # for the first time after a restart
+        ''' Variables used by enterprise collector in resmgr '''
```
jpeacock-zenoss: Please don't use doc strings for comments.

PR #2237, opened 2017-03-29
ZEN-26955: set Not Modeled when we get 0 from getSnmpLastCollection

rderkach:
The model collection age report displayed epoch time instead of
displaying 'Not Modeled'.
If device wasn't modeled getSnmpLastCollection will return zero,
in this case we use getSnmpLastCollectionString which will return
'Not Modeled'.

[JIRA&DEMO](https://jira.zenoss.com/browse/ZEN-26955)


jpeacock-zenoss:
Any reason why the output from `getSnmpLastCollectionString` can't be used all the time?

rderkach:
@jpeacock-zenoss Yes there is a reason. If it return 'Not Modeled' we will have a traceback from timeTime() method.

jpeacock-zenoss:
OK.  I see.  The time has to pass through the `convertToUsersTimeZone` method because `getSnmpLastCollectionString` doesn't do that (it converts the time to the system's local timezone), so we only conditionally use `getSnmpLastCollectionString`.

PR #2241, opened 2017-03-29
ZEN-27126 Flares on component graphs

carlcamera:
Fix problem with flares on component graphs.

### DIFF
--- a/Products/Zuul/facades/devicefacade.py
+++ b/Products/Zuul/facades/devicefacade.py
@@ -11,6 +11,7 @@
 import socket
 import re
 import os
+import copy
 import logging
 import subprocess
 from itertools import imap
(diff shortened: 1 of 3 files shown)


Products/Zuul/facades/devicefacade.py
```diff
@@ -953,9 +954,12 @@ def getComponentGraphs(self, uid, meta_type, graphId, allOnSame=False):
             return []
 
         if allOnSame:
-            # ZEN-26498 identify by name to match individual charts
+            # ZEN-26498 ZEN-27126 get docker legends to match individual graphs
             for comp in components:
-                comp.id = comp.name()
+                if (hasattr(comp,'docker_host')):
```
csmouse: is it cool to be putting something specific like 'docker_host' in this generic devicefacade class?

carlcamera: The other devices seem to provide the proper objects when grouped into a multigraph -- only Docker containers (via the Docker ZenPack) has been shown to behave this way.

csmouse: I suggest looking at the monkeypatch technique used in https://github.com/zenoss/ZenPacks.zenoss.vSphere/blob/develop/ZenPacks/zenoss/vSphere/patches/platform.py and https://github.com/zenoss/ZenPacks.zenoss.Microsoft.Windows/blob/develop/ZenPacks/zenoss/Microsoft/Windows/patches/platform.py as an alternative that doesn't taint DeviceFacade with docker-specific knowledge. This would apply to the Docker ZP.

jplouis: It may not be a monkey patch, but I think talking to solutions may help understand why the object doesn't behave the same as others and a change can be made to the zenpack.

PR #2242, opened 2017-03-30
ZEN-26198 MultiGraph reports show only the last group name above all …

romanbakaleyko:
…graph groups


romanbakaleyko:
DEMO https://youtu.be/yl9dtoK2rPo

PR #2249, opened 2017-04-05
ZEN-24543: OS Process Monitoring (SNMP) Can Send Duplicate Events

olst:


Products/ZenEvents/events2/processing.py
```diff
@@ -605,9 +607,11 @@ def __call__(self, eventContext):
                             objList = [objList]
                         uuids = set(sum((self._manager.getUuidsOfPath(obj) for obj in objList), []))
                         if uuids:
-                            eventContext.eventProxy.tags.addAll(tagType, uuids)
+                            evtproxy.tags.addAll(tagType, uuids)
                         componentOrgs[orgTypeName] = [obj.getOrganizerName() for obj in objList]
                 self._addComponentOrganizers(eventContext, componentOrgs)
+            elif evtproxy.eventClass == Status_OSProcess and evtproxy.severity > 0:
+                evtproxy._action = 'drop'
```
jpeacock-zenoss: This is a rather event class specific test.  I'm now wondering, after reading the ticket, whether **all** event classes can have this behavior.  The reproduction steps in the ticket seem contrived, so let's add another layer of contrivance: instead of killing a process to force a `Status_OSProcess` event, manually add some other event related to a model element that can be deleted and delete that element while `zenhub` is down.

PR #2253, opened 2017-04-06
Fixes ZEN-25893

zsalinas:
If the top level Mibs organizer is passed as a target into 'getChildMoveTarget' it results in a REQUEST attribute error because it is unable to resolve its own organizer off of self.dmdRoot ... which is it's own organizer. If we know the moveTarget is the root Mibs organizer, we should resolve it immediately and not pass it into getChildMoveTarget.


csmouse:
Closing this PR for now. There are merge conflicts to resolve and then please resubmit for code review.

PR #2259, opened 2017-04-10
ZEN-27309 Add user-agent string to curl healthchecks

ed-:
Several healthchecks involve the use of curl. This migration
adds a unique user-agent to these calls to better track where
the calls are coming from in logs.

Several calls in the bin/healthchecks directory also use curl
and have had similar user-agent strings added.


johnlhamilton:
retest this

PR #2284, opened 2017-04-21
IsAudit Flag Migration

cchristiezen:
Add a migration for the new IsAudit Flag.  If the audit log is present in the logConfigs for a service, the IsAudit flag will be set to true.


Products/ZenModel/migrate/addAuditLogFlag.py
```diff
@@ -0,0 +1,45 @@
...
+            return
+
+        changed = False
+
+        for service in ctx.services:
+            if not service.logConfigs:
+                continue
+
+            for config in service.logConfigs:
+                if config.path == "/opt/zenoss/log/audit.log":
```
tgjonestx: nit-pick - technically should be "and not config.isAudit"

PR #2291, opened 2017-04-26
Improve search, fix some of the Zuul tests

johnlhamilton:
This PR allows catalog searches against unindexed fields and fixes some issues with search revealed by the Zuul integration tests.  It also makes changes to the test framework to enable tests that use the ModelCatalog.

### DIFF
--- a/Products/Zuul/facades/__init__.py
+++ b/Products/Zuul/facades/__init__.py
@@ -155,12 +155,17 @@ class TreeFacade(ZuulFacade):
         if self.validRegex(filterRegex):
             orgquery = (Eq('objectImplements','Products.ZenModel.%s.%s' % (organizerClass, organizerClass)) &
                         MatchRegexp('uid', filterRegex))
-            paths = [ "{0}/*".format(b.getPath()) for b in IModelCatalogTool(self._dmd).search(query=orgquery)]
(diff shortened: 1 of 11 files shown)


Products/Zuul/facades/__init__.py
```diff
@@ -155,12 +155,17 @@ def findMatchingOrganizers(self, organizerClass, organizerPath, userFilter):
         if self.validRegex(filterRegex):
             orgquery = (Eq('objectImplements','Products.ZenModel.%s.%s' % (organizerClass, organizerClass)) &
                         MatchRegexp('uid', filterRegex))
-            paths = [ "{0}/*".format(b.getPath()) for b in IModelCatalogTool(self._dmd).search(query=orgquery)]
+            paths = [ "{0}/".format(b.getPath()) for b in IModelCatalogTool(self._dmd).search(query=orgquery)]
```
fraibacas: Do we need the trailing slash? right now it does not matter bc path is tokenized, but we may change it in the future?

fraibacas:
With the changes needed to make the tests work we are breaking compatibility of zp tests with previous versions of Zenoss

Products/Zuul/facades/__init__.py
johnlhamilton: Yeah I'll add a TODO.  If we change it, it will break a test so we'll know right away.

PR #2295, opened 2017-04-27
Feature/zen 27332

ecifreo:
Service migration for rabbitmq to run under supervisord along with new metrics collection script.


bin/metrics/gather.py
```diff
@@ -0,0 +1,167 @@
...
+        self.tenant_id = os.environ.get("CONTROLPLANE_TENANT_ID", "")
+        self.tags = {
+            "serviceId": self.service_id,
+            "instance": self.instance_id,
+            "hostId": self.host_id,
+            "tenantId": self.tenant_id
+        }
+
+    def build_metric(self, name, value, timestamp, tags=None):
+        _value = self.float_value(value)
```
jplouis: why is it always converted to float?

ecifreo: Inspired by this code I originally meant to re-use: https://github.com/zenoss/zenoss-prodbin/blob/develop/Products/ZenHub/metricpublisher/utils.py#L88

PR #2296, opened 2017-04-28
Add commit_dirty param, add tests.

spopezen:


Products/Zuul/catalog/tests/test_model_catalog.py
```diff
@@ -0,0 +1,169 @@
...
+from Products.Zuul.tests.base import BaseTestCase, init_modelcatalog
+from Products.Zuul.catalog.model_catalog import ModelCatalogClient
+from Products.Zuul.catalog.interfaces import IModelCatalog
+from Products.ZenModel.Device import Device
+from Products.ZenModel.ZDeviceLoader import JobDeviceLoader
+from Products.ZenUtils.ZenScriptBase import ZenScriptBase
+from zenoss.modelindex.model_index import SearchParams
+from zope.component import getUtility
+
+dmd = ZenScriptBase(connect=True).dmd
```
fraibacas: also, here https://github.com/zenoss/zenoss-prodbin/blob/develop/Products/ZenTestCase/BaseTestCase.py#L149 we make sure that tests don't commit. Are your calls to commit working? if so we may be introducing undesired behavior in the test framework

spopezen:
Opening separate pull request to add commit_dirty.
Closing this pull request, as we have decided to do these tests once we have a better way to deal with testing the transactions.

PR #2305, opened 2017-05-05
ZEN-27420: More prodbin test fixes

fraibacas:


Products/ZenModel/tests/testDeviceClass.py
```diff
@@ -52,22 +52,21 @@ def testCreateInstanceDeviceAndIndex(self):
 
     def testSearchDevicesOneDevice(self):
         devices = self.dmd.Devices
-        self.assertRaises(Redirect, devices.searchDevices, "testdev2",
-                          REQUEST=dict(junk=1))
+        self.assertEqual( len(  devices._findDevice("testdev2") ), 1 )
```
johnlhamilton: Why did this change from raising an exception to returning 1 result?

fraibacas: The method the test was calling has been deprecated. https://github.com/zenoss/zenoss-prodbin/blob/model_catalog/Products/ZenModel/ZentinelPortal.py#L91

PR #2313, opened 2017-05-11
ZEN-27478 Trigger a job to rename metrics after re-identifying a device

youngklee:


Products/ZenModel/Device.py
```diff
@@ -1909,11 +1910,32 @@ def renameDevice(self, newId=None, REQUEST=None):
             parent.manage_renameObject(oldId, newId)
             self.restoreCurrentProdStates(currProdStates)
             self.setLastChange()
+
+            # Replace the old id in performance data with the new id.
+            # See ZEN-27329.
+            self.amendPerfDataAfterRename(oldId, newId)
```
jplouis: Should this be optional? That is should a user be able to rename a device with out "moving" metric data?

Products/ZenModel/Device.py
```diff
@@ -1909,11 +1910,32 @@ def renameDevice(self, newId=None, REQUEST=None):
...
+        """
+        Replace a dev id in metric names and tag values with the new id after
+        renaming the device.
+        """
+        self.dmd.JobManager.addJob(
+            FacadeMethodJob,
+            description='Rename device {} to {} in performance data'.format(oldId, newId),
+            kwargs=dict(
+                facadefqdn='Products.Zuul.facades.metricfacade.MetricFacade',
+                method='renameDevice',
```
jplouis: Does this actually exist yet?

youngklee: It is in the next PR I'm going to make.

PR #2312, opened 2017-05-11
ZEN-27496 Tree nodes need a manual refresh to be visible after creation

ssOlexBaiko:
[Jira](https://jira.zenoss.com/browse/ZEN-27496)

PR #2314, opened 2017-05-12
ZEN-27394 Tree nodes need a manual refresh to be visible after creation

ssOlexBaiko:
port from https://github.com/zenoss/zenoss-prodbin/pull/2312
[Jira](https://jira.zenoss.com/browse/ZEN-27394)


ssOlexBaiko:
:)

PR #2317, opened 2017-05-12
fix ZEN-27470: deploy zenreports as new service; remove extra zproxy svcdef configs

corntoole:


corntoole:
retest this

tgjonestx:
The migration itself looks good - my only suggestion is to move the .json file for the service definition into a subdirectory like `migrate/data`. Delivering JSON embedded as part of the migration code is something new, so keeping those files isolated from the actual python code seems a little cleaner.

corntoole:
Thanks, I made the change.

PR #2324, opened 2017-05-17
ZEN-27507: delete relationship between products and product classes t…

fraibacas:
…o reduce ConflictErrors

### DIFF
--- a/Products/ZenModel/ProductClass.py
+++ b/Products/ZenModel/ProductClass.py
@@ -31,6 +31,8 @@ from ZenPackable import ZenPackable
 
 from Products.ZenRelations.RelSchema import *
 
+from Products.Zuul.catalog.interfaces import IModelCatalogTool
+
 class ProductClass(ZenModelRM, ZenPackable):
     implements(IIndexed)
     meta_type = "ProductClass"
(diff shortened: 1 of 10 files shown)


Products/ZenModel/ProductClass.py
```diff
@@ -108,18 +110,27 @@ def __init__(self, id, title="", prodName=None,
...
+        if count:
+            return model_catalog.search(query=query, limit=0)
+        else:
+            return model_catalog.search(query=query)
 
-    def count(self):
-        """Return the number of existing instances for this class.
-        """
-        return self.instances.countObjects()
+    def instances(self):
```
iancmcc: This is too expensive. Make it a generator.

```
for brain in self._find_instances_in_catalog():
    if brain:
        yield brain.getObject()
```

Products/Zuul/catalog/model_catalog.py
```diff
@@ -131,7 +131,7 @@ def model_index(self):
         return self._data_manager.model_index
 
     def _get_forbidden_classes(self):
-        return (Software, OperatingSystem)
+        return ()
```
iancmcc: Why?

Products/ZenModel/ProductClass.py
```diff
@@ -52,8 +54,9 @@ class ProductClass(ZenModelRM, ZenPackable):
     )
 
     _relations = ZenPackable._relations + (
-        ("instances", ToMany(ToOne, "Products.ZenModel.MEProduct", "productClass")),
         ("manufacturer", ToOne(ToManyCont,"Products.ZenModel.Manufacturer","products")),
+        # Deprecated
```
iancmcc: The migrate script drops the relationship, so you can just drop this. It isn't deprecated so much as gone.

Products/Zuul/catalog/model_catalog.py
fraibacas: Software and Operating systems are MEProduct. If we don't index them we will not be able to get the instances for the related ProductClass

PR #2325, opened 2017-05-17
ZEN-27520: changed api query to elstic

rderkach:
Number of log files for zope was different
from number of log files that should be
created. In addition, support bundle wasn't
working at all, since we moved to newer
version of ES.
Changed query to elastic, now using scroll api,
file name format now include instance id
instead of host name.

[JIRA](https://jira.zenoss.com/browse/ZEN-27520)


rderkach:
retest

rderkach:
retest

PR #2344, opened 2017-05-31
ZEN-25529 zencommand 'Missed_Runs' increments for devices with state …

romanbakaleyko:
…of 'Down'

### DIFF
--- a/Products/ZenCollector/scheduler.py
+++ b/Products/ZenCollector/scheduler.py
@@ -189,10 +189,8 @@ class CallableTask(object):
         if self.task.state is TaskStates.STATE_PAUSED and not self.paused:
             self.task.state = TaskStates.STATE_IDLE
 
-        if self.task.state is TaskStates.STATE_IDLE:
-            if self.paused:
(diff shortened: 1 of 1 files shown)


romanbakaleyko:
demo in https://jira.zenoss.com/browse/ZEN-25529

Products/ZenCollector/scheduler.py
```diff
@@ -189,10 +189,8 @@ def __call__(self):
         if self.task.state is TaskStates.STATE_PAUSED and not self.paused:
             self.task.state = TaskStates.STATE_IDLE
 
-        if self.task.state is TaskStates.STATE_IDLE:
-            if self.paused:
-                self.task.state = TaskStates.STATE_PAUSED
```
jplouis: It seems this might be the only place the TaskStates.STATE_PAUSED was ever assigned.  Can you please verify for me that the `PAUSED` state is set when there is a ping down event for collectors other than zencommand?

You can run `zenperfsnmp stats` to get the output of the tasks in zenperfsnmp.

PR #2349, opened 2017-05-31
Add tool to reindex solr

spopezen:
This is a tool that offers two ways to reindex solr;
- Hard reindex, which clears solr model index, then navigates the entire ZODB tree and indexes the appropriate data
- Soft reindex, which reads uids from solr, then looks them up in ZODB and indexes the appropriate data

RAM usage rockets quite high during a reindex, so that would be a good place to find improvement.

Edit:
Merge simultaneously with https://github.com/zenoss/product-assembly/pull/403


fraibacas:
This needs to replace the previous reindexing tool. We need to either rename it or update product_assembly to use the new tool: https://github.com/zenoss/product-assembly/blob/bdc23de4c611264239a6107d9e37c4194550b95e/product-base/install_scripts/install_lib.sh#L213

Products/Zuul/catalog/reindex_solr.py
```diff
@@ -0,0 +1,308 @@
...
+    Index all items in list if fit.
+    Repeat.
+    """
+    def __init__(self, queue, idx, parent_queue):
+        super(SoftReindex, self).__init__(queue, idx, parent_queue)
+
+    def get_work(self):
+        yield self.parent_queue.get(timeout=30)
+        while True:
+            yield self.parent_queue.get(timeout=2)
```
fraibacas: add a comment to clarify the method raises QueueEmpty excpetion or add a try/except and return empy list when there are no more elements in the queue. Same for the hard version

Products/Zuul/catalog/reindex_solr.py
```diff
@@ -0,0 +1,308 @@
...
+            # Mandatory to commit before we bail
+            self.index(batch, True)
+            return WorkerReport(self.idx, count)
+        except Queue.Empty:
+            # The parent_queue is empty, which means we're done
+            self.index(batch, True)
+            return WorkerReport(self.idx, count)
+        except Exception as e:
+            # grab the traceback for the report
+            return WorkerReport(self.idx, count - len(batch), traceback.format_exc())
```
spopezen: Addressed in `do_work` loop

Products/Zuul/catalog/reindex_solr.py
```diff
@@ -0,0 +1,320 @@
+import Globals
```
fraibacas: Please add copyright header

Products/Zuul/catalog/reindex_solr.py
```diff
@@ -257,14 +260,12 @@ def run(processor_count=8, hard=False):
...
+        index_client.clear_data()
         work.append("/zport")
         Worker = HardReindex
     else:
         log.info("Reading uids from solr")
-        uids = [brain.uid for brain in get_uids()]
-        uids.sort()
-        work = uids
+        work = get_uids(index_client)
         Worker = SoftReindex
```
fraibacas: Is solr data cleared for Soft reindex?

Products/Zuul/catalog/reindex_solr.py
```diff
@@ -0,0 +1,320 @@
...
+                    current_node = self.dmd.unrestrictedTraverse(current_uid)
+                except NotFound:
+                    log.error("'{0}' was not found in ZODB.  A hard reindex or investigation may be needed.".format(current_uid))
+                    continue
+                self.push_children(current_node)
+                if self.include_node(current_node):
+                    batch.append(current_node)
+                    count += 1
+                    # the increment on counter is non-atomic, but we don't care
+                    self.counter.value += 1
```
fraibacas: Shouldn't we also pass a lock to the process and use it? Also, it would be probably better to update the lock every, let's say, 1k elements.
I prefer the approach of having the worker send a msg to the main process (we have the queue already set up) instead of using a shared variable.

fraibacas:
Please delete the old tool

Products/Zuul/catalog/reindex_solr.py
spopezen: Done

PR #2355, opened 2017-06-05
ZEN-27611 zencommand 'Missed_Runs' increments for devices with state …

romanbakaleyko:
…of 'Down'


romanbakaleyko:
port https://github.com/zenoss/zenoss-prodbin/pull/2344/files

PR #2358, opened 2017-06-06
the test is now for a queue with both messages and no consumers

sockterrier:


bin/metrics/gather.py
```diff
@@ -139,15 +139,16 @@ def _extract_data(self, queue, timestamp, without_consumers_aggregate):
         # aggregate these
         for qtype in self.HUB_QUEUE_TYPES:
             if queue['name'].startswith('zenoss.queues.hub.%s' % qtype):
-                without_consumers_aggregate[qtype] += 1 if not queue.get('consumers', 0) else 0
+                if queue.get('consumers', 0) is 0 and queue.get('messages', 0) is not 0:
```
ecifreo: Should be '==' instead of 'is' when checking for value equality? I just refreshed my python operator knowledge, and it appears since this is a small int being checked, it will work. What do you think

PR #2367, opened 2017-06-08
add ability to set tracer tags on metrics

morr-zenoss:
Add "Tracer Bullet" functionality for ZEN-27694 in support of debugging ZEN-27365
Coordinate this change with pulls on [metricshipper](https://github.com/zenoss/metricshipper/tree/feature/ZEN-27694) and [metricconsumer](https://github.com/zenoss/zenoss.metric.consumer/pull/97) under the same branch name


Products/ZenCollector/daemon.py
```diff
@@ -359,6 +370,25 @@ def generateEvent(self, event, **kw):
...
+        :param contextkey: context key of the metric in question
+        :return: boolean indicating whether to trace this metric/key
+        """
+        tests = []
+        if self.options.traceMetricName:
+            tests.append((self.options.traceMetricName, metric))
+        if self.options.traceMetricKey:
+            tests.append((self.options.traceMetricKey, contextkey))
+
+        result = [bool(re.compile(exp).match(subj)) for exp, subj in tests]
```
jplouis: If you are going to compile the expression for efficiency it should be done only once.

jplouis: Also see https://docs.python.org/2/library/re.html#search-vs-match Any reason to use match instead of search?

morr-zenoss: I've taken out the compile() - per the note [here](https://docs.python.org/2/library/re.html#re.compile) and [this](https://stackoverflow.com/questions/452104/is-it-worth-using-pythons-re-compile) StackOverflow thread, search() and match() cache up to MAXCACHE (=100) expressions. So, this will compile the regex the first time it's used, and use the cached compiled version thereafter.

morr-zenoss:
@jplouis : I've updated again (thanks for the catch) - I just took out the compile so the caching that is done by re.search() will be used.

PR #2372, opened 2017-06-09
ZEN-27602 Filter should be available for Device Status column

romanbakaleyko:


romanbakaleyko:
please see comments in JIRA

PR #2374, opened 2017-06-12
Revert "ZEN-27507: delete relationship between products and product classes t…"

fraibacas:
Reverts zenoss/zenoss-prodbin#2324


fraibacas:
Reverting this changes bc we found out that the root cause of this issue was at a lower level in  the ZenRelation code: https://github.com/zenoss/zenoss-prodbin/pull/2361

PR #2378, opened 2017-06-12
Add migration for Kibana auth checks

jpeacock-zenoss:


Products/ZenModel/migrate/updateNginxKibanaRule.py
```diff
@@ -0,0 +1,82 @@
+    ##############################################################################
```
morr-zenoss: check spacing

PR #2389, opened 2017-06-15
ZEN-27731: vSphere devices are not filtered with Down status on Infra…

olst:
…structure page

### DIFF
--- a/Products/Zuul/facades/__init__.py
+++ b/Products/Zuul/facades/__init__.py
@@ -279,12 +279,9 @@ class TreeFacade(ZuulFacade):
         devices = list(imap(IInfo, imap(unbrain, brains)))
         if isinstance(params, dict):
             statuses = params.pop('status', None)
-            # Don't filter if we want to see devices with both states UP and DOWN what is set by default
+            # Don't filter if we want to see devices with all statuses UP, DOWN and UNKNOWN what is set by default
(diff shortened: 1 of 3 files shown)


Products/Zuul/facades/__init__.py
```diff
@@ -280,8 +280,11 @@ def getDevices(self, uid=None, start=0, limit=50, sort='name', dir='ASC',
         if isinstance(params, dict):
             statuses = params.pop('status', None)
             # Don't filter if we want to see devices with both states UP and DOWN what is set by default
-            if statuses is not None and len(statuses) < 2:
-                devices = [d for d in devices if d.status in statuses]
+            if statuses is not None and len(statuses) < 3:
+                if None in statuses:
+                    devices = [d for d in devices if (d.status in statuses or d.status is None)]
+                else:
+                    devices = [d for d in devices if d.status in statuses]
```
jpeacock-zenoss: This is more complicated that it needs to be and can be rewritten more simply. Read the following for why it can be simplified:
```
>>> None in [True, False]
False
>>> None in [True, None]
True
>>> None in [False, None]
True
>>> None in [True]
False
>>> None in [False]
False
>>> None in [None]
True
>>>
```

olst:
retest this

PR #2395, opened 2017-06-15
add method to perform same thread soft reindexing

fraibacas:


Products/Zuul/catalog/model_catalog_init.py
```diff
@@ -320,13 +322,32 @@ def run(processor_count=8, hard=False):
...
+    start = time.time()
+    log.info("Performing soft reindex on model_catalog. Params = root:'{}' / idxs:'{}'".format(root, idxs))
+    modelindex = init_model_catalog()
+    uids = get_uids(modelindex, root=root)
+    if uids:
+        index_updates = []
+        for uid in uids:
+            try:
+                obj = dmd.unrestrictedTraverse(uid)
+            except KeyError:
```
spopezen: I'm not what the scenario is where it happens, but I believe `dmd.unrestrictedTraverse` can also throw `zExceptions.NotFound` when a path is not found in ZODB

fraibacas: I tested it yesterday and it seems to raise a KeyError when the uid does not exists

PR #2398, opened 2017-06-16
Soft reindexing of specific types of objects

johnlhamilton:


Products/Zuul/catalog/model_catalog_init.py
```diff
@@ -229,12 +230,18 @@ def index(self, batch, commit=True):
     def reconcile(self):
         pass
 
-def get_uids(index_client, root=""):
+def get_uids(index_client, root="", types=()):
```
fraibacas: I believe, there are issues when passing a tuple as default args. Or is it only with lists?

fraibacas: I was wrong xD

PR #2405, opened 2017-06-16
Feature/zen 27478

youngklee:


Products/Zuul/facades/devicefacade.py
```diff
@@ -499,10 +499,10 @@ def resetCommunityString(self, uid):
         dev = self._getObject(uid)
         dev.manage_snmpCommunity()
 
-    def renameDevice(self, uid, newId):
+    def renameDevice(self, uid, newId, retainGraphData):
```
jplouis: This should default to some value to maintain backwards compatibility.

Products/Zuul/facades/metricfacade.py
```diff
@@ -620,3 +676,82 @@ def getMetricsForDevices(self, devices, metrics, start=None,
...
+    def renameDevice(self, oldId, newId, joblog=None):
+        # joblog is injected in the FacadeMethodJob class when this method runs
+        # as a job. The messages written to joblog will be displayed in the Jobs
+        # pane of Zenoss.
+
+        success = False
+
+        try:
+            path = RENAME_URL_PATH
+            request = {'oldId': oldId, 'newId': newId}
```
jplouis: See comments in query. These should probably contain the trailing "/" to prevent renaming the wrong metrics. Also consider breaking up into 2 queries, one to rename metrics and the other to rename tag values.

Products/Zuul/routers/device.py
```diff
@@ -476,7 +476,7 @@ def getDevices(self, uid=None, start=0, params=None, limit=50, sort='name',
         return DirectResponse(devices=data, totalCount=devices.total,
                               hash=devices.hash_)
 
-    def renameDevice(self, uid, newId):
+    def renameDevice(self, uid, newId, retainGraphData):
```
jplouis: This needs to default to maintain backwards compatibility

Products/Zuul/facades/devicefacade.py
youngklee: Revised as suggested.

Products/Zuul/routers/device.py
youngklee: Revised as suggested.

Products/Zuul/facades/metricfacade.py
youngklee: Agree. Working on it.

youngklee: Addressed in https://github.com/zenoss/query/pull/300

Products/Zuul/routers/device.py
```diff
@@ -363,6 +363,14 @@ def _getInfoData(self, info, keys):
         return values
 
     @require('Manage Device')
+    def resumeCollection(self, id):
+        uid = id.split('/')[-1]
+        facade = self._getFacade()
+        device = facade._getObject(id)
+        device.resumeCollection(uid)
```
jplouis: Logic should be moved into the facade. By moving there will also only be one object load instead of two;  replaces the object load and then the search.

Products/ZenModel/DeviceClass.py
```diff
@@ -371,6 +371,10 @@ def moveDevices(self, moveTarget, deviceNames=None, REQUEST=None):
                 numExports += 1
         return numExports
 
+    def resumeCollection(self, devname):
```
jplouis: Move to facade. See comment below.

Products/Zuul/facades/metricfacade.py
```diff
@@ -683,33 +685,101 @@ def renameDevice(self, oldId, newId, joblog=None):
...
+
+            # key tagv for device metrics
+            request = {
+                'oldName': 'Devices/' + oldId,
+                'newName': 'Devices/' + newId,
+                'type': 'tagv',
+                'patternType':'whole'
+            }
+
+            totalFails += processRename()
```
jplouis: Do we not retry on any failures?

Products/ZenModel/Device.py
```diff
@@ -1887,7 +1888,11 @@ def renameDevice(self, newId=None, REQUEST=None, retainGraphData=False):
         if newId == '' or newId == oldId:
             return path
 
-        device = self.dmd.Devices.findDeviceByIdExact(newId)
```
jplouis: You still want to do a find because we are looking for a device with the same id regardless of path

Products/Zuul/facades/devicefacade.py
```diff
@@ -541,6 +541,12 @@ def renameDevice(self, uid, newId, retainGraphData=False):
         # pass in the request for the audit
         return dev.renameDevice(newId, self.context.REQUEST, retainGraphData)
 
+    def resumeCollection(self, id):
+        uid = id.split('/')[-1]
```
jplouis: I'm not sure how this is working. Want a demo tomorrow

Products/Zuul/facades/metricfacade.py
youngklee: We do retry in central query.

Products/ZenModel/Device.py
mayathegreat: changed it back

Products/Zuul/facades/devicefacade.py
mayathegreat: renamed some vars to make this a bit more accurate

mayathegreat: changed to use full uid -- seems to work for both

PR #2422, opened 2017-06-21
Handle queries passed as kwargs to legacy adapters.

johnlhamilton:


Products/Zuul/catalog/legacy.py
```diff
@@ -312,6 +312,22 @@ def search(self, query, sort_index=None, reverse=False, limit=None, **kw):
         model_catalog = self._get_model_catalog()
         translator = self._get_translator()
         search_kw = {}
+        old_fields = translator.get_old_field_names()
+        old_fields.append('uid')
```
fraibacas: you should not need to add the uid. Remove it from here and also from the device catalog test that I accidentally created a query for uid and it is not a valid field for device catalog

Products/Zuul/catalog/tests/testLegacyCatalogAdapters.py
```diff
@@ -115,6 +115,31 @@ def validate_device_catalog(self):
...
+                brains = self.device_catalog(query, **kw)
+                self.assertEqual(len(brains), 1)
+                self.assertEqual(brains[0].getPrimaryId, uid)
+                self.assertEqual(brains[0].id, dev.device.id)
+                self.assertEqual(sorted(brains[0].path), sorted(dev.device.path()))
+
+                # Search with both (query is dict)
+                query = {}
+                query.update(kw)
+                brains = self.device_catalog(query, **kw)
```
fraibacas: add some sample queries with format self.device_catalog(field=value)

PR #2435, opened 2017-06-23
ZEN-27778 Invalid TALES expressions cause notifications to fail

ISPHOST:


ISPHOST:
backport from https://jira.zenoss.com/browse/ZEN-27777

PR #2440, opened 2017-06-26
Remove the 1st part of the prefix since it is specified in the reporter.

ecifreo:


ecifreo:
Closing and reopening again, trigger rebuild.

PR #2457, opened 2017-06-29
Fix race condition in new indexer that led to workers dying prematurely

iancmcc:
There was a race where a parent queue could temporarily be empty depending on the state of other workers, which workers took as a signal that they were done. This PR changes that scenario. Now, workers never die. Instead, they signal the master when they become idle. The master can then investigate whether the exit condition has been achieved -- namely, all workers are idle and the queue is empty.

The logging was also confusing. I removed the WorkerReports in favor of a global counter with a rate. It logs at most once every 10,000 objects indexed. Workers can still send exceptions back up to the master, though.


Products/Zuul/catalog/model_catalog_init.py
```diff
@@ -64,82 +64,86 @@ def get_work(self):
...
-        raise NotImplementedError
+        self.index_client.process_batched_updates(
+            [IndexUpdate(node, op=INDEX) for node in batch],
+            commit=commit)
+
+    def index(self, commit=False):
+        if self._batch:
+            self.batch_update(self._batch, commit)
+            self._batch[:] = []
+	# transaction.abort()  # Allow cache boundaries to be enforced
```
johnlhamilton: Any reason to keep this around?

iancmcc: I guess not.

PR #2477, opened 2017-07-05
fix model index init issues

johnlhamilton:
1. Added new options to soft reindexer that lets you scope reindexing by root and type and perform soft reindexing of specific fields only
2. Fixed a race condition where the soft reindexer would never terminate by making it so that workers hold the semaphore while they are idle (waiting on the parent queue) vs. the opposite.
3. Fixed a race condition where the soft reindexer would terminate prematurely by using a sentinel value to indicate the end of the data (for soft re-indexing only)
4. Fixed an issue where some workers may get killed during their final index batch due to a short timeout on the join() call.
5. Made the time calculation more accurate.

### DIFF
--- a/Products/Zuul/catalog/model_catalog_init.py
+++ b/Products/Zuul/catalog/model_catalog_init.py
@@ -37,6 +37,7 @@ log.setLevel(logging.INFO)
 
 MODEL_INDEX_BATCH_SIZE = 10000
 INDEX_SIZE = 5000
+QUEUE_TIMEOUT = 2
 
 
 class ReindexProcess(multiprocessing.Process):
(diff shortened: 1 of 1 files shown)


Products/Zuul/catalog/model_catalog_init.py
```diff
@@ -109,6 +115,27 @@ def process(self, uid):
...
+        if self.semaphore_acquired:
+            self.semaphore.release()
+            self.semaphore_acquired = False
+
+    def get_work_from_parent(self):
+        try:
+            # We hold the semaphore while we are waiting on the queue
+            # This is safe to call multiple times
+            self.acquire_semaphore()
+            new_work = self.parent_queue.get(timeout=QUEUE_TIMEOUT)
```
iancmcc: This won't release the semaphore in case of exception. Make it a context manager, or put the release in a finally

johnlhamilton: It is intentional that the worker will continue to hold the semaphore until it succeeds in getting an item off the parent queue.  The parent process now terminates when all workers are holding the semaphore (vs. the opposite, as it was previously).  Note that calling acquire_semaphore will not actually do anything if we already have it.

PR #2478, opened 2017-07-06
 ZEN-27877 Improving loading time of Infrastructure page

karolzen:
https://jira.zenoss.com/browse/ZEN-27879


Products/Zuul/facades/__init__.py
```diff
@@ -206,8 +204,8 @@ def getObjectBrains(self, uid=None, start=0, limit=50, sort='name',
 
 
     def getDevices(self, uid=None, start=0, limit=50, sort='name', dir='ASC',
-                   params=None, hashcheck=None):
-
+                   params={}, hashcheck=None):
+        statuses = params.pop('status', None)
```
jpeacock-zenoss: Let's comment this line as its placement here is important to the performance of this method.  I suggest the following:
```
    # Passing a key that's not an index in the catalog will
    # cause a serious performance penalty.  Remove 'status' from
    # the 'params' parameter before passing it on.
    statuses = params.pop('status', None)
```

karolzen: Good point! :+1:

Products/Zuul/facades/__init__.py
```diff
@@ -207,7 +205,11 @@ def getObjectBrains(self, uid=None, start=0, limit=50, sort='name',
 
     def getDevices(self, uid=None, start=0, limit=50, sort='name', dir='ASC',
                    params=None, hashcheck=None):
-
+        # Passing a key that's not an index in the catalog will
+        # cause a serious performance penalty.  Remove 'status' from
+        # the 'params' parameter before passing it on.
+        params = params if params else {}
```
jpeacock-zenoss: This line should go before the comment.

Products/Zuul/facades/__init__.py
```diff
@@ -172,8 +172,6 @@ def getObjectBrains(self, uid=None, start=0, limit=50, sort='name',
         query = None
         globFilters = {}
         prodStates = None
-        if params is None:
-            params = {}
```
jpeacock-zenoss: This should also be restored (or replaced with `params = params if params else {}`.  `getObjectBrains` can be called from other places as well.

karolzen: Done

Products/Zuul/facades/__init__.py
karolzen: Done

PR #2481, opened 2017-07-07
ZEN-26521: Zenoss should warn when collection fails due to an IP addr…

olst:
…ess not being configured on a device


jpeacock-zenoss:
You're updates in the ticket suggest that the manage IP disappears between modeling and collection.  If that's the case, then simply logging that fact isn't enough to address the issue.

olst:
Yes, I was not sure about this, but IP does not disappear. According to reproduction steps, we don't have it from the beginning. It can be when we add a device by host name, but not by IP address.

PR #2483, opened 2017-07-07
update 'zencatalog' comand to work with model catalog

johnlhamilton:
Part of ZEN-27671

```
$ zencatalog --help
Usage: /opt/zenoss/bin/zencatalog {run|start|stop|restart|status|help|genconf|genxmlconfigs|debug|stats} [options]

  where the commands are:

    run     - start the program but don't put it in the background.
              NB: This mode is good for debugging.

    start   - start the program in daemon mode -- running in the background,
              detached from the shell

    stop    - stop the program

    restart - stop and then start the program
              NB: Sometimes the start command will run before the daemon
                  has terminated.  If this happens just re-run the command.

    status  - Check the status of a daemon.  This will print the current
              process number if it is running.

    help    - display the options available for the daemon

    genconf - create an example configuration file with default settings 

    genxmlconfigs - create an XML file with default settings

    debug   - toggle the logging of daemons between Debug level and the default

    stats   - display detailed statistics of the deamon
```
```
$ zencatalog help
Usage: zencatalog.py [options]

Options:
  --version             show program's version number and exit
  -h, --help            show this help message and exit
  -C CONFIGFILE, --configfile=CONFIGFILE
                        Use an alternate configuration file
  --genconf             Generate a template configuration file
  --genxmltable         Generate a Docbook table showing command-line
                        switches.
  --genxmlconfigs       Generate an XML file containing command-line switches.
  --uid=UID             User to become when running default:zenoss
  -c, --cycle           Cycle continuously on cycleInterval from Zope
  -D, --daemon          Launch into the background
  --duallog             Log to console and log file
  --logfileonly         Log to log file and not console
  --weblog              output log info in HTML table format
  --watchdog            Run under a supervisor which will restart it
  --watchdogPath=WATCHDOGPATH
                        The path to the watchdog reporting socket
  --starttimeout=STARTTIMEOUT
                        Wait seconds for initial heartbeat
  --socketOption=SOCKETOPTION
                        Set listener socket options. For option details: man 7
                        socket
  --heartbeattimeout=HEARTBEATTIMEOUT
                        Set a heartbeat timeout in seconds for a daemon
  --createcatalog       Create global catalog and populate it
  --forceindex          works with --createcatalog to re-create index, if
                        catalog exists it will be dropped first
  --reindex             reindex existing catalog
  --permissionsOnly     Only works with --reindex, only update  permissions
  --resume              DEPRECATED
  --clearmemcached      clears memcached after processing
  --zodb-cacheservers=ZODB_CACHESERVERS
                        memcached servers (used for --clearmemcached option
  --workers=WORKERS     Number of processes working simultaneously
  --buffersize=BUFFERSIZE
                        DEPRECATED
  --inputqueuesize=INPUTQUEUESIZE
                        DEPRECATED
  --processedqueuesize=PROCESSEDQUEUESIZE
                        DEPRECATED

  Logging Options:
    -v LOGSEVERITY, --logseverity=LOGSEVERITY
                        Logging severity threshold
    --logpath=LOGPATH   Override the default logging path; default
                        $ZENHOME/log
    --maxlogsize=MAXLOGKILOBYTES
                        Max size of log file in KB; default 10240
    --maxbackuplogs=MAXBACKUPLOGS
                        Max number of back up log files; default 3

```


Products/Zuul/catalog/model_catalog_init.py
```diff
@@ -311,12 +342,13 @@ def init_model_catalog(collection_name=ZENOSS_MODEL_COLLECTION_NAME):
     index_client.init(config)
     return index_client
 
-def run(processor_count=8, hard=False, root="", indexes=None, types=[]):
+
+def run(processor_count=8, hard=False, root="", indexes=None, types=[], terminate=None, toggle_debug=None):
```
iancmcc: Technically `types` shouldn't be a literal mutable there, but this is only called once so it's academic.

johnlhamilton: I'm gonna fix it anyway.

PR #2487, opened 2017-07-10
ZEN-26681 Excessive logging in opentsdb

ISPHOST:


Products/ZenModel/migrate/addOpenTSDBLogbackConfig.py
```diff
@@ -0,0 +1,65 @@
...
+                name="/opt/opentsdb/src/logback.xml",
+                filename="/opt/opentsdb/src/logback.xml",
+                owner="root:root",
+                permissions="0664",
+                 content=content
+            )
+
+            # If there's a config with the same name but is different from
+            # the new config, overwrite it.
+            if all([not equal(config, newConfig) for config in service.originalConfigs]):
```
jpeacock-zenoss: The use of `[...]` is unnecessary here.  The `all` function support generator comprehensions.  So this can be rewritten as
```
if all(not equal(config, newConfig) for config in service.originalConfigs):
```

Products/ZenModel/migrate/addOpenTSDBLogbackConfig.py
```diff
@@ -0,0 +1,65 @@
...
+                service.originalConfigs.append(newConfig)
+                changed = True
+                log.info("Adding a configuration to OriginalConfigs of %s", service.name)
+
+            # Add this config only if there's no config with the same name.
+            # If there is such config, honor it.
+            if all([not config.name == newConfig.name for config in service.configFiles]):
```
jpeacock-zenoss: Same here.  Use a generator comprehension.

PR #2489, opened 2017-07-10
Migrate the zenpython service def

achow-zenoss:


Products/ZenModel/migrate/updateZenPythonSvc.py
```diff
@@ -0,0 +1,64 @@
...
+
+            # the graph config part
+            commit = False
+            for g_conf in m_prof.graphConfigs:
+                for p in g_conf.datapoints:
+                    if p.rate == True and ( not p.rateOptions ):
+                        commit = True
+                        p.rateOptions = {
+                                "counter": True,
+                                "resetThreshold": 1048576000
```
jplouis: What is special about this number?

achow-zenoss: It is an arbitrary number that is small comparing to the wrap around anomaly but at the same time large enough that will not block a valid data value.
Any suggestion of a better number will be fine.

csmouse:
Is this PR still relevant @achow-zenoss ?

PR #2490, opened 2017-07-10
Use metrology for daemon metrics. Create twisted enabled MetricReport…

jplouis:
…er for metrology metrics


Products/ZenHub/PBDaemon.py
```diff
@@ -864,7 +886,6 @@ def sendEvent(self, event, **kw):
         # deprecated counter name
         self.counters['eventCount'] += 1
         # prefer this counter name for future refs
```
ecifreo: these comments about counter names no longer makes sense

PR #2493, opened 2017-07-11
ZEN-27287: Add migrate script to add Solr service and health checks

iancmcc:


Products/ZenModel/migrate/addSolrService.py
```diff
@@ -0,0 +1,153 @@
...
+import servicemigration as sm
+sm.require("1.0.0")
+from servicemigration import HealthCheck
+
+
+class AddSolrService(Migrate.Step):
+    """
+    Add Solr service and associated healthchecks.
+    """
+    version = Migrate.Version(116, 0, 0)
```
fraibacas: Are we incrementing the migrate version every time we add a new migrate script? The highest migrate script in develop currently is 115

iancmcc: The 115 ones look like pre-Thebe stuff (api zopes, e.g.). But also it doesn't matter much.

Products/ZenModel/migrate/addSolrService.py
```diff
@@ -0,0 +1,153 @@
...
+                    break
+
+        if changed:
+            ctx.commit()
+
+
+def default_solr_service():
+    return {
+    "CPUCommitment": 2,
+    "Command": "setuser zenoss /opt/solr/zenoss/bin/start-solr -cloud -Dbootstrap_confdir=/opt/solr/server/solr/configsets/zenoss_model/conf -Dcollection.configName=zenoss_model",
```
fraibacas: we have a ticket open to add a new parameter to the startup command. Maybe we could add it to the migrate now?
https://jira.zenoss.com/browse/ZEN-27900
-Dsolr.jetty.request.header.size=MAX_HEADER_SIZE. the default seems to not to be good enough

iancmcc: Can we do that without knowing the value yet? We can update the migrate script once that value is determined, I think.

iancmcc:
retest this

PR #2506, opened 2017-07-12
ZEN-27932: change debug to error log and raise an exception

rderkach:
When we were adding CC host first as /Server/SSH/Linux or
/Server/Linux device class and then using the VIP to
the /Control Center device class The CC device will not
appear even though the job completes successfully.

A device can't be added since the VIP was already assigned
to device when it was first added and modeled as
/Server/SSH/Linux or /Server/Linux. An error wasn't raised
since exception was catched. Now give an error log and raise
an exception for job to be failed.

[JIRA&DEMO](https://jira.zenoss.com/browse/ZEN-27932)

### DIFF
--- a/Products/ZenModel/ZDeviceLoader.py
+++ b/Products/ZenModel/ZDeviceLoader.py
@@ -234,11 +234,12 @@ class CreateDeviceJob(Job):
             return '/'.join(device.getPhysicalPath())
         except DeviceExistsError as e:
             transaction.abort()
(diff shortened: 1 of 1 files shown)


rderkach:
Expected result in the ticket say that 
_**The CC device should be added and appear, or there should be a failure.**_
The failure looks more logical, since the shouldn't be an an that is an interface of one and a manage IP of another.

PR #2512, opened 2017-07-12
Add __len__ to legacy adapters

johnlhamilton:
Part of ZEN-27671 because zennetworkclean uses it.


Products/Zuul/catalog/legacy.py
```diff
@@ -249,6 +249,10 @@ def _get_model_catalog(self):
     def __call__(self, query=None, **kw):
         return self.search(query, **kw)
 
+    def __len__(self):
+        model_catalog = self._get_model_catalog()
+        return len(model_catalog.search())
```
fraibacas: Would `return model_catalog.search(limit=0).total` perform better?

PR #2533, opened 2017-07-18
Hide Password Field in Audit Log Messages

cchristiezen:
When editing multiline credentials, the credentials are now masked
in the audit log.


cchristiezen:
retest this

PR #2539, opened 2017-07-18
ZEN-27914: Zenhub connects to collectorredis

youngklee:
Collectorredis is changed to have master and slave instances, and zenhub is changed to connect to the master instance among them.

### DIFF
--- a/Products/ZenHub/metricpublisher/publisher.py
+++ b/Products/ZenHub/metricpublisher/publisher.py
@@ -126,7 +126,7 @@ class RedisListPublisher(BasePublisher):
     """
 
     def __init__(self,
-                 host='localhost',
+                 host='rd1',
                  port=defaultRedisPort,
                  buflen=defaultMetricBufferSize,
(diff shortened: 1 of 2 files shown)


Products/ZenHub/metricpublisher/publisher.py
```diff
@@ -126,7 +126,7 @@ class RedisListPublisher(BasePublisher):
     """
 
     def __init__(self,
-                 host='localhost',
+                 host='rd1',
```
jplouis: I'm not sure how this ensures `rd1` is the host used.  See https://github.com/zenoss/zenoss-prodbin/blob/3919975aba0edf15ff3203a8546f1561b447a34b/Products/ZenHub/PBDaemon.py#L611-L612

jplouis:
Young and I figured out that using `import_all` lets us use `localhost:6379` and it will alway connect to the 0 instance of redis.  This change is not needed.

PR #2541, opened 2017-07-18
ZEN-28017: use the global_catalog helper in CatalogTool to inject glo…

fraibacas:
…balcatalog fields when calling search or getBrain

### DIFF
--- a/Products/Zuul/tree.py
+++ b/Products/Zuul/tree.py
@@ -236,7 +236,7 @@ class CatalogTool(object):
         self.model_catalog = IModelCatalogTool(self.context)
 
     def getBrain(self, path):
-        return self.model_catalog.getBrain(path)
+        return self.model_catalog.global_catalog.getBrain(path)
 
     def parents(self, path):
         return self.model_catalog.parents(path)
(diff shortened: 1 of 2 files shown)


Products/Zuul/tree.py
```diff
@@ -236,7 +236,7 @@ def __init__(self, context):
         self.model_catalog = IModelCatalogTool(self.context)
 
     def getBrain(self, path):
-        return self.model_catalog.getBrain(path)
+        return self.model_catalog.global_catalog.getBrain(path)
```
johnlhamilton: Should we maybe just put the fields in manually here, rather than use the legacy helper in platform code?

fraibacas: that is not a legacy catalog. it is a helper:
https://github.com/zenoss/zenoss-prodbin/blob/develop/Products/Zuul/catalog/model_catalog_tool.py#L69-L71
https://github.com/zenoss/zenoss-prodbin/blob/develop/Products/Zuul/catalog/model_catalog_tool_helper.py#L14

PR #2542, opened 2017-07-18
Add decoder for numerical values

lundybernard:
Fixes ZEN-27876

* Add test case for `int` and `long` types
* Add decoder that returns the unaltered value for numerics

### DIFF
--- a/Products/ZenEvents/zentrap.py
+++ b/Products/ZenEvents/zentrap.py
@@ -775,31 +775,30 @@ class Decoders:
         try:
             for decoder in decoders:
                 out = decoder(value)
-                if out or out == 0:
+                if out is not None:
                     return out
         except Exception as err:
             log.exception("Unexpected exception: %s", err)
 
     @staticmethod
     def dateandtime(value):
-        """
-            Tries converting a DateAndTime value to a printable string.
+        """Tries converting a DateAndTime value to a printable string.
 
-            A date-time specification.
-            field  octets  contents                  range
-            -----  ------  --------                  -----
-            1      1-2     year*                     0..65536
-            2        3     month                     1..12
-            3        4     day                       1..31
-            4        5     hour                      0..23
-            5        6     minutes                   0..59
-            6        7     seconds                   0..60
-                          (use 60 for leap-second)
-            7        8     deci-seconds              0..9
-            8        9     direction from UTC        '+' / '-'
-            9       10     hours from UTC*           0..13
-            10      11     minutes from UTC          0..59
-            """
+        A date-time specification.
+        field  octets  contents                  range
+        -----  ------  --------                  -----
+        1      1-2     year*                     0..65536
+        2        3     month                     1..12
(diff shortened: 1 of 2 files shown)


Products/ZenEvents/zentrap.py
```diff
@@ -762,71 +766,79 @@ def decode(value):
...
+            3        4     day                       1..31
+            4        5     hour                      0..23
+            5        6     minutes                   0..59
+            6        7     seconds                   0..60
+                          (use 60 for leap-second)
+            7        8     deci-seconds              0..9
+            8        9     direction from UTC        '+' / '-'
+            9       10     hours from UTC*           0..13
+            10      11     minutes from UTC          0..59
+            """
```
jpeacock-zenoss: This docstring is indented one level too far.

Products/ZenEvents/zentrap.py
```diff
@@ -762,71 +766,79 @@ def decode(value):
...
+                    else:
+                        utc_dir = '+'
+                    utc_hours = tz_mins / 60
+                    utc_mins = tz_mins % 60
+                return "%04d-%02d-%02dT%02d:%02d:%02d.%d00%s%02d:%02d" % (
+                    year, mon, day, hour, mins, secs,
+                    dsecs, utc_dir, utc_hours, utc_mins)
+
+        except TypeError:
+            pass
```
jpeacock-zenoss: Perhaps these should be logged?  With this, I foresee future "where's the trap?" customer calls and we have nothing in our logs to show for it.

Products/ZenEvents/zentrap.py
```diff
@@ -762,71 +766,79 @@ def decode(value):
...
                     Decoders.encode_base64]
 
-        for decoder in decoders:
-            out = decoder(value)
-            if out:
-                return out
+        try:
+            for decoder in decoders:
+                out = decoder(value)
+                if out or out == 0:
```
jpeacock-zenoss: To be more inclusive of possible valid values, this should probably be: `if out is not None:`
After all, we only _don't_ have a value when `out` is `None`.

Products/ZenEvents/zentrap.py
lundybernard: logging in a decoder would produce an excessive amount of noise in the log file.
Unexpected exceptions are caught and logged in the decode_snmp_value method.

Products/ZenEvents/zentrap.py
```diff
@@ -835,26 +822,55 @@ def oid(value):
...
+             Decoders.dateandtime,
+             Decoders.encode_base64]
+
+
+def decode_snmp_value(value):
+    """Given a raw OID value
+    Itterate over the list of decoder methods in order
+    Returns the first value returned by a decoder method
+
+    NOTE: The order of decoders in the list determines their priority
```
jpeacock-zenoss: Should move this note  up to the `_decoders` definition above as a comment.

PR #2545, opened 2017-07-19
ZEN-27668 Deleting Components Leaves an Incorrect Count on ToManyRela…

romanbakaleyko:
…tionshipBase Objects


romanbakaleyko:
I've done that following the same logic i found here https://github.com/zenoss/zenoss-prodbin/pull/2545/files#diff-cb6c8693e13745d8dfc44017ecc8de64R408

PR #2555, opened 2017-07-21
ZEN-27728 Modeling devices fails when device name contains parentheses.

romanbakaleyko:


romanbakaleyko:
DEMO in the ticket https://jira.zenoss.com/browse/ZEN-27728

PR #2559, opened 2017-07-25
ZEN-13979 Do not monitor interface if admin status is down

ssOlexBaiko:
[Jira](https://jira.zenoss.com/browse/ZEN-13979)


Products/ZenModel/IpInterface.py
```diff
@@ -581,6 +581,14 @@ def getRRDTemplates(self):
         return []
 
 
+    def monitored(self):
+        '''
+        Return True if this instance should be monitored. False
+        otherwise.
+        '''
+        return False if self.adminStatus != 1 else self.monitor
```
jplouis: You should call the super method in the else clause incase the base class implementation ever changes.

Products/ZenModel/IpInterface.py
```diff
@@ -581,6 +581,14 @@ def getRRDTemplates(self):
...
 
 
+    def monitored(self):
+        '''
+        Return True if this instance should be monitored. False
+        otherwise.
+        '''
+        return False if self.adminStatus != 1 else super(IpInterface, self).monitor
+
+
```
jplouis: super(IpInterface, self).monitored()

PR #2566, opened 2017-07-27
ZEN-26681 Excessive logging in opentsdb

ISPHOST:


Products/ZenModel/migrate/addOpenTSDBLogbackConfig.py
```diff
@@ -0,0 +1,72 @@
...
+log = logging.getLogger("zen.migrate")
+
+import Migrate
+from Products.ZenUtils.Utils import zenPath
+import servicemigration as sm
+sm.require("1.0.0")
+
+class addOpenTSDBLogbackConfig(Migrate.Step):
+    """ Set Editable Logback configuration file for OpenTSDB. See ZEN-26681 """
+    version = Migrate.Version(115, 0, 0)
```
tgjonestx: You can use 114 instead of 115 because the highest version in the most recent release (5.2.6) was 1.1.3.  All of the existing migration scripts that use 114 are targeted for the 5.3.0 so this one can use the same version

PR #2576, opened 2017-08-01
ZEN-28165: Instrument zenmodeler for self-monitoring

youngklee:
modeledDevices and the number of failures are kept track of.

### DIFF
--- a/Products/DataCollector/zenmodeler.py
+++ b/Products/DataCollector/zenmodeler.py
@@ -62,7 +62,6 @@ import traceback
 from random import randint
 from itertools import chain
 from metrology import Metrology
-from metrology.instruments import Gauge
 
 defaultPortScanTimeout = 5
 defaultParallel = 1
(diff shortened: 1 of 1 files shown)


Products/DataCollector/zenmodeler.py
```diff
@@ -61,6 +61,8 @@
 import traceback
 from random import randint
 from itertools import chain
+from metrology import Metrology
+from metrology.instruments import Gauge
```
jplouis: This doesn't need to be imported

Products/DataCollector/zenmodeler.py
```diff
@@ -710,10 +715,12 @@ def processClientFinished(result):
             @type result: object
             """
             self.counters['modeledDevicesCount'] += 1
+            self._modeledDevicesMetric.mark(self.counters['modeledDevicesCount'])
```
jplouis: You mark a meter with the delta, not an entire new value

Products/DataCollector/zenmodeler.py
youngklee: Gauge doesn't need to be, but Metrology does.

Products/DataCollector/zenmodeler.py
youngklee: Revised as suggested.

Products/DataCollector/zenmodeler.py
youngklee: Removed import Gauge

PR #2579, opened 2017-08-02
ZEN-24255: add new rpn field to threshold

rderkach:
It was confusing for user what value should he enter in min/max
field, since after that value was calculated with the same rpn
of the datapoint threshold is related to.
Added new field to to Threshold UI which show the rpn expression
that will be used to threshold min/max value. This will help to
adopt the value to rpn expression.

[JIRA&DEMO](https://jira.zenoss.com/browse/ZEN-24255)

Related PR https://github.com/zenoss/ZenPacks.zenoss.PredictiveThreshold/pull/24

### DIFF
--- a/Products/ZenModel/ThresholdClass.py
+++ b/Products/ZenModel/ThresholdClass.py
@@ -92,3 +92,16 @@ class ThresholdClass(ZenModelRM, ZenPackable):
             else:
                 names.append('%s(<span style="color: red">missing</span>)' % dsName)
         return ','.join(names)
+
+
+    def getDataPointsWithRPN(self, all=False):
+        """
+        Return a dictionary where key is datapoint name and value
+        is rpn used to that datapoint.
+        """
+        dpsrpn = []
+        dpsnames = self.rrdTemplate.getRRDDataPointNames() if all else self.dsnames
+        for point in dpsnames:
+            for graph in self.rrdTemplate.getGraphDefs():
+                dpsrpn.extend([(point, pobj.rpn if pobj.rpn else 'No RPN') for pobj in graph.getDataPointGraphPoints(point)])
+        return dict(dpsrpn)
--- a/Products/ZenUI3/browser/resources/js/zenoss/form/DataPointItemSelector.js
+++ b/Products/ZenUI3/browser/resources/js/zenoss/form/DataPointItemSelector.js
@@ -13,6 +13,19 @@
 Ext.define("Zenoss.form.DataPointItemSelector", {
     alias:['widget.datapointitemselector'],
     extend:"Ext.ux.form.ItemSelector",
+    listeners: {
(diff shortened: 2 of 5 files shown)


Products/ZenModel/ThresholdClass.py
```diff
@@ -92,3 +92,16 @@ def getDataPointNamesString(self):
...
+    def getDataPointsWithRPN(self, all=False):
+        """
+        Return a dictionary where key is datapoint name and value
+        is rpn used to that datapoint.
+        """
+        dpsrpn = []
+        dpsnames = self.rrdTemplate.getRRDDataPointNames() if all else self.dsnames
+        for point in dpsnames:
+            for graph in self.rrdTemplate.getGraphDefs():
+                dpsrpn.extend([(point, pobj.rpn if pobj.rpn else 'No RPN') for pobj in graph.getDataPointGraphPoints(point)])
```
jpeacock-zenoss: the `extend` method can accept an iterable type, so a generator comprehension can be used here, e.g.
```
    dpsrpn.extend(
        (point, pobj.rpn if pobj.rpn else 'No RPN')
        for pobj in graph.getDataPointGraphPoints(point)
    )
```
_note the lack of square brackets_

Products/Zuul/interfaces/template.py
```diff
@@ -148,6 +148,8 @@ class IMinMaxThresholdInfo(IThresholdInfo):
     """
     minval = schema.TextLine(title=_t(u'Minimum Value'), order=6)
     maxval = schema.TextLine(title=u'Maximum Value', order=7)
+    rpnused = schema.Text(title=_t(u'RPN is used'),
```
jpeacock-zenoss: I suggest that "DataPoint RPN(s) in Use" (or just "DataPoint RPN(s)") is a clearer title.

Products/ZenUI3/browser/resources/js/zenoss/form/DataPointItemSelector.js
```diff
@@ -13,6 +13,19 @@
...
+        change: function(obj, event) {
+            rpnTextObj = Ext.ComponentQuery.query('[name=rpnused]')[0];
+            if (rpnTextObj) {
+                rpnTextObj.setValue('');
+                Ext.each(obj.getValue(), function(dpname, index){
+                    rpnTextObj.setValue(rpnTextObj.value + dpname + ": " + rpnTextObj.record.allDataPoints[dpname] + "; ");
+                });
+            }
+        }
+    },
```
jpeacock-zenoss: When there are multiple datapoints, the presentation of datapoint and its RPN becomes difficult to read.  It'll be easy for users to misread or completely overlook a relevant RPN.  Having a (read-only) listview (i.e. a table) of the datapoint name and corresponding RPN would be an easy-to-read way to present this information to the user.

PR #2581, opened 2017-08-02
ZEN-28160 TALES expression inspection code breaks existing notifications

ISPHOST:


Products/ZenModel/actions.py
```diff
@@ -446,13 +446,18 @@ def executeBatch(self, notification, signal, targets):
                 log.debug("Adding recipients defined in the event %s", mail_targets)
                 targets |= set(mail_targets)
 
+            if notification.content.has_key('skipfails'):
+                skipfails = notification.content['skipfails'];
+            else:
+                skipfails = False
```
jpeacock-zenoss: Assuming that `content` is of type `dict`, this can be written more cleanly as
```
skipfails = notification.content.get('skipfails', False)
```

Products/ZenModel/actions.py
```diff
@@ -559,11 +564,17 @@ def _execute(self, notification, signal, env={}):
         self.setupAction(notification.dmd)
         log.debug('Executing page action: %s', self.name)
         data = self._signalToContextDict(signal, notification)
+
+        if notification.content.has_key('skipfails'):
+            skipfails = notification.content['skipfails']
+        else:
+            skipfails = False
```
jpeacock-zenoss: Assuming that `content` is of type `dict`, this can be written more cleanly as
```
skipfails = notification.content.get('skipfails', False)
```

PR #2588, opened 2017-08-04
ZEN-28106: add option to load infrastructure page device tree from ca…

fraibacas:
…talog


Products/Zuul/infos/catalog_tree_builder.py
```diff
@@ -0,0 +1,136 @@
...
+from Products.Zuul.catalog.model_catalog import OBJECT_UID_FIELD as UID
+
+# TODO: Get the Networks tree to use this.
+
+class ModelCatalogTreeNode(object):
+    def __init__(self, path):
+        self.path = path
+        self.id = path.split("/")[-1]
+        self.child_trees = {}
+        self.partial_leave_count = 0   # count not including children
```
johnlhamilton: Nitpick, but this should probably be "partial_leaf_count"

PR #2593, opened 2017-08-09
ZEN-28005 XSS fix

mayathegreat:
initial commit


Products/Zuul/routers/zep.py
```diff
@@ -919,6 +919,10 @@ def add_event(self, summary, device, component, severity, evclasskey,
         """
         device = device.strip()  # ZEN-2479: support entries like "localhost "
         try:
+            # Fix for ZEN-28005 to thwart XSS attacks from incoming events
```
jplouis: Any reason not to escape the same fields from above?

Products/ZenHub/XmlRpcService.py
```diff
@@ -36,6 +37,11 @@ def __init__(self, dmd):
 
 
     def xmlrpc_sendEvent(self, data):
+        # Fix for ZEN-28005 to thwart XSS attacks from incoming events
+        for f in data:
+            if f in ("eventClassKey", "component", "summary", "device", "eventClass"):
+                if data[f] is not None and len(data[f]) > 0:
+                    data[f] = cgi.escape(data[f])
```
jplouis: Should this be added to zen.sendevent instead to catch other entry points?

Products/Zuul/routers/zep.py
mayathegreat: done

Products/ZenHub/XmlRpcService.py
mayathegreat: there seems to be a few sendevent methods -- is this the specific one you are referring to? https://github.com/zenoss/zenoss-prodbin/blob/develop/Products/ZenEvents/SendEvent.py

jplouis: I meant this one: https://github.com/zenoss/zenoss-prodbin/blob/2cb7c3de2d33a94d6a4c161817ea0cf68bccf03b/Products/ZenEvents/MySqlSendEvent.py#L56-L56

PR #2605, opened 2017-08-14
Fix migration versions

tgjonestx:
The first commit syncs up service migrations on `develop` with the corresponding files on `support/5.3.x`. This insures that customers who migrate from 5.1.x or 5.2.x to Thebe without going through 5.3.x will execute the proper migrations.  In other words, once a particular service migration is in a GA release, the logic on the `develop` branch should be the same (or if changed, then it should have a higher `Migrate.Version()` value than the last GA copy).

The second commit is superceded by the 3rd and 4th commits which implement a new scheme whereby values for `Migrate.Version` are set programmatically instead of manually.

See corresponding changes in zendev here - https://github.com/zenoss/zendev/pull/330


tgjonestx:
Retest this

PR #2606, opened 2017-08-14
Update Local Docs for 5.3.0

lundybernard:
Fixes ZEN-28198


tgjonestx:
Is there some file elsewhere which needs to be updated to reference these particular files by name?

lundybernard:
Judging by ZEN-23016, and grepping through prodbin and product-assembly, it looks like just the binary files need to be updated, nothing else references them directly.

PR #2614, opened 2017-08-16
ZEN-28292 Datasource-related passwords are in plain text inside audit…

romanbakaleyko:
….log


romanbakaleyko:
demo in the ticket https://jira.zenoss.com/browse/ZEN-28292

Products/Zuul/routers/template.py
```diff
@@ -290,6 +291,15 @@ def getDataPointDetails(self, uid):
...
         return data
 
+    def _getPasswordFields(self, obj):
+        """
+        Return fields for obj if it has xtype and it is password.
+        """
+        fields = IFormBuilder(IInfo(obj)).fields()
+        return [field for field in fields
+                        if fields[field].get('xtype', None)
+                        and fields[field].get('xtype') == 'password']
```
tgjonestx: I'm not familiar with this part of the code. Just wondering if the check should cover a wider range of names.
Is  it possible to have fields with names like `Password` or `passwd`? Or maybe if the string contains `password`.

Products/Zuul/routers/template.py
```diff
@@ -306,14 +316,16 @@ def setInfo(self, **data):
...
         oldData = self._getInfoData(obj, data)
         info = self._getFacade().setInfo(uid, data)
         newData = self._getInfoData(obj, data)
         # Trac #29376: Consistently show thresholdType with threshold operations.
         thresholdType = obj.getTypeName() if isinstance(obj, ThresholdClass) else None
         audit(['UI', getDisplayType(obj), 'Edit'], obj, thresholdType=thresholdType,
               data_=newData, oldData_=oldData,
-              skipFields_=('newId',))  # special case in TemplateFacade.setInfo()
+              skipFields_=('newId',), # special case in TemplateFacade.setInfo()
+              askFields_=passwordFields)
```
tgjonestx: Wasn't the property name for blacklist `maskFields`??

Products/Zuul/routers/template.py
romanbakaleyko: all password fields has html xtype=password, so I believe yes it covers all cases.

Products/Zuul/routers/template.py
romanbakaleyko: yes, my bad

PR #2625, opened 2017-08-21
ZEN-27087 Kibana should retrieve Toolbox Logs as well

romanbakaleyko:


Products/ZenModel/migrate/addToolboxLogsToKibana.py
```diff
@@ -0,0 +1,59 @@
...
+
+import Migrate
+import servicemigration as sm
+
+sm.require("1.1.5")
+
+class AddToolboxLogsToKibana(Migrate.Step):
+    "Pull logs provided by toolbox utilities to logstash."
+
+    version = Migrate.Version(115, 0, 0)
```
tgjonestx: There's a new convention for this release regarding the schema versions. See the README in this directory for more information, or grep the other migrations in this directory for `SCHEMA_` to see some examples

Products/ZenModel/migrate/addToolboxLogsToKibana.py
```diff
@@ -0,0 +1,59 @@
...
+                                 isAudit=False)
+            )     
+
+        zopes = filter(lambda s: s.name == "Zope", ctx.services)
+        for zope in zopes:
+            zope.logConfigs.extend(new_logs)
+        
+        ctx.commit()
+
+AddToolboxLogsToKibana()
```
tgjonestx: This migration needs a unit-test

Products/ZenModel/migrate/addToolboxLogsToKibana.py
```diff
@@ -0,0 +1,59 @@
...
+                                 path=log_paths[name],
+                                 logType=logType,
+                                 filters=['pythondaemon'],
+                                 logTags=None, 
+                                 isAudit=False)
+            )     
+
+        zopes = filter(lambda s: s.name == "Zope", ctx.services)
+        for zope in zopes:
+            zope.logConfigs.extend(new_logs)
```
tgjonestx: The migration logic needs to be idempotent. This will break in cases where the `zope.logConfigs` already contains these logs.  In other words, the logic needs to be something like - only add these new logs, if zope.logConfigs does not contain any logs from the `/opt/zenoss/log/toolbox` dirctory

Products/ZenModel/migrate/addToolboxLogsToKibana.py
```diff
@@ -0,0 +1,59 @@
...
+        
+        ctx.commit()
```
tgjonestx: Per the comment above about idempotent, the migration should only call `commit()` in cases where it actually added logs.

PR #2637, opened 2017-08-22
ZEN-28227: add a migration script for new conf in Zope definition

rderkach:
[JIRA&DEMO](https://jira.zenoss.com/browse/ZEN-28227) 

Service def PR https://github.com/zenoss/zenoss-service/pull/446


rderkach:
There is one question regarding this migration:

After I run this migration on a new instance I can see changes in Zope service definition, but on CC UI they only appears when I make some changes to Zope in CC UI, change instance number e.g. Also after those change  to Zope in CC UI I can see the following, which is good. 

`serviceconfig.go:241] Adding config file /opt/zenoss/etc/audit_log_level.conf for service 5tyzvknenzjf8bgvyh8cyfi98`

**Why doesn't this happen when I commit the context of the service definition in migration ?**

PR #2638, opened 2017-08-22
ZEN-26937 Display full class path in device banner

karolzen:
https://jira.zenoss.com/browse/ZEN-26937

### DIFF
--- a/Products/ZenUI3/browser/resources/js/zenoss/itinfrastructure.js
+++ b/Products/ZenUI3/browser/resources/js/zenoss/itinfrastructure.js
@@ -1408,12 +1408,43 @@ Ext.define("Zenoss.InfraDetailNav", {
     }
 });
 
+var truncate = function(cmp_name, width, uid) {
+    var path = uid.replace('/zport/dmd', ''),
+        textwidth = Ext.util.TextMetrics.measure(cmp_name, path).width;
+    if (textwidth < width) {
+        return path;
+    } else {
+        var path_array = path.split('/'),
+            device_class = path_array[path_array.length-1];
+
+        // check if DeviceClass's name length > field's length
+        if (Ext.util.TextMetrics.measure(cmp_name, ('/../'+device_class)).width < width) {
+            path_array[path_array.length-2] = '..';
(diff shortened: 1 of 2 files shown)


Products/ZenUI3/browser/resources/js/zenoss/itinfrastructure.js
```diff
@@ -1408,12 +1408,45 @@ Ext.define("Zenoss.InfraDetailNav", {
...
+        titlePosition: 0,
+        items: [{
+                   xtype:'displayfield',
+                   value: _t('/'),
+                   width: 'auto',
+                   cls: 'x-device-panel-header',
+                   id: 'device_panel_header',
+                   allowBlank : true,
+                   readOnly: true
+               }]
```
jpeacock-zenoss: Please deindent this by two levels.  It doesn't need to be pushed so far to the right.  E.g.
```
    items: [{
        xtype:'displayfield',
        value: _t('/'),
        width: 'auto',
        cls: 'x-device-panel-header',
        id: 'device_panel_header',
        allowBlank : true,
        readOnly: true
    }]
```

Products/ZenUI3/browser/resources/js/zenoss/itinfrastructure.js
```diff
@@ -1408,12 +1408,45 @@ Ext.define("Zenoss.InfraDetailNav", {
     }
 });
 
+var truncate = function(cmp_name, width, uid) {
+    var path = uid.replace('/zport/dmd/', ''),
```
jpeacock-zenoss: In 4.x, where the full path was visible, the path started with a `/`.  You can see this in the screenshots shown in the ticket.

To preserve this, the `replace` should be rewritten as `uid.replace('/zport/dmd', '')`.

Products/ZenUI3/browser/resources/js/zenoss/itinfrastructure.js
karolzen: Ok :+1:

Products/ZenUI3/browser/resources/js/zenoss/itinfrastructure.js
karolzen: Ok, I will change it to 4.x format.

PR #2640, opened 2017-08-22
ZEN-28204 Wrong time format on Events/Event Console

ISPHOST:


lundybernard:
The datetime values for events are not formatted with the user's timezone, and the filter does not work properly.
create an event and note its firstSeen time, change the user's timezone, and note that the event time does not change.

The timestamps (update_time, first_seen, status_change, last_seen) for events are saved as unix timestamps (ex: 1503925262818).  Our objective is to use these values for everything and only format them when they are displayed.

Products/ZenUI3/browser/resources/js/zenoss/zenoss.js
```diff
@@ -1373,13 +1345,17 @@ Ext.apply(Zenoss.date, {
...
         }
         return moment.utc(value, "X").tz(Zenoss.USER_TIMEZONE).format(format);
     }
     return value;
 };
 
+Zenoss.date.renderPickerField = function(date){
+    return Ext.isDate(date) ? moment(date).format(Zenoss.USER_DATE_FORMAT + ' ' + Zenoss.USER_TIME_FORMAT) : date;
+};
+
```
lundybernard: ZEN-28538 replaces Zenoss.date.renderWithTimeZone with Zenoss.render.date

PR #2642, opened 2017-08-22
Cleanup catalogservice removal

spopezen:
https://jira.zenoss.com/browse/ZEN-28341

`find /opt/zenoss/.ZenPacks/ -name ZenPacks.zenoss.CatalogService*` caused the issue in ZEN-28341, as a system could have multiple versions of CatalogService ZenPack, but during investigation, I discovered a few more problems with the migration and fixed those up.

### DIFF
--- a/Products/ZenModel/migrate/addSolrService.py
+++ b/Products/ZenModel/migrate/addSolrService.py
@@ -44,6 +44,13 @@ class AddSolrService(Migrate.Step):
             ctx.deployService(json.dumps(new_solr), infrastructure)
             changed = True
 
+        # Remove the zencatalogservice-uri global conf option from the top service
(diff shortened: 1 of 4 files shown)


spopezen:
retest this

Products/ZenModel/migrate/addSolrService.py
```diff
@@ -62,6 +69,14 @@ def cutover(self, dmd):
                 ctx.services.remove(svc)
                 changed = True
                 continue
+            # Remove zencatalogservice response prereq from zenhub and add solr one
+            if svc.name == "zenhub":
```
johnlhamilton: Might as well check all services, I think?

PR #2645, opened 2017-08-23
Feature/zen 28333 2

sockterrier:


sockterrier:
screen shot 2017-08-23 at 9 45 04 am

Products/ZenUtils/MetricReporter.py
```diff
@@ -30,6 +33,42 @@
...
 
 
+class TimeOnce(object):
+    """
+    a simple context manager to time something and save tag values and
+    a measurement.
+    """
+    def __init__(self, gauge, *args):
+        if len(gauge.tagKeys) != len(args):
+            raise RuntimeError('The number of tag values provided does not match the number of configured tag keys')
```
jplouis: Can the error checking and updating the deque be moved to the QueueGauge?

Products/ZenUtils/MetricReporter.py
```diff
@@ -30,6 +33,42 @@
...
+        self.gauge = gauge
+        self.tagValues = args
+    def __enter__(self):
+        self.interval = Interval.now()
+    def __exit__(self, *args):
+        self.gauge.queue.appendleft(self.tagValues + (self.interval.stop(),))
```
jplouis: From comment above, This requires knowledge of the internal data structures.  Something like self.gauge.update(tagValues, metricValue) would allow refactor ing of the queugauge and may be prone to less errors.

Products/ZenUtils/MetricReporter.py
sockterrier: Yes.

Products/ZenUtils/MetricReporter.py
sockterrier: See next commit.

PR #2646, opened 2017-08-23
ZEN-28051 Modeling devices fails when device name contains parentheses

romanbakaleyko:


romanbakaleyko:
backport https://github.com/zenoss/zenoss-prodbin/pull/2555

PR #2648, opened 2017-08-23
Fix for SOLR log duplicates

mayathegreat:
Added should_log parameter to avoid oversubscribing of loggers in CmdBase

### DIFF
--- a/Products/ZenUtils/CmdBase.py
+++ b/Products/ZenUtils/CmdBase.py
@@ -90,10 +90,7 @@ class CmdBase(object):
     """
     Class used for all Zenoss commands
     """
-
-    doesLogging = True
-
-    def __init__(self, noopts=0, args=None):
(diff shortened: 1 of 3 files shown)


Products/ZenUtils/CmdBase.py
```diff
@@ -90,10 +90,7 @@ class CmdBase(object):
     """
     Class used for all Zenoss commands
     """
-
-    doesLogging = True
-
-    def __init__(self, noopts=0, args=None):
+    def __init__(self, noopts=0, args=None, should_log=True):
```
jplouis: Want to preserve back compat if a subclass had set `doesLogging` to False.  You could set should_log to None and if set to something other than None it is used otherwise use old behavior.

mayathegreat: Good idea, done

PR #2650, opened 2017-08-24
ZEN-22112 All on same graph check issues flare

carlcamera:
Yellow flare issued after checking "all on same graph" checkbox on components graph panel when device contains no components.   Now checks for valid component ID before trying to update graph.

### DIFF
--- a/Products/ZenUI3/browser/resources/js/zenoss/form/ComponentGraphPanel.js
+++ b/Products/ZenUI3/browser/resources/js/zenoss/form/ComponentGraphPanel.js
@@ -214,21 +214,22 @@
         updateGraphs: function() {
             var meta_type = this.compType, uid = this.uid,
                 graphId = this.graphId, allOnSame = this.allOnSame.checked;
-            Zenoss.remote.DeviceRouter.getComponentGraphs({
-                uid: uid,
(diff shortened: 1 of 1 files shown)


Products/ZenUI3/browser/resources/js/zenoss/form/ComponentGraphPanel.js
```diff
@@ -214,21 +214,22 @@
...
-                if (response.success) {
-                    var graphs=[], fn;
-                    this.removeAll();
-                    fn = ZC.getComponentGraphRenderer(meta_type);
-                    graphs = fn(meta_type, uid, graphId, allOnSame, response.data);
-                    this.add(graphs);
-                }
-            }, this);
-
+            if (graphId != undefined) {
```
jpeacock-zenoss: I don't recall whether it's considered safer to use `!==` rather than `!=` ('safer' meaning fewer surprising default conversions).

carlcamera: I've adjusted the operator to specifically check for `undefined` which is what I am seeing when the device has no components modeled.

PR #2657, opened 2017-08-28
ZEN-27288 UI improvements to Event Class Mapping Form

karolzen:
https://jira.zenoss.com/browse/ZEN-27288


Products/ZenEvents/EventClass.py
```diff
@@ -240,11 +240,11 @@ def getInstances(self):
...
-        idx = 0
-        insts = self.find(key)
-        if len(insts) > 0:
-            idx = insts[-1].sequence + 1
-        return idx
+        instances = self.find(key)
+        if len(instances) > 0:
+            return max([i.sequence for i in instances])+1
+        else:
+            return 0
```
jpeacock-zenoss: The `else:` is redundant when there's no work to do when the `if` statement is false. Also, the list comprehension given to `max` should be a generator comprehension.  This block of code can be written more succinctly as
```
return max(i.sequence for i instances) + 1 if len(instances) else 0
```
or
```
if len(instances):
    return max(i.sequence for i in instances) + 1
return 0
```
Note: I used `len(instances)` as the test condition because `len(seq)` and `len(seq) > 0` are equivalent in boolean contexts (such as the condition of an `if` statement).

Products/ZenUI3/browser/resources/js/zenoss/form/Action.js
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * This content is made available according to terms specified in
+ * License.zenoss under the directory where your Zenoss product is installed.
+ *
+ ****************************************************************************/
+
+
+/* package level */
+(function() {
+    Ext.define("Action", {
```
jpeacock-zenoss: This should be defined in the `Zenoss` namespace.  E.g.
```
    Ext.ns("Zenoss");
    Ext.define("Zenoss.Action"), {
        ...
```

Products/ZenEvents/EventClass.py
```diff
@@ -67,52 +71,56 @@ class EventClass(EventClassPropertyMixin, Organizer, ManagedEntity, ZenPackable)
...
-                , 'action'        : 'eventClassStatus'
-                , 'permissions'   : (
-                  Permissions.view,)
+            'id': 'EventClass',
+            'meta_type': 'EventClass',
+            'description': """Base class for all event classes""",
+            'icon': 'EventClass.gif',
+            'product': 'ZenEvents',
+            'factory': 'manage_addEventClass',
+            'immediate_view': 'eventClassStatus',
```
jpeacock-zenoss: For readability purposes, having a stack of key/value pairs should probably be formatted into columns, e.g.
```
            'id':             'EventClass',
            'meta_type':      'EventClass',
            'description':    'Base class for all event classes',
            'icon':           'EventClass.gif',
            'product':        'ZenEvents',
            'factory':        'manage_addEventClass',
            'immediate_view': 'eventClassStatus',
```
and using triple quotes (`"""`) is just a syntax for writing multi-line strings.  No need to use them if you have only a single line of text.
Note: The colon (`:`) is still next to the key name.  Adding spaces between the colon and the value hasn't been flagged by the PEP8 checkers I've used.

Products/ZenEvents/EventClass.py
```diff
@@ -67,52 +71,56 @@ class EventClass(EventClassPropertyMixin, Organizer, ManagedEntity, ZenPackable)
...
+                    'id': 'config',
+                    'name': 'Configuration Properties',
+                    'action': 'zPropertyEditNew',
+                    'permissions': ("Change Device",)
                 },
             )
-         },
-        )
+        },
+    )
```
jpeacock-zenoss: I've found that the conventional Javascript object literal style is more compact and works in Python as well (and still appears to be PEP8 compliant), e.g.
```
    some_variable = ({
        "id":    "id01",
        "value": "value01"
    }, {
        "id":    "id02",
        "value": "value02"
    },)
```
instead of
```
    some_variable = (
        {
            "id":    "id01",
            "value": "value01"
        },
        {
            "id":    "id02",
            "value": "value02"
        },
    )
```
This is just a suggestion.  If it looks too weird, you can ignore it.

Products/ZenEvents/EventClass.py
```diff
@@ -369,18 +386,16 @@ def reIndex(self):
...
                 ip.index_object()
 
-
     security.declareProtected(ZEN_ADD, 'createCatalog')
     def createCatalog(self):
         """Create a catalog for EventClassRecord searching"""
         from Products.ZCatalog.ZCatalog import manage_addZCatalog
         manage_addZCatalog(self, self.default_catalog,
-                            self.default_catalog)
+                           self.default_catalog)
```
jpeacock-zenoss: I don't think this needs to be two lines.  It looks likes this can be all one line and still be under 80 characters, e.g.
```
       manage_addZCatalog(self, self.default_catalog, self.default_catalog)
```

Products/ZenEvents/EventClass.py
karolzen: Good point! :+1:

Products/ZenEvents/EventClass.py
karolzen: +

Products/ZenEvents/EventClass.py
karolzen: It looks fine for me :+1:

Products/ZenEvents/EventClass.py
karolzen: Indeed, fixed.

Products/ZenEvents/EventClass.py
```diff
@@ -236,15 +233,13 @@ def getInstances(self):
...
         """Get next sequence number for instance.
         """
-        idx = 0
-        insts = self.find(key)
-        if len(insts) > 0:
-            idx = insts[-1].sequence + 1
-        return idx
+        instances = self.find(key)
+        if len(instances):
+            return max([i.sequence for i in instances]) + 1
```
jpeacock-zenoss: generator comprehension would be better here.

PR #2663, opened 2017-08-29
Don't delete MIB OIDs on ZenPack install.

jpeacock-zenoss:
Also improved error handling when resolving MIB catalog brains into ZODB objects.

Fixes ZEN-28458.

### DIFF
--- a/Products/ZenModel/MibModule.py
+++ b/Products/ZenModel/MibModule.py
@@ -22,22 +22,26 @@ from ZenModelRM import ZenModelRM
 from ZenPackable import ZenPackable
 
 
-def createOID(dmd, container, new_oid, logger=None):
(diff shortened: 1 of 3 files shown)


Products/ZenModel/MibModule.py
```diff
@@ -22,22 +22,26 @@
...
-        if logger and old_oid.moduleName != new_oid.moduleName:
+def createOID(dmd, container, new_node, logger=None):
+    for brain in dmd.mibSearch(oid=new_node.oid):
+        old_node = dmd.unrestrictedTraverse(brain.getPath(), None)
+        if old_node is None:
+            # Brain's path not found, so delete it.
+            dmd.mibSearch.uncatalog_object(brain.getPath())
+            continue
+        if old_node.moduleName != new_node.moduleName:
             logger.warn(
```
fraibacas: logger could be None

PR #2667, opened 2017-08-30
ZEN-28356: new arguments to extend remodel API functionality

rderkach:
The API Call DeviceRouter.remodel is too limited.
It will run all modeler plugins assigned to a device
There are many cases where only a subset of all modeler
plugins will need to be run, frequently only 1 or 2.
Added two more argument to remodel method, which
will extend its API functionality.

[JIRA&DEMO](https://jira.zenoss.com/browse/ZEN-28356)

Related PR https://github.com/zenoss/ZenPacks.zenoss.DistributedCollector/pull/63


Products/Zuul/facades/devicefacade.py
```diff
@@ -711,11 +711,13 @@ def addDevice(self, deviceName, deviceClass, title=None, snmpCommunity="",
                                                title=title)
         return jobrecords
 
-    def remodel(self, deviceUid):
-        fake_request = {'CONTENT_TYPE': 'xml'}
+    def remodel(self, deviceUid, collectPlugins, background):
```
jplouis: The new options should be optional keyword args so as not to break backwards compatibility or customer scripts.

rderkach: Yes sure, good point.

rderkach:
@jplouis Applied requested changes.

PR #2671, opened 2017-08-30
Bump version to 5.3.2

tgjonestx:
The tip of support/5.3.x should be versioned as 5.3.2 now that we're in the process of releasing a hotfix for ZEN-28477 as 5.3.1


tgjonestx:
Retest this

PR #2686, opened 2017-09-04
ZEN-13992: SNMP Datasource Test function does not escape special char…

olst:
…acters in community string


olst:
There is a new commit. I tested this community string: `z$&|+-!(){}[]^~*?:noss`.
I checked data source testing, perf collection and modeling. Please see demo:
https://youtu.be/H2CMuXISFL0

olst:
Data source testing doesn't work without double $.

olst:
Data source testing doesn't work without double $.
Please see this video: https://youtu.be/0kxf96Ng4LU

PR #2695, opened 2017-09-07
added an audit log message

sockterrier:


sockterrier:
These are example lines logged from the application router now (note that this facility saves all config files on save, not just changed config files):

2017-09-07 18:23:16 INFO user=admin action=UpdateConfigFiles kind=Applications filename=/opt/zenoss/etc/central-query/central-query_supervisor.conf service=amai47zvkn0493d26jdqc35v4 servicename=CentralQuery
2017-09-07 18:23:16 INFO user=admin action=UpdateConfigFiles kind=Applications filename=/opt/zenoss/etc/central-query/configuration.yaml service=amai47zvkn0493d26jdqc35v4 servicename=CentralQuery

PR #2698, opened 2017-09-08
ZEN-28378 Multi Add Device requires a password

romanbakaleyko:


romanbakaleyko:
After digging in the code base I see that the restriction was added in scope of this ticket https://jira.zenoss.com/browse/ZEN-14908. I've read this ticket and not sure why did we do that, no clue how proposed fix fixes the issue described in ZEN-14908. Another thing is that username/password/community are not mandatory in the single adding device form so I decided to remove that restriction. I've checked and confirm that creating windows/ssh linux devices works fine even with empty username/password fields (modeling fails but that is expected).

PR #2702, opened 2017-09-11
ZEN-28547 Muliti Add Device requires a password

romanbakaleyko:


romanbakaleyko:
port https://github.com/zenoss/zenoss-prodbin/pull/2698

PR #2730, opened 2017-09-20
ZEN-28478: adding a migration script for opentsdb OriginalConfig

rderkach:
This is a migration script for an old ticket https://jira.zenoss.com/browse/ZEN-19520

Back than there was no service migration SDK, so service migrations was manually "hacked". There was a custom script to migrate, but it only changed  ConfigFiles in service definition. 

This is needed for the systems that were originally upgraded from 5.0.5 and lower

[JIRA](https://jira.zenoss.com/browse/ZEN-28478)


ssOlexBaiko:
lgtm

PR #2733, opened 2017-09-22
ZEN-28539: Openstack - Traceback modeling

olst:


Products/ZenUtils/ShardedBTree.py
```diff
@@ -12,6 +12,8 @@
 from itertools import chain
 from zlib import adler32
 
+import logging
+log = logging.getLogger('zen.GuidManager')
```
fraibacas: Please name the logger something else. Currently we only use the sharded btree for the guid table but in the future we can use it to store any other data.

Products/ZenUtils/ShardedBTree.py
```diff
@@ -74,7 +87,8 @@ def __len__(self):
 
     def __delitem__(self, key):
         shard = self._get_shard(key)
-        del shard[key]
+        if shard is not None:
+            del shard[key]
 
     def __contains__(self, key):
         return self.has_key(key)
```
fraibacas: has_key needs to be updated as well

Products/ZenUtils/ShardedBTree.py
```diff
@@ -84,15 +98,18 @@ def __iter__(self):
...
 
     def clear(self):
         for shard in self.shards:
             shard.clear()
 
     def has_key(self, key):
         shard = self._get_shard(key)
-        return shard.has_key(key) > 0
+        if shard is not None:
+            return shard.has_key(key) > 0
```
fraibacas: if shard is None return False

PR #2734, opened 2017-09-22
ZEN-28475 Wrong time format in exported event files .xls, .csv

ISPHOST:


ISPHOST:
Added dialog with export options.

PR #2748, opened 2017-09-28
ZEN-28465 Migrate script: rateOptions for counters

karolzen:
https://jira.zenoss.com/browse/ZEN-28465


karolzen:
"ResetValue" is specified only for zenpython and only in develop. After changes to Zenoss' query we don't need to specify that (if reset values and counter max are defaults "dropReset" option will be send to DB).

PR #2766, opened 2017-10-04
Add the installation script for nagios plugins

youngklee:


bin/nagios-plugins.sh
```diff
@@ -0,0 +1,97 @@
+#! /usr/bin/env bash
+##############################################################################
+#
+# Copyright (C) Zenoss, Inc. 2015, all rights reserved.
```
ed-: 2017

bin/nagios-plugins.sh
```diff
@@ -0,0 +1,97 @@
...
+        return $RC
+    fi
+
+    DEST=/root/rpm/nagios
+    if [[ ! -z "$1" ]]; then
+        DEST=$1
+        shift
+    fi
+    pushd $DEST
+    yum localinstall -y $COMMON_NAME-$COMMON_VER-$PKG_SUFFIX
```
ed-: catch the RC of this line like L78-82

bin/nagios-plugins.sh
```diff
@@ -0,0 +1,97 @@
...
+    done
+    popd
+    return 0
+}
+
+if [[ "$1" == "install" ]]; then
+    shift
+    install $@
+elif [[ "$1" == "download" ]]; then
+	shift
```
ed-: this is a nitpick but the indentation is inconsistent here

bin/nagios-plugins.sh
```diff
@@ -0,0 +1,97 @@
...
+  	download $@
+else
+    help
+fi
+return $?
```
ed-: will this return the error codes caught as RC, or will it be 0 as long as `fi` succeeds?
besides that you can't return unless you're in a function, this should be exit

bin/nagios-plugins.sh
youngklee: Revised as suggested.

bin/nagios-plugins.sh
youngklee: Revised as suggested.

bin/nagios-plugins.sh
youngklee: Revised as suggested.

bin/nagios-plugins.sh
youngklee: Fixed as suggested.

youngklee:
The management has decided to include the nagios plugins in the image for Thebe. This script is no longer needed.

PR #2767, opened 2017-10-04
avoiding the IndexableWrapper since some of them call notify

sockterrier:


Products/Zuul/catalog/model_catalog.py
```diff
@@ -119,8 +119,6 @@ def getObject(self):
             msg = "Unable to get object from brain. Path: {0}. Model catalog may be out of sync. "
             msg += "Will attempt to delete the object from model catalog."
```
fraibacas: We should remove the logging that states the object is going to be unindexed

Products/Zuul/catalog/tests/testModelCatalog.py
```diff
@@ -44,14 +44,16 @@ def test_stale_brain(self):
...
+        # an object, but it is not done by ModelCatalogBrain.getObject() :
+        exception_raised = False
         try:
             obj = brain.getObject()
         except:
-            expception_raised = True
-        self.assertTrue(expception_raised)
+            exception_raised = True
+        self.assertTrue(exception_raised)
         results = self.model_catalog.search(query={UID:bad_uid})
```
fraibacas: That search and assert dont make sense anymore. Lines 42 and 43 do it.

PR #2770, opened 2017-10-09
ZEN-28747 After upgrade from 5.3.2 to Thebe can't open device details.

romanbakaleyko:


romanbakaleyko:
Closed as in my environment manual run of this script ends up with a traceback, need to debug to make sure this is not a problem with this migration script.

PR #2773, opened 2017-10-10
ZEN-28720: AttributeError when attempting to remove a Realm

olst:


Products/Zuul/facades/networkfacade.py
```diff
@@ -79,6 +81,7 @@ def findSubnet(self, netip, netmask, contextUid):
     def deleteSubnet(self, uid):
         toDel = self._dmd.restrictedTraverse(uid)
         aq_parent(toDel)._delObject(toDel.id)
+        mc_utils.unindex_by_uid(uid)
```
fraibacas: That is not the proper way to unindex within a transaction. ModelCatalogUtils is meant to be used as a tool. To unindex an object within a transaction you need to raise an IndexingEvent.

fraibacas: Actually, you should not need to create an IndexingEvent, delObject should do it for you.

PR #2776, opened 2017-10-11
ZEN-7184: sort templates in a right order

rderkach:
When you reorder templates in a 'Bind Templates' window
an order of selected wasn't shown after you reopen a window.
Sort templates according to zDeviceTemplates property.

[DEMO&JIRA](https://jira.zenoss.com/browse/ZEN-7184)

### DIFF
--- a/Products/Zuul/facades/devicefacade.py
+++ b/Products/Zuul/facades/devicefacade.py
@@ -770,8 +770,11 @@ class DeviceFacade(TreeFacade):
 
     def _getBoundTemplates(self, uid, isBound):
         obj = self._getObject(uid)
-        templates = (template for template in obj.getAvailableTemplates()
-             if (template.id in obj.zDeviceTemplates) == isBound)
+        templates = (
+            template
+            for template in obj.getAvailableTemplates()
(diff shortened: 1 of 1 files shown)


Products/Zuul/facades/devicefacade.py
```diff
@@ -770,9 +770,13 @@ def getBoundTemplates(self, uid):
 
     def _getBoundTemplates(self, uid, isBound):
         obj = self._getObject(uid)
-        for template in obj.getAvailableTemplates():
-            if (template.id in obj.zDeviceTemplates) == isBound:
-                yield template
+        templates = (template for template in obj.getAvailableTemplates()
+             if (template.id in obj.zDeviceTemplates) == isBound)
```
jpeacock-zenoss: May I suggest a possibly easier-to-read formatting of this statement?
```
templates = (
    template
    for template in obj.getAvailableTemplates()
    if (template.id in obj.zDeviceTemplates) == isBound
)
```
Comprehensions like this have a lot going on and I've found the above formatting to have a lower cognitive load when reading it.

rderkach: Yep, why not.

PR #2779, opened 2017-10-11
ZEN-28777: Fix catalog search scope findDeviceByIdExact in ZenModel.D…

eimajenthat:
…eviceClass


Products/ZenModel/DeviceClass.py
```diff
@@ -559,7 +559,7 @@ def getSubComponents(self, meta_type="", monitored=True):
         """
         Return generator of components, by meta_type if specified
         """
-        catalog = IModelCatalogTool(self)
+        catalog = IModelCatalogTool(self.dmd.Devices)
```
fraibacas: This needs to be self bc results need to depend on the current context. See https://github.com/zenoss/zenoss-prodbin/blob/support/5.3.x/Products/ZenModel/DeviceClass.py#L571

eimajenthat: Oops!  I see what you mean now.  Fixed.

PR #2783, opened 2017-10-12
ZEN-28589: Zencommand not functioning from the UI

olst:


Products/ZenModel/RRDDataSource.py
```diff
@@ -179,7 +179,7 @@ def getCommand(self, context, cmd=None, device=None):
         """
         # Perform a TALES eval on the expression using self
         if cmd is None:
-            cmd = self.commandTemplate
+            cmd = self.commandTemplate.replace("$", "$$")
```
jplouis: getCommand is called by CommandPerformanceConfig which is what is used for collection. Seems like this would affect collection.  Ticket states that collection works fine.

PR #2793, opened 2017-10-16
Add support for indicating when model change events are caused by a

ecifreo:
Add support for indicating when model change events are caused by a maintenance window.

### DIFF
--- a/Products/ZenMessaging/queuemessaging/publisher.py
+++ b/Products/ZenMessaging/queuemessaging/publisher.py
@@ -48,6 +48,7 @@ class ModelChangePublisher(object):
         self._publishable = []
         self._discarded = 0
         self._total = 0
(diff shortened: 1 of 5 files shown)


Products/ZenMessaging/queuemessaging/publisher.py
```diff
@@ -271,7 +273,7 @@ def beforeCompletionHook(self, tx):
             if publisher:
                 msgs = self.correlateEvents(publisher.events)
                 with _getPrepublishingTimer():
-                    notify(MessagePrePublishingEvent(msgs))
+                    notify(MessagePrePublishingEvent(msgs, maintWindowChange=publisher._maintWindowChange))
```
jplouis: This presumes all events are maintenance window changes if any one was caused by a production state change.

PR #2801, opened 2017-10-18
ZEN-28735: Zencommand is depending on an IP address for devices Unkno…

olst:


Products/ZenRRD/zencommand.py
```diff
@@ -384,6 +384,14 @@ def __init__(self,
         self.executed = 0
         self._lastErrorMsg = ''
 
+        self.manage_ip_event = {
+            'eventClass': Cmd_Fail,
+            'component': 'command',
+            'device': self._devId,
+            'summary': 'Unknown host',
+            'component' : COLLECTOR_NAME,
```
jplouis: The summary is vague, it should be more specific. Maybe something like "IP address not set, collection will be attempted with host name"

olst:
In ZEN-26521 there are no exceptions (only `Connection refused` events) but we need "IP Address not set..." event. Looks like we can't reduce the number of "IP Address not set..." events.

Products/ZenRRD/zencommand.py
```diff
@@ -405,7 +414,9 @@ def doTask(self):
         self.state = SshPerformanceCollectionTask.STATE_CONNECTING
         try:
             if not self._manageIp:
-                raise Exception("Unknown host")                
+                self._eventService.sendEvent(self.manage_ip_event, severity=Event.Warning)
```
joshw: In the non-ssh case, not having a manageIp is not necessarily any kind of problem.. The Openstack and OpenstackInfrastructure zenpacks are examples of this- they execute a command locally which connects to a URL specified in a zproperty, and have no mangageIp.

olst: you mean we should send events if this is True?
`if not self._manageIp and self._useSsh:`

joshw: That would be fine for me.  I think otherwise you'd need to have a way to determine whether the manageIp is used.  (either manually specify it, or determine if it's passed to the command as an argument?)

olst: https://github.com/zenoss/zenoss-prodbin/pull/2812

jplouis:
See PR https://github.com/zenoss/zenoss-prodbin/pull/2812

PR #2802, opened 2017-10-19
Zencommand is depending on an IP address for devices Unknown host

olst:


olst:
Some customers have fake devices without IP addreses, they use datasources without `Use SSH`.

PR #2803, opened 2017-10-19
ZEN-28780 Graph Improvements

carlcamera:
Improved padding for footers. Flare when trying to create a stacked line chart.

### DIFF
--- a/Products/ZenModel/GraphPoint.py
+++ b/Products/ZenModel/GraphPoint.py
@@ -1,10 +1,10 @@
 ##############################################################################
-# 
+#
 # Copyright (C) Zenoss, Inc. 2007, all rights reserved.
-# 
+#
 # This content is made available according to terms specified in
 # License.zenoss under the directory where your Zenoss product is installed.
-# 
+#
 ##############################################################################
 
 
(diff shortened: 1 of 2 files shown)


Products/ZenModel/GraphPoint.py
```diff
@@ -100,6 +100,17 @@ def manage_editProperties(self, REQUEST):
...
+        lineType = REQUEST.get('lineType', '').strip().lower()
+        stacked = str(REQUEST.get('stacked', '')).strip().lower()
+        stackedLine = lineType == 'line' and stacked == 'true'
+        if stackedLine:
+            messaging.IMessageSender(self).sendToBrowser(
+                'Invalid definition',
+                'Only Area graphs may be stacked. ',
+                priority=messaging.WARNING
+            )
+
```
rphillipsz: Does this mean if a graph is already defined as line + stacked, that they're going to get a flare warning every time a page is shown with that graph on it?

carlcamera: This is only checked after the [Save] button is pressed.  Existing definitions will continue to show as non-stacked line charts -- the stacked==true is ignored when line is defined.

PR #2812, opened 2017-10-25
ZEN-28735: Zencommand is depending on an IP address for devices Unkno…

olst:
…wn host

### DIFF
--- a/Products/ZenRRD/zencommand.py
+++ b/Products/ZenRRD/zencommand.py
@@ -384,6 +384,15 @@ class SshPerformanceCollectionTask(BaseTask):
         self.executed = 0
         self._lastErrorMsg = ''
 
+        self.manage_ip_event = {
+            'eventClass': Cmd_Fail,
+            'component': 'command',
+            'device': self._devId,
+            'summary': 'IP address not set, collection will be attempted\
+                        with host name',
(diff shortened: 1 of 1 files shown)


Products/ZenRRD/zencommand.py
```diff
@@ -404,12 +413,14 @@ def doTask(self):
         self._doTask_start = datetime.now()
         self.state = SshPerformanceCollectionTask.STATE_CONNECTING
         try:
-            if not self._manageIp:
-                raise Exception("Unknown host")                
-
-            yield self._connector.connect(self)
+            if not self._manageIp and self._useSsh:
+                self._eventService.sendEvent(self.manage_ip_event, severity=Event.Warning)
```
jplouis: Drop this to info.

Products/ZenRRD/zencommand.py
```diff
@@ -404,12 +413,14 @@ def doTask(self):
...
+            else:
+                self._eventService.sendEvent(self.manage_ip_event, severity=Clear)
 
             if self._useSsh:
+                # we need connection only when we use ssh.
+                yield self._connector.connect(self)
```
jplouis: connect() will be a no-op when not using SSH but adds debug logging. Put it back where it was.

PR #2821, opened 2017-10-31
ZEN-28797: mask password type properties.

rderkach:
On 'Overridden Objects' page when you select password
type property from a drop-down list its value will
be shown as a plain text.
Mask the password type properties with asterisk.

[JIRA&DEMO](https://jira.zenoss.com/browse/ZEN-28797)


rderkach:
@romanbakaleyko  Done

PR #2833, opened 2017-11-04
Check if the queue exists before creating it

rphillipsz:
https://jira.zenoss.com/browse/ZEN-28918

If the arguments used to create the queue are different from the arguments zenoss-protocols is trying to create it as, the create will fail and we can't connect to the queue.  If the queue exists, use the queue as-is without trying to create it.


webhookbotuser:
Build finished.

PR #2835, opened 2017-11-06
Check if the queue exists before creating it

rphillipsz:
https://jira.zenoss.com/browse/ZEN-28918

The port from develop to check if a rabbit queue exists before trying to create it.


webhookbotuser:
Build finished.

PR #2837, opened 2017-11-07
ZEN-28116: filter components by their production state for config

rderkach:
MW wasn't working for Component Group.
Some of the objects n the component group may not
have production state attribute, skip them.
When creating device config for cllector daemon
filter components by their production state in the
same way as devices are.

### DIFF
--- a/Products/ZenModel/MaintenanceWindow.py
+++ b/Products/ZenModel/MaintenanceWindow.py
@@ -555,6 +555,11 @@ class MaintenanceWindow(ZenModelRM):
 
         def _setProdState(devices_batch):
             for device in devices_batch:
+                # In case we try to move Component Group into MW
+                # some of objects in CG may not have production state
+                # we skip them.
+                if not hasattr(device, "privateattr_productionState"):
+                    continue
                 if ending:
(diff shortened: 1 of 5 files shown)


webhookbotuser:
Build finished.

Products/ZenModel/MaintenanceWindow.py
```diff
@@ -555,6 +555,11 @@ def setProdState(self, state, ending=False, batchSize=None,
 
         def _setProdState(devices_batch):
             for device in devices_batch:
+                # In case we try to move Component Group into MW
+                # some of objects in CG may not have production state
+                # we skip them.
+                if not hasattr(device, "privateattr_productionState"):
```
johnlhamilton: We should not reference privateattr_productionState directly.  Also, what are the cases where this is true?  Most components will NOT have a production state, but will acquire their device's production state.

webhookbotuser:
Build finished.

webhookbotuser:
Build finished.

webhookbotuser:
Build finished.

webhookbotuser:
Build finished.

PR #2852, opened 2017-11-10
Add a script to optionally force reindex during upgrades

rphillipsz:
https://jira.zenoss.com/browse/ZEN-29061
Called from the upgrade with the top-level service name as the argument.  Checks if the major version is prior to the addition of Solr to conditionally force a catalog reindex.


webhookbotuser:
Build finished.

webhookbotuser:
Build finished.

webhookbotuser:
Build finished.

PR #2854, opened 2017-11-12
Use /var/zenoss/upgrade_from_version.txt

rphillipsz:
https://jira.zenoss.com/browse/ZEN-29075
ZVersion.py has the to-version, not the from-version.  The current_version.txt upgrade step places /var/zenoss/upgrade_from_version.txt into the DFS, so we can use this to find out what version we're upgrading from.


webhookbotuser:
Build finished.

PR #2871, opened 2017-11-21
Zen 28615

ISPHOST:


Products/ZenCallHome/CallHomeStatus.py
```diff
@@ -0,0 +1,123 @@
...
+        except Exception as e:
+            log.warning("Exception trying to push metric to redis: {0}".format(e))
+            self._redis_client = None
+            return
+
+    def load_from_redis(self):
+        # Is redis up?
+        if not self._connected_to_redis():
+            return
+        data = {}
```
sockterrier: looks unused

Products/ZenCallHome/CallHomeStatus.py
```diff
@@ -0,0 +1,123 @@
...
+                "Exception trying to recive data from redis: {0}".format(e))
+            self._redis_client = None
+            return
+
+    def init(self):
+        """Sets empty data for CallHomeStatus before run
+        """
+        data = dict()
+        stages = ('Request to CallHome server', 'CallHome start',
+                  'Update report', 'CallHome Collect', 'GatherProtocol')
```
sockterrier: Move these stages out into something visible on the module or class level. Then in cycler, directpost, or callhome below, do something like 'chs.stage(CALLHOME_COLLECT)' where you currently have 'chs.stage("CallHome Collect")'

Products/ZenCallHome/CallHomeStatus.py
```diff
@@ -0,0 +1,123 @@
...
+                'description': v,
+                'status': '2',
+                'error': '',
+                'stime': '-1'
+            }
+        data = pickle.dumps(data)
+        self.push_to_redis(data)
+
+
+    def stage(self, stage, status="RUNNING", err=""):
```
sockterrier: reject stages that don't match one of the anticipated ones with an exception.

PR #2874, opened 2017-11-23
ZEN-29155: Can't create a maintenance window correctly using JSON api.

olst:


Products/Zuul/infos/devicemanagement.py
```diff
@@ -38,8 +40,26 @@ def startTime(self):
...
+        duration = int(p['durationMinutes']) or int(p['durationHours']) \
+            or int(p['durationDays'])
+        if not duration:
+            p['durationHours'] = 1
+        if p['repeat'] not in self._object.REPEAT:
+            p['repeat'] = self._object.REPEAT[0]
         self._object.manage_editMaintenanceWindow( 
-                                     startDateTime=p['startDateTime'],
+                                     startDateTime=startDateTime,
                                      durationDays=p['durationDays'],
```
jplouis: Seems like this method is duplicating some default value picking logic defined in the manage_editMaintenanceWindow param and method.  Specifically the logic around startDateTime and startHours etc.... Those values should be passed in only if they are present in the the `p` param.

Why is it default to 1 hour duration if the duration not set?

Seems like an exception should be thrown if an invalid `repeat` param is sent.

olst:
updated

olst:
Please see this videos:
issue: https://youtu.be/-_YnjSHGW58
fix: https://youtu.be/ZDS0QDFA7mg

Products/Zuul/infos/devicemanagement.py
```diff
@@ -39,7 +39,10 @@ def startTime(self):
 
     def updateWindow(self, p):
         self._object.manage_editMaintenanceWindow( 
-                                     startDateTime=p['startDateTime'],
+                                     startDateTime=p.get('startDateTime', None),
+                                     startDate=p.get('startDate', None),
+                                     startHours=p.get('startHours', None),
+                                     startMinutes=p.get('startMinutes', None),
```
jplouis: If startDateTime, startMinutes and startHours are not in the `p` map and get set to None, the manage_editMaintenanceWindow method will throw an exception.

PR #2889, opened 2017-12-11
ZEN-28772 Graph height 100

carlcamera:
Catch user-input of 100 chart height and adjust width to prevent backwards-compat code from adjusting height to 500.  Hide width row because width values are ignored.


carlcamera:
Whoops I forgot a semicolon. Hold on.

PR #2902, opened 2017-12-13
ZEN-29115: Json api DeviceManagementRouter addMaintWindow() method do…

olst:
https://jira.zenoss.com/browse/ZEN-29115
Fixes:
- invalid value for enabled throws exception instead of clean error response 
- able to add an invalid prod state
- api returned success when attempt to add MW without a name value (it did not add, but api returned success)


olst:
fixed

PR #2908, opened 2017-12-14
Fixed issue when pickle tried loads None value

ISPHOST:
https://jira.zenoss.com/browse/ZEN-28615

### DIFF
--- a/Products/ZenCallHome/CallHomeStatus.py
+++ b/Products/ZenCallHome/CallHomeStatus.py
@@ -110,9 +110,11 @@ class CallHomeStatus(object):
     def _init(self):
         """Sets empty data for CallHomeStatus before run
         """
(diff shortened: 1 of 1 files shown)


Products/ZenCallHome/CallHomeStatus.py
```diff
@@ -110,9 +110,11 @@ def getStatUI(self):
     def _init(self):
         """Sets empty data for CallHomeStatus before run
         """
-        data = pickle.loads(self.load_from_redis())
-        if data == None:
+        rdata = self.load_from_redis()
+        if rdata == None:
```
jpeacock-zenoss: "Correct" Python requires this to be written as `if rdata is None:`.  Since `None` is a singleton value, you should use identity, i.e. `is`, rather than equality for comparison.

PR #2910, opened 2017-12-14
ZEN-28615 Update Report button on Licensing tab does not update

ISPHOST:
licensing details with silent failure
Fixed issue when pickle tried loads None value


Products/ZenCallHome/CallHomeStatus.py
```diff
@@ -93,8 +93,11 @@ def getStatUI(self):
         """Returns status informations to UI
         """
         l = list()
-        data = dict()
-        data = pickle.loads(self.load_from_redis())
+        rdata = self.load_from_redis()
+        if rdata is None:
+            data = dict()
+        else:
+            data = pickle.loads(rdata)
```
jpeacock-zenoss: These lines of code are duplicated in at least two places.  Perhaps this logic should be moved into the `load_from_redis` method so that we don't have to worry about whether `None` is returned in multiple places.

Products/ZenCallHome/CallHomeStatus.py
```diff
@@ -76,10 +76,9 @@ def load_from_redis(self, k="CallHomeStatus"):
             log.warning(
                 "Exception trying to recive data from redis: {0}".format(e))
             self._redis_client = None
-            return
+            return dict()
```
jpeacock-zenoss: It looks like the `load_from_redis` method can now return _three_ different types of values: `None`, a `dict` object, and a string (or buffer) object.  A method or function should only be _one_ type or, at worst, two types as long as the second value type is `None`.

jpeacock-zenoss: I haven't examined all of the code, but I suspect that usages of `push_to_redis` and `load_from_redis` can be simplified if those methods handled object pickling themselves.  I.e. Move all the calls to `pickle.dumps` into `push_to_redis` and the calls to `pickle.loads` into `load_from_redis`.

ISPHOST:
RETEST

Products/ZenCallHome/CallHomeStatus.py
```diff
@@ -76,45 +78,45 @@ def load_from_redis(self, k="CallHomeStatus"):
...
                 else:
                     err = "No errors"
         l.append({'id': 'updresult', 'description': 'Updating result', 'value': err, 'type': 'text'})
         return l
 
+    def _pickleLoadRedis(self):
+        data = dict()
+        if self.load_from_redis() is not None:
+            data = pickle.loads(self.load_from_redis())
+        return data
```
jpeacock-zenoss: Please don't call `load_from_redis` twice.  Call it once and then check the result in a separate statement.

Products/ZenCallHome/CallHomeStatus.py
```diff
@@ -76,45 +78,45 @@ def load_from_redis(self, k="CallHomeStatus"):
             log.warning(
                 "Exception trying to recive data from redis: {0}".format(e))
             self._redis_client = None
-            return
+            return None
```
jpeacock-zenoss: `return None`, while explicit, is redundant.  A bare `return` returns `None` by default.

PR #2915, opened 2017-12-19
ZEN-18523 Device tree collapses on refresh

carlcamera:
Maintains visual state of device tree in side panel across refreshes.


carlcamera:
retest this

PR #2916, opened 2017-12-20
ZEN-29178: set buffer attribute of store to false

rderkach:
If you move  a sequence located at the top to the very bottom
it moves back to its starting location after moving it.
This was happening since data was buffered and then restored
to the original list.

[JIRA&DEMO](https://jira.zenoss.com/browse/ZEN-29178)


ssOlexBaiko:
lgtm

PR #2923, opened 2018-01-03
ZEN-29127: remove response result unmarshal, move log level to debug

rderkach:
If we try to unmarshal response result which is not a json valid
string we will get a traceback, which will then be logged to the user.
There is no reason to unmarshal result since it will be then
logged as a string again.

[JIRA](https://jira.zenoss.com/browse/ZEN-29127)


Products/ZenHub/metricpublisher/publisher.py
```diff
@@ -126,7 +126,7 @@ def put(self, metric, value, timestamp, tags):
 
     def _putLater(self, scheduled):
         def handleError(val):
-            log.warn("Error sending metric: %s", val)
+            log.debug("Error sending metric: %s", val)
```
rderkach: I think we should not print Tracebacks at **_warn_** log level

PR #2928, opened 2018-01-07
ZEN-29263 Administered Objects recurse Devices but not Events from to…

romanbakaleyko:
…p level organizer

### DIFF
--- a/Products/ZenModel/UserSettings.py
+++ b/Products/ZenModel/UserSettings.py
@@ -1172,7 +1172,7 @@ class UserSettings(ZenModelRM):
                 if ar.managedObject().getPrimaryId() in rootOrganizers:
                     guids.extend(IGlobalIdentifier(child).getGUID() for child in ar.managedObject().children())
                 else:
-                    guids.extend(IGlobalIdentifier(ar.managedObject()).getGUID())
+                    guids.append(IGlobalIdentifier(ar.managedObject()).getGUID())
 
         return guids
 
--- a/Products/ZenModel/UserSettings.py
+++ b/Products/ZenModel/UserSettings.py
@@ -1151,26 +1151,28 @@ class UserSettings(ZenModelRM):
                 )
             return self.callZenScreen(REQUEST)
 
+
     security.declareProtected(ZEN_CHANGE_SETTINGS, 'getAllAdminGuids')
     def getAllAdminGuids(self, returnChildrenForRootObj=False):
         """
         Return all guids user has permissions to.
         """
         guids = []
-        ars = self.getAllAdminRoles()
-        for ar in ars:
-            if not returnChildrenForRootObj:
-                guids.append(IGlobalIdentifier(ar.managedObject()).getGUID())
-            else:
-                if ar.managedObject().getPrimaryId() in (
+        rootOrganizers = (
                         '/zport/dmd/Devices',
                         '/zport/dmd/Locations',
                         '/zport/dmd/Groups',
                         '/zport/dmd/Systems'
-                ):
-                    guids.extend([IGlobalIdentifier(child).getGUID() for child in ar.managedObject().children()])
+        )
+        ars = self.getAllAdminRoles()
+        if not returnChildrenForRootObj:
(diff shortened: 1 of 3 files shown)


Products/ZenModel/UserSettings.py
```diff
@@ -1151,6 +1151,29 @@ def manage_deleteAdministrativeRole(self, delids=(), REQUEST=None):
...
+    def getAllAdminGuids(self, returnChildrenForRootObj=False):
+        """
+        Return all guids user has permissions to.
+        """
+        guids = []
+        ars = self.getAllAdminRoles()
+        for ar in ars:
+            if not returnChildrenForRootObj:
+                guids.append(IGlobalIdentifier(ar.managedObject()).getGUID())
+            else:
```
jpeacock-zenoss: The `for` loop and this `if` statement regarding the `returnChildrenForRootObj` should be inverted.  The result of the `if` is the same regardless of the value of `ar`.  Granted, there will be two loops, but they will be less complicated.  E.g.
```
if not returnChildrenForRootObj:
    guids.extend(
        IGlobalIdentifier(ar.managedObject()).getGUID()
        for ar in ars
    )
else:
    for ar in ars:
        if ar.managedObject().getPrimaryId() in ....
```

Products/ZenModel/UserSettings.py
```diff
@@ -1151,6 +1151,29 @@ def manage_deleteAdministrativeRole(self, delids=(), REQUEST=None):
...
+                if ar.managedObject().getPrimaryId() in (
+                        '/zport/dmd/Devices',
+                        '/zport/dmd/Locations',
+                        '/zport/dmd/Groups',
+                        '/zport/dmd/Systems'
+                ):
```
jpeacock-zenoss: The tuple of organizer names should be kept in a variable with an informative name, e.g.
```
    rootOrganizers = (
        '/zport/dmd/Devices',
        ...
    )
    if ar.managedObject().getPrimaryId() in rootOrganizers:
        ...
```

Products/ZenModel/UserSettings.py
```diff
@@ -1151,6 +1151,29 @@ def manage_deleteAdministrativeRole(self, delids=(), REQUEST=None):
...
+                    guids.extend([IGlobalIdentifier(child).getGUID() for child in ar.managedObject().children()])
```
jpeacock-zenoss: You can use a generator here, i.e. you don't need the square brackets.

jpeacock-zenoss: Does `children()` return all the child objects in the tree?  If there are three levels of organizers, will objects of the 3rd nested be returned?

Products/Zuul/facades/zepfacade.py
```diff
@@ -369,17 +369,16 @@ def getEventSummaries(self, offset, limit=1000, sort=None, filter=None, exclusio
         if use_permissions:
             user = getSecurityManager().getUser()
             userSettings = self._dmd.ZenUsers._getOb(user.getId())
-            hasGlobalRoles = not userSettings.hasNoGlobalRoles()
+            hasGlobalRoles = not userSettings.hasNoGlobalRoles()
```
jpeacock-zenoss: There's a trailing space on this line that should be removed.

Products/ZenModel/UserSettings.py
```diff
@@ -1152,6 +1152,31 @@ def manage_deleteAdministrativeRole(self, delids=(), REQUEST=None):
...
+        )
+        ars = self.getAllAdminRoles()
+        if not returnChildrenForRootObj:
+            guids.extend(IGlobalIdentifier(ar.managedObject()).getGUID() for ar in ars)
+        else:
+            for ar in ars:
+                if ar.managedObject().getPrimaryId() in rootOrganizers:
+                    guids.extend(IGlobalIdentifier(child).getGUID() for child in ar.managedObject().children())
+                else:
+                    guids.extend(IGlobalIdentifier(ar.managedObject()).getGUID())
```
jpeacock-zenoss: This one should be `append`, not `extend`; only a single element is added to the list.

PR #2929, opened 2018-01-08
ZEN-29333 Allow to update distribution if other version already exists

karolzen:
https://jira.zenoss.com/browse/ZEN-29333
Upgrade of ZPs will fail if custom ZP installed on pre-upgrade system requires higher version of some dependency than to-upgrade package contains.
https://github.com/pypa/setuptools/blob/master/pkg_resources/__init__.py#L776


ssOlexBaiko:
lgtm

PR #2937, opened 2018-01-16
ZEN-29335 During RM Upgrade to 6.0.1 and 6.1 message is displayed tel…

romanbakaleyko:
…ling user to rebuild their catalog, (they don't actually have to, this is misleading)

### DIFF
--- a/bin/remove_CatalogServiceZenPack.sh
+++ b/bin/remove_CatalogServiceZenPack.sh
@@ -9,5 +9,14 @@ fi
 
 echo "Found Catalog Service egg at $CATALOG_SERVICE_EGG"
 /opt/zenoss/bin/zenpack --files-only --install $CATALOG_SERVICE_EGG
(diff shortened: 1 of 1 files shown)


bin/remove_CatalogServiceZenPack.sh
```diff
@@ -9,5 +9,14 @@ fi
 
 echo "Found Catalog Service egg at $CATALOG_SERVICE_EGG"
 /opt/zenoss/bin/zenpack --files-only --install $CATALOG_SERVICE_EGG
-/opt/zenoss/bin/zenpack --remove ZenPacks.zenoss.CatalogService
+#show log only if something goes wrong
+tmp_output="/tmp/"`cat /dev/urandom | tr -cd 'a-f0-9' | head -c 32`
+/opt/zenoss/bin/zenpack --remove ZenPacks.zenoss.CatalogService  > tmp_output 2>&1
```
jpeacock-zenoss: I think the redirection should be `$tmp_output`, right?

PR #2957, opened 2018-01-25
ZEN-29492: RM upgrade fails on dependencies for zenpacks that don't s…

olst:
https://jira.zenoss.com/browse/ZEN-29492

### DIFF
--- a/Products/ZenUtils/zenpack.py
+++ b/Products/ZenUtils/zenpack.py
@@ -445,7 +445,7 @@ class ZenPackCmd(ZenScriptBase):
         # Respect the source of the pack
         zpsToSort = {}
 
-        pattern = '(ZenPacks\.zenoss\.[a-zA-Z\.]*)'
+        pattern = '(ZenPacks\.[a-zA-Z\.]+)'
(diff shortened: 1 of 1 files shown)


Products/ZenUtils/zenpack.py
```diff
@@ -445,7 +445,7 @@ def restore(self):
         # Respect the source of the pack
         zpsToSort = {}
 
-        pattern = '(ZenPacks\.zenoss\.[a-zA-Z\.]*)'
+        pattern = '(ZenPacks\.[a-zA-Z]\.[a-zA-Z\.]*)'
```
jpeacock-zenoss: I think this will only match names that use a single character between the periods, e.g. `ZenPacks.a.name`.
Shouldn't this pattern be: `(ZenPacks\.[a-zA-Z]+\.[a-zA-Z\.]*)`?

rphillipsz: Yep.  You could actually remove the interior pattern and replace the second pattern * with + to simplify this more.

PR #2959, opened 2018-01-25
Improve SNMP varbind capture.

jpeacock-zenoss:
When a series of trap variables are detected, a 'sequence' event detail is added to show the order of the sequence.  E.g., given the following SNMP trap varbinds:
```
   someVar.0  Data0
   someVar.1  Data1
```
the following details are added to the event:
```
   someVar.0  Data0
   someVar.1  Data1
   someVar.sequence  0,1
```
Fixes ZEN-29016.

### DIFF
--- a/Products/ZenEvents/tests/test_zentrap.py
+++ b/Products/ZenEvents/tests/test_zentrap.py
@@ -1,3 +1,4 @@
+import base64
 import logging
 
 from struct import pack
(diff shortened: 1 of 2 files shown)


Products/ZenEvents/tests/test_zentrap.py
```diff
@@ -70,8 +76,670 @@ def test_decode_unexpected_object_type(self):
...
+        task = MockTrapTask({})
+        eventType, result = task.decodeSnmpv1(("localhost", 162), pckt)
+        self.assertEqual(result["snmpV1SpecificTrap"], 5)
+
+    def test_EnterpriseOIDWithoutExtraZero(self):
+        pckt = self.makePacket()
+        task = MockTrapTask({})
+        eventType, result = task.decodeSnmpv1(("localhost", 162), pckt)
+        self.assertEqual(eventType, "1.2.3.4.5")
+        self.assertEqual(result["oid"], "1.2.3.4.5")
```
lundybernard: I would combine these tests (line 158 - 200), because they are all testing the same input to the function under test, and validating attributes on the same result that it returns.

You could create an expected_result dictionary, and assert that it is equal to the returned result, but this would make the test more brittle.  It can be a good way to ensure future developers review these tests when they make a change to the method's result value.

Products/ZenEvents/tests/test_zentrap.py
```diff
@@ -70,8 +76,670 @@ def test_decode_unexpected_object_type(self):
...
+
+        pckt.trap_type = 5
+        eventType, result = task.decodeSnmpv1(("localhost", 162), pckt)
+        self.assertEqual(eventType, "egpNeighorLoss")
+        self.assertEqual(result["snmpV1GenericTrapType"], 5)
+
+        pckt.trap_type = 6
+        eventType, result = task.decodeSnmpv1(("localhost", 162), pckt)
+        self.assertEqual(eventType, "1.2.3.4.5")
+        self.assertEqual(result["snmpV1GenericTrapType"], 6)
```
lundybernard: I would break these up into separate tests, since they are testing decoding on different tryap types.

You can deduplicate the  ```pckt = self.makePacket() \ task = MockTrapTask({})``` boilerplate by adding it to a setUp method for the test Class.  like so:

```
def setUp(self):
    self.pckt = self.makePacket()
    self.task = MockTrapTask({})
```

Products/ZenEvents/zentrap.py
```diff
@@ -814,7 +796,8 @@ def dateandtime(value):
     @staticmethod
     def oid(value):
         if isinstance(value, tuple) and len(value) > 2:
```
lundybernard: we should be able to combine this into a single if/and/and/... statement

```
if (isinstance(value, tuple)
    and len(value) > 2
    and value[0] in (0, 1, 2)
    and all(isinstance(i, int) for i in value)
):
```

PR #2971, opened 2018-01-31
ZEN-29457 Update instances during Rename/Move/Delete Manufacturer's p…

karolzen:
…roduct
https://jira.zenoss.com/browse/ZEN-29457

### DIFF
--- a/Products/Zuul/facades/manufacturersfacade.py
+++ b/Products/Zuul/facades/manufacturersfacade.py
@@ -73,6 +73,8 @@ class ManufacturersFacade(TreeFacade):
         prodName = params['oldname']
         prod = manufacturer.products._getOb(prodName, None)
         if prod:
+            instances = [i.device() for i in prod.instances()]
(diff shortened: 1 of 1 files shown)


Products/Zuul/facades/manufacturersfacade.py
```diff
@@ -73,6 +73,8 @@ def editProduct(self, params=None):
         prodName = params['oldname']
         prod = manufacturer.products._getOb(prodName, None)
         if prod:
+            instances = [i.device() for i in prod.instances()]
```
ssOlexBaiko: if it possible to use here "generator comprehension"?
E.g. (i.device() for i in prod.instances())

karolzen: Unfortunately we need to store information about instances before product is ie. renamed. After renaming UID of product is changed and data about instances is lost.

ssOlexBaiko: got you

PR #2986, opened 2018-02-08
zing-connector

smousa:


Products/DataCollector/ApplyDataMap.py
```diff
@@ -216,6 +247,9 @@ def _applyDataMapImpl(self, device, datamap):
             self.num_obj_changed,
             logname)
 
+        handle_datamap(device, datamap, self.context)
```
fraibacas: Shouldnt we call handle_datamap once we are sure this has successfully completed?https://github.com/zenoss/zenoss-prodbin/blob/004974ca24e7c265a6f704bb59d236f47b2595e6/Products/DataCollector/ApplyDataMap.py#L152-L163 in case the transact fails bc of a conflict error or something?

Products/DataCollector/ApplyDataMap.py
```diff
@@ -10,9 +10,14 @@
 
 import sys
 from collections import defaultdict
+import urlparse
+
 import logging
 log = logging.getLogger("zen.ApplyDataMap")
 
+import requests
+session = requests.Session()
```
fraibacas: Should we wrap this and the method handle_datamap in a class so we have all the zing datamap forwarding logic encapsulated?

smousa: sure

PR #2992, opened 2018-02-13
ZEN-29570 Zenoss Core - AttributeError message when attempt to add Ci…

ISPHOST:
…sco device

### DIFF
--- a/Products/Zuul/routers/device.py
+++ b/Products/Zuul/routers/device.py
@@ -1576,7 +1576,8 @@ class DeviceRouter(TreeRouter):
         if deviceByIp and organizer.getZ('zUsesManageIp', True) \
                 or deviceByName and deviceClass == deviceByName.getDeviceClassName():
             primaryId = deviceByName.getPrimaryId() if isinstance(deviceByName, Device) \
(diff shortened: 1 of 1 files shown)


Products/Zuul/routers/device.py
```diff
@@ -1575,7 +1575,8 @@ def addDevice(self, deviceName, deviceClass, title=None,
         deviceByName = facade.getDeviceByName(safeDeviceName)
         if deviceByIp and organizer.getZ('zUsesManageIp', True) \
                 or deviceByName and deviceClass == deviceByName.getDeviceClassName():
-            primaryId = deviceByName.getPrimaryId() if deviceByName.getDeviceClassName() == deviceClass else deviceByIp.getPrimaryId()
+            primaryId = deviceByName.getPrimaryId() if isinstance(deviceByName, Device) \
+                    and deviceByName.getDeviceClassName() == deviceClass else deviceByIp.getPrimaryId()
```
jpeacock-zenoss: The `if` statement allows `deviceByIp` to have a `None` value when `deviceByName` is a valid device object.  The `deviceByIp.getPrimaryId()` call should also be protected with a check for validity as well.

PR #2994, opened 2018-02-14
 ZEN-29355: create IpAddress on device creation

rderkach:


Products/ZenModel/tests/testFileSystem.py
```diff
@@ -29,11 +29,11 @@ def afterSetUp(self):
 
 
     def testSetManageIp(self):
-        self.fs.setManageIp('1.2.3.4/24')
-        self.assert_(self.fs.getManageIp() == '1.2.3.4/24')
-        self.assert_(self.dev.getManageIp() == '1.2.3.4/24')
-        self.dev.setManageIp('2.3.4.5/24')
-        self.assert_(self.fs.getManageIp() == '2.3.4.5/24')
```
jpeacock-zenoss: `setManageIp` needs to be able to accept `/24` IP notation.  This is likely to break existing code that expects `getManageIp` to accept such IP addresses.

Products/ZenModel/migrate/addDeviceIpAddrRelation.py
```diff
@@ -0,0 +1,45 @@
...
+        log.info("Updating relationships on Devices")
+        for dev in dmd.Devices.getSubDevices():
+            try:
+                dev.buildRelations()
+            #some objects can fail relations building process
+            except (Exception,):
+                continue
+
+        log.info("Updating relationships on Ip Addresses")
+        catalog = ICatalogTool(dmd.Networks)
```
fraibacas: ICatalogTool is deprecate. Use IModelCatalogTool

rderkach: Yep, thanks.

Products/ZenModel/IpAddress.py
```diff
@@ -66,6 +66,7 @@ class IpAddress(ManagedEntity, IpAddressIndexable):
     _relations = ManagedEntity._relations + (
         ("network", ToOne(ToManyCont,"Products.ZenModel.IpNetwork","ipaddresses")),
         ("interface", ToOne(ToMany,"Products.ZenModel.IpInterface","ipaddresses")),
+        ("manageDevice", ToOne(ToOne,"Products.ZenModel.Device","ipaddress")),
```
jpeacock-zenoss: How is this relationship going to work with MultiRealmIP?

Products/ZenModel/tests/testDevice.py
```diff
@@ -67,6 +67,32 @@ def testManage_createDeviceDupIp(self):
...
+        tmpIface = IpInterface('testNIC')
+        self.dev.os.interfaces._setObject('testNIC', tmpIface)
+        self.iface1 = self.dev.getDeviceComponents()[0]
+        self.iface1.addIpAddress(testIp)
+
+        ip = self.dev.getNetworkRoot().findIp(testIp)
+        self.assert_(ip is not None)
+
+        dev = manage_createDevice(self.dmd, 'mydevice', '/', manageIp=testIp)
+        self.assert_(dev is not None)
```
jpeacock-zenoss: There should be a check to validate that the `self.dev` and `dev` devices have different `manageIp` values.

Products/ZenModel/tests/testDevice.py
```diff
@@ -70,19 +70,21 @@ def testManage_createDeviceDupIp(self):
...
+        iface.addIpAddress(testIp)
 
-        ip = self.dev.getNetworkRoot().findIp(testIp)
+        ip = dev1.getNetworkRoot().findIp(testIp)
         self.assert_(ip is not None)
 
-        dev = manage_createDevice(self.dmd, 'mydevice', '/', manageIp=testIp)
-        self.assert_(dev is not None)
+        dev2 = manage_createDevice(self.dmd, 'myseconddevice', '/', manageIp=testIp)
+        self.assertNotEqual(dev1.manageIp, dev2.manageIp)
```
rderkach: @jpeacock-zenoss added

Products/ZenModel/tests/testDevice.py
```diff
@@ -231,13 +258,6 @@ def testSetManageIp(self):
         self.dev.setManageIp(testIp)
         self.assertEqual(self.dev.getManageIp(), testIp)
 
-        # Need a network interface to register an IP in catalog
-        from Products.ZenModel.IpInterface import IpInterface
-        tmpIface = IpInterface('testNIC')
-        self.dev.os.interfaces._setObject('testNIC', tmpIface)
-        self.iface1 = self.dev.getDeviceComponents()[0]
-        self.iface1.addIpAddress('1.2.3.4')
```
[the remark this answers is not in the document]
rderkach: I removed it from here because before my changes it was necessary to create an Interface with Ip Addr to emulate modeled device behavior since afterward, we were using that IP for checking duplicate devices. Now that Ip address is created when the device is created with manageIp even if there is no Interface on it.

But this testcase with addIpAddress is valid, so I will put it back, to check case when device firstly create an Ip and then Interface also try to assign that ip to itself.

rderkach:
I performed some testing regarding migration

Migration on 1200 fake device without components and no IpAddresses

>Building relations took 3970.008 ms
Whole process
real    1m19.717s
user    0m56.248s
sys     0m4.733s

Migration on some client system with 15000 real devices
>Building relations took 59766.446 ms
Whole process
real    7m42.590s
user    4m20.685s
sys     0m34.017s

Products/ZenModel/migrate/addDeviceIpAddrRelation.py
```diff
@@ -40,6 +36,22 @@ def cutover(self, dmd):
             except (Exception,):
                 continue
 
+        log.info("Updating relationships on Devices")
+        for dev in dmd.Devices.getSubDevices():
+            try:
+                dev.buildRelations()
+                if dev.manageIp:
+                    ipobj = dev.getNetworkRoot().findIp(dev.manageIp)
```
StevePC: perhaps we should check to see if the device already has an ipAddress object in it's relation?

StevePC:
Did you re-run the perofrmance test after fixing the migration?

rderkach:
Yes, I did re-run test after adding changes to migration. 

I tested in the worst case scenario 
The system with 2000 devices, all the devices had manageIP, so migration created an IpAddr object for all of them.

>migrate without commit took 54873.659 ms
> Full process
>real     2m52.377s
>user    2m15.365s
>sys      0m6.677s

Products/ZenModel/migrate/addDeviceIpAddrRelation.py
rderkach: I suppose there should be no ipAddress object in device relation before running this migration.

StevePC: That assumes this migration will only ever run once.  Probably likely, but in the case where the migration runs but the upgrade fails after, is it possible that the migration will re-run?  We don't really want to end up in a situation where it tries to make a bunch of duplicate IpAddress objects.

jpeacock-zenoss: There's a `if ipobj` statement below which sets up the relation if the IpAddress object already exists.

PR #2996, opened 2018-02-15
ZEN-29623 Theme selector

carlcamera:
provides ability to select user interface theme and persist changes across logins.


Products/ZenModel/UserSettings.py
```diff
@@ -246,6 +246,13 @@ def getUserSettingsUrl(self, userid=None):
         if uf: return uf.getPrimaryUrlPath()
         return ""
 
+    def getUserTheme(self):
+        ut = self.getUserSettings().userTheme
+        if ut: return ut
+        return "z-cse"
```
jpeacock-zenoss: More Pythonic:
```
    return ut if ut else "z-cse"
```

carlcamera: okay. i'm all for more pythonic code.

PR #2999, opened 2018-02-16
ZEN-29549:  Disable pipe component tracking except in debug

lundybernard:
This will slow the memory leak by disabling execution time tracking on individual pipeline components, except in debug mode.  the underlying memory leak is still an issue, but restarts will be required less frequently.


Products/ZenEvents/zeneventd.py
```diff
@@ -145,20 +145,15 @@ class when it is done with the message
                     # extract event from message body
                     zepevent = ZepRawEvent()
                     zepevent.event.CopyFrom(message)
-                    if log.isEnabledFor(logging.DEBUG):
-                        log.debug("Received event: %s",
-                                  to_dict(zepevent.event))
+                    log.debug("Received event: %s", to_dict(zepevent.event))
```
ecifreo: I've found these checks to be redundant too. The pro argument for them is that you don't need to allocate all those logging strings and format them first, only to throw that work away when the logging subsystem checks the level. It's debatable how much that saves though....maybe when logging some very large message with many format strings.

PR #3019, opened 2018-02-22
ZEN-29254 Use separate Solr instance by BaseTestCase

karolzen:
https://jira.zenoss.com/browse/ZEN-29254


bin/runtests
```diff
@@ -1,471 +1,37 @@
-#! /usr/bin/env python
+#!/bin/sh
```
jpeacock-zenoss: To be sure, this should be `/bin/bash`.  On Ubuntu, `/bin/sh` is not `bash`.

karolzen: 👍

jpeacock-zenoss:
retest this

jpeacock-zenoss:
retest this

PR #3030, opened 2018-02-26
Drop dms

fraibacas:
- Extract the objects' relevant information before buffering it in the current tx instead of storing the whole object
- Send facts in batches to zing connector
- Drop facts without uuid
- Send a dummy fact per processed device to ensure Zing has facts for all devices

### DIFF
--- a/Products/DataCollector/zing/DatamapHandler.py
+++ b/Products/DataCollector/zing/DatamapHandler.py
@@ -12,7 +12,7 @@ import requests
 import urlparse
 import transaction
 
-from Products.DataCollector.zing.fact import FactContext, Fact, facts_from_datamap, serialize_facts, FactKeys
+from Products.DataCollector.zing.fact import FactContext, Fact, facts_from_datamap, serialize_facts
 from Products.ZenUtils.GlobalConfig import getGlobalConfiguration
 
 
@@ -70,18 +70,6 @@ class ZingDatamapHandler(object):
             log.info("Zing AfterCommitHook added. Zing state added to current transaction.")
         return zing_tx_state
 
-    def _build_fact_for_device(self, device):
-        device_fact = Fact.from_object(device)
-        allowed_data_keys = { FactKeys.NAME_KEY }
-        allowed_metadata_keys = { FactKeys.CONTEXT_UUID_KEY, FactKeys.META_TYPE_KEY }
-        for k in device_fact.data.iterkeys():
-            if k not in allowed_data_keys:
-                del device_fact.data[k]
-        for k in device_fact.metadata.iterkeys():
-            if k not in allowed_metadata_keys:
-                del device_fact.metadata[k]
-        return device_fact
-
     def add_datamap(self, device, datamap):
         """ adds the datamap to the ZingTxState in the current tx"""
         if self.zing_connector_url: # dont bother to store maps if the url no set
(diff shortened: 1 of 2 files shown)


Products/DataCollector/zing/DatamapHandler.py
```diff
@@ -88,21 +94,36 @@ def process_datamaps(self, tx_success, zing_state):
...
+            # add dummy facts for the processed devices
+            device_facts = zing_state.devices_fact.values()
+            if device_facts:
+                facts.extend(device_facts)
+            self._send_facts_in_batches(facts)
         except Exception as ex:
             log.warn("Exception sending datamaps to zing-connector. {}".format(ex))
 
-    def _send_datamap(self, device, datamap, context):
+    def _send_facts_in_batches(self, facts, batch_size=500):
```
bharatZen: Does this also need an alternate time qualifier? For example, as in batch_size=500 || batch_time=300s . So the batch is sent when either of the size or time constraints are met

We need to handle the case wherein the number of facts are less than the batch size and so we are not waiting indefinitely (or rather, for the remodel interval ) to complete the batch size

Products/DataCollector/zing/DatamapHandler.py
```diff
@@ -88,21 +94,36 @@ def process_datamaps(self, tx_success, zing_state):
...
+            batch = facts[:batch_size]
+            del facts[:batch_size]
+            self._send_facts(batch)
+
+    def _send_facts(self, facts):
         try:
-            dm = serialize_datamap(device, datamap, context)
-            if not dm: # nothing to send
+            serialized = serialize_facts(facts)
+            if not serialized: # nothing to send
```
bharatZen: Do we need to log this as a warning or error? I'm assuming that normally we would always get a serialized version, right?

Products/DataCollector/zing/fact.py
```diff
@@ -8,6 +8,15 @@
 
 import time
 
+
+class FactKeys(object):
+    CONTEXT_UUID_KEY = "contextUUID"
+    META_TYPE_KEY = "meta_type"
+    NAME_KEY = "name"
+    MEM_CAPACITY_KEY = "mem_capacity"
```
bharatZen: This particular one seems rather type specific? The others are good generally applicable ones. Just curious why this got picked up here...

Products/DataCollector/zing/fact.py
```diff
@@ -69,55 +88,86 @@ def default(self, o):
...
-    """
-    Converts a datamap to a JSON-encoded list of facts.
-    """
+class FactContext(object):
+
+    def __init__(self, obj):
+        self.uuid = None
+        self.meta_type = None
+        self.name = None
+        self.mem_capacity = None
```
bharatZen: Similar to earlier comment, it seems odd that we are capturing mem capacity as a context

Products/DataCollector/zing/fact.py
```diff
@@ -69,55 +88,86 @@ def default(self, o):
...
+        try:
+            self.name = obj.titleOrId()
+        except Exception:
+            pass
+
+        from Products.ZenModel.Device import Device
+        if isinstance(obj, Device):
+            self.is_device = True
+            try:
+                self.mem_capacity = obj.hw.totalMemory
```
bharatZen: Ah -- it's the only device specific field that we can reliably extract?

Products/DataCollector/zing/fact.py
fraibacas: This comes from the initial implementation we had: https://github.com/zenoss/ZenPacks.zenoss.ZingConnector/blob/develop/ZenPacks/zenoss/ZingConnector/fact.py#L91-L122
it's device specific information that we can reliably extract

Products/DataCollector/zing/DatamapHandler.py
fraibacas: that should be `if not facts`

Products/DataCollector/zing/DatamapHandler.py
fraibacas: We always send all the facts once the tx is successfully committed

bharatZen: Ok, that works

PR #3036, opened 2018-02-27
Datamap to Facts improvements

fraibacas:
- Add a connection timeout for connections to zing-connector
- If the fact data contains lists of something other than scalar, convert the non scalar values to string
- If zing-connector fails to process a batch of Facts returning a 500, re-try sending facts one by one

### DIFF
--- a/Products/DataCollector/zing/fact.py
+++ b/Products/DataCollector/zing/fact.py
@@ -78,7 +78,14 @@ class _FactEncoder(JSONEncoder):
         data_out = {}
         for k, v in data_in.iteritems():
             if isinstance(v, list) or isinstance(v, tuple) or isinstance(v, set):
-                data_out[k] = sorted(v)
+                # whatever comes in the list, tuple etc, needs to be scalar
+                # if not, cast it to string
+                values = []
+                for x in v:
+                    if not isinstance(x, (str, int, float, bool)):
+                        x = str(x)
+                    values.append(x)
+                data_out[k] = sorted(values)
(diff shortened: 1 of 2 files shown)


Products/DataCollector/zing/fact.py
```diff
@@ -78,7 +78,14 @@ def _tweak_data(self, data_in):
         data_out = {}
         for k, v in data_in.iteritems():
             if isinstance(v, list) or isinstance(v, tuple) or isinstance(v, set):
-                data_out[k] = sorted(v)
+                # whatever comes in the list, tuple etc, needs to be scalar
+                # if not, cast it to string
+                values = []
+                for x in v:
+                    if not isinstance(x, (str, int, float, bool)):
```
bharatZen: should long be accounted for explicitly?

Products/DataCollector/zing/fact.py
```diff
@@ -78,7 +78,14 @@ def _tweak_data(self, data_in):
...
+                        x = str(x)
```
bharatZen: Should this be logged so we can track it better?

Products/DataCollector/zing/fact.py
```diff
@@ -78,7 +78,14 @@ def _tweak_data(self, data_in):
...
+                    values.append(x)
```
bharatZen: We lose the "grouping" aspect of the list-of-list or the list-of-dict in this case. That's okay for now, but should think about how we can propagate that info...

PR #3053, opened 2018-03-14
ZEN-29740 Cannot load devices in dropdown - 'Test Process Class Regul…

ISPHOST:
…ar Expressions' modal

### DIFF
--- a/Products/Zuul/routers/device.py
+++ b/Products/Zuul/routers/device.py
@@ -392,13 +392,12 @@ class DeviceRouter(TreeRouter):
         audit('UI.Device.Edit', uid, data_=data)
         return DirectResponse()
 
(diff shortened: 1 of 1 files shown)


Products/Zuul/routers/device.py
```diff
@@ -398,6 +398,7 @@ def getDeviceUuidsByName(self, query="", start=0, limit=25, page=1, uuid=None):
         If uuid is set, ensures that it is included in the returned list.
         """
         facade = self._getFacade()
+        query = "*" if not query else query
```
rderkach: Can we just put `query="*"` as an argument of a method

`def getDeviceUuidsByName(self, query="*", start=0, limit=25, page=1, uuid=None):`

PR #3055, opened 2018-03-14
Fixing ZEN-29391 - Adding manageIp as an internalVars element

jlampe:
Fixing ZEN-29391 - When batchloading a device before fix will always print an error about 'function manageIp...' - after fix should be clean


StevePC:
This change may have a side effect of ensuring that batchload will not change the manageIp of existing devices.  If that is acceptable, then this change is ok.

PR #3056, opened 2018-03-15
make zing connector ping url configurable

fraibacas:


tgjonestx:
@fraibacas - sorry I missed your slack. The problem is not just the endpoint, but also the port number. I have a separate PR to address that, so I closed this one.

PR #3059, opened 2018-03-16
ZEN-29782 Cann't add new child item in Reports tree

bmoskalyk:


Products/ZenUI3/browser/resources/js/zenoss/Reports.js
```diff
@@ -119,7 +119,8 @@ Ext.define('Zenoss.ReportTreePanel', {
         var selNode = this.getSelectionModel().getSelectedNode(),
             tree = this,
             newNode;
-        if(!selNode.data.uid){
+        // we should also check is something is selected;
+        if(!selNode || !selNode.data.uid){
```
jpeacock-zenoss: The `!selNode.data.uid` operand of the `||` operator is irrelevant here because if `!selNode` is true, then it doesn't matter whether `!selNode.data.uid` is true or false.  If `!selNode` is false, then `!selNode.data.uid` can _never_ be true.

The `if` statement can be simplified to
```
   if (!selNode) {
```

bmoskalyk: i thought that there could be two options
1.) no selected node
2.) selected node has no "uid"
that why i check both statesments

jpeacock-zenoss: Hmm.  I blame it on trying to do truth tables in my head without my coffee.

Although, is it ever the case that `selNode` can be a 'select node' object, but `selNode.data.uid` is null (or equivalent)?

bmoskalyk: I check tree store data and it sims that "selNode" always has uid in data, so it is enought to check only for selected node.

PR #3075, opened 2018-03-23
ZEN-29800 Class misassigned when the device is added with the default…

ISPHOST:
… Device Class


Products/ZenUI3/browser/resources/js/zenoss/itinfrastructure.js
```diff
@@ -106,7 +107,10 @@ Ext.onReady(function () {
             root: 'deviceClasses',
             totalProperty: 'totalCount',
             model: 'Zenoss.model.Name',
-            directFn: REMOTE.getDeviceClassesToAdd
+            directFn: REMOTE.getDeviceClassesToAdd,
+            filters: function(rec) {
+                return rec.get('name') != "/" && !!rec.get('name');
```
jpeacock-zenoss: I think these operands should be reversed.  First test whether there is a value, then test what that value is.

PR #3078, opened 2018-03-26
ZEN-29723 The login page is showing an overlapping error message after an invalid login

rzino:


Products/ZenModel/skins/zenmodel/login_form.pt
```diff
@@ -1,140 +1,171 @@
...
     <head>
+        <meta charset="utf-8">
         <link rel="shortcut icon"
               type="image/x-icon"
               href="/zport/dmd/favicon.ico"/>
         <title>Login</title>
         <style>
+            * {
+                zoom: 1; /* IE7 compatibility */
+            }
```
jpeacock-zenoss: IE11 is the minimum version of IE that Zenoss supports.

rzino:
Yes, backgroung image was changed. Previously was used this one:

zenoss-login

As you can see there are labels and inputs on this image. In the previous solution, we had transparent inputs that were overlayed over inputs from the image and do not have labels in HTML at all (just on image), everything was positioned using ``position: absolute`` over the image and it caused problems with the flexibility of the HTML structure on this page, it was hard to make a change and do not break a whole page. So I decided slightly change an approach to make a page more flexible.

Products/ZenModel/skins/zenmodel/login_form.pt
rzino: I can remove a code dedicated to IE11-, or even switch to flexbox for more flexibility

rzino:
I tested my changes with the UCSXSkin ZenPack. Nothing was broken, everything works as expected.
I looked through this zenpack and it has own CSS, background image, and template for login which overrides all my changes.

ucsxskin zenpack

PR #3080, opened 2018-03-29
ZEN-28829 RM - ADVANCED - Settings - Users - On IE11, updating user's…

rzino:
… information (i.e. roles) doesn't show confimation message


rzino:
Demo is here: https://jira.zenoss.com/browse/ZEN-28829

PR #3081, opened 2018-03-29
ZEN-29815 Filtering is broken for Storage Pools -> Hard Disks on RM 6…

rzino:
….2.x


rzino:
Demo in attachments: https://jira.zenoss.com/browse/ZEN-29815

PR #3091, opened 2018-04-09
ZEN-29823 Make ZEP more robust

lundybernard:
Add a timeout to ZEP, to catch long-running event transforms and pipeline components, and convert them to processing errors if they exceed the time limit.

### DIFF
--- a/Products/ZenEvents/tests/test_zeneventd.py
+++ b/Products/ZenEvents/tests/test_zeneventd.py
@@ -18,12 +18,6 @@ from zenoss.protocols.protobufs.zep_pb2 import EventActor
 PATH = {'zeneventd': 'Products.ZenEvents.zeneventd'}
 
 
-class DummyPipe(EventProcessorPipe):
-
-    def __call__(self, eventContext):
-        return eventContext
-
-
 class EventPipelineProcessorTest(TestCase):
 
     def setUp(self):
@@ -91,19 +85,13 @@ class EventPipelineProcessorTest(TestCase):
             exception_event.event.message
         )
 
-    def test_synchronize_with_database(self):
-        '''if self.SYNC_EVERY_EVENT:
-            doSync = True
-        else:
-            # sync() db if it has been longer than self.syncInterval
-            # since the last time
-            currentTime = datetime.now()
-            doSync = currentTime > self.nextSync
-            self.nextSync = currentTime + self.syncInterval
+    class ErrorPipe(EventProcessorPipe):
+        ERR = Exception('pipeline failure')
 
-        if doSync:
-            self.dmd._p_jar.sync()
-        '''
+        def __call__(self, eventContext):
+            raise self.ERR
+
+    def test_synchronize_with_database(self):
         self.epp._synchronize_with_database()
         self.dmd._p_jar.sync.assert_called_once_with()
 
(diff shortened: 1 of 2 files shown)


dougsyer:
It would help if there was a way to create a lower priority pipeline that can handle long running requests.  I know what transforms run long id just tag those events somehow in the mapping or something to run om a best effort speed.

lundybernard:
@dougsyer 
I have seen many instances where an extremely slow transform, or a new pipeline component added by a zenpack, can block ZEP for upwards of 30 minutes.  You could probably create an infinite loop in a transform and totally halt processing (not yet tested).  In some cases, when we get a flood of events that take an inordinate amount of time to process, even dozens of instances of zeneventd can grind to a halt.  The Timeout context wraps the entire pipeline, and guards against both slow transforms and pipeline components.

# Handling known slow transformations:
As I've implemented the timeout, it will catch a signal, and raise a TimeoutException from the current scope.

There are a few good ways to guard against the timeout exception if you need to allow a long-running transform.  I personally like explicitly marking these slow transformations, and setting an upper limit on them.

## Use the signal library to adjust the timeout durration
```
import signal
signal.alarm(300)  # 5min timeout
```
if you set `signal.alarm(0)` the timeout will be disabled (there can only be one alarm set at a time).  This method is the simplest, and allows you to extend the default timeout, and still guard against issues with pipeline components after transforms take place.

## Use the Timeout context manager within a long-running transform to manage slow sections of code.
```
2018-04-10 14:46:05,900 INFO zen.Events: Transform error timeout from transform; aborting transaction
2018-04-10 14:46:05,903 WARNING zen.Events: Error processing transform/mapping on Event Class /Status/Ping
Problem on line 9: TimeoutError: timeout from transform
    sleep(10)

Transform:
     1 from time import sleep
     2 import signal
     3 from Products.ZenEvents.zeneventd import Timeout
     4 
     5 with Timeout(evt, seconds=2, error_message='do not timeout'):
     6     sleep(1)
     7     
     8 with Timeout(evt, seconds=1, error_message='timeout from transform'):
 *   9     sleep(10)

2018-04-10 14:46:05,911 WARNING zen.Events: Event transform took 2.0 seconds (threshold 2.0 seconds), event context is {_zepRawEvent:{'clear_event_class': [u'/Status/Wmi/Conn'], 'event': {'uuid': u'cd9bace4-3cf7-11e8-a069-0242ac11001a', 'tags': [{'type': u'zenoss.event.event_class', 'uuid': [u'cb0940f9-6435-47c1-9cec-31670aff9eec', u'26093055-f8a4-4f97-9c77-0ac3f2a8f838']}], 'first_seen_time': 1523389563891L, 'monitor': u'localhost', 'actor': {'element_identifier': u'false_device', 'element_type_id': 1}, 'summary': u'l TEST PING EVENT', 'created_time': 1523389563891L, 'message': u'l TEST PING EVENT', 'event_key': u'', 'event_class': u'/Status/Ping', 'severity': 5}}}, transform is: from time import sleep ...
```
Note, after you exit a Timeout context, you would need to set `signal.alarm(x)` again, to guard against slow pipeline components, nesting is not supported in this version.

## Catch the TimeoutError, and handle it in your transform/zp
catching the exception will break out of the try-block, so will probably be useful in fewer cases than manipulating the Timeout directly.
```
from zenoss.Products.ZenEvents.zeneventd import TimeoutError
try: ...
except TimeoutError: ...
```

lundybernard:
closing this PR to re-open it to merge into develop.

PR #3103, opened 2018-04-16
Link to Zenoss Cloud from Collection Zone Device

carlcamera:


csmouse:
"View in Zenoss Cloud" implies that the CZ is separate from Zenoss Cloud, which isn't true.

Let's come up with some alternatives.

PR #3127, opened 2018-04-26
ZEN-29852: Fix the AWS account password representation as plain text

drevak:


Products/ZenUtils/Utils.py
```diff
@@ -2228,3 +2230,21 @@ def getQueryArgsFromRequest(request):
...
+            continue
+        for key, value in fields.iteritems():
+            if isinstance(value, Password):
+                passwordFields.add(key)
+    return passwordFields
+
+
+def maskSecureProperties(data, secure_properties=[]):
+    [data.update({prop: '*' * len(data[prop])})
+        for prop in secure_properties if data.get(prop, None)]
```
jpeacock-zenoss: This is considered an abuse of the list comprehension syntax.  Your intent is to modify the contents of a parameter, not to create a list.  Please rewrite this as a loop that expresses your intent more accurately.  Additionally, using a list comprehension like this creates a list object that's unused.

drevak: Thank you for your reviewing. I rewrote function by using the loop and got rid of that comprehension.

PR #3131, opened 2018-04-27
Fix Auth0 to use silent authentication, checking for callback errors.

rphillipsz:
https://jira.zenoss.com/browse/ZEN-29947

There were problems with the method used to authenticate with Auth0, and the errors were never caught.  This updates the challenge to use silent authentication, catching the error to present a login screen from Auth0 on error.  Now, authenticating from either Zing or CZ will allow access to the other, and logging out of CZ actually works (previously an automatic renewal happened), fixing SSO.

Changed the "Sign out" to "Logout" to match the ZING UI (per Smouse request)

Single-Sign-Off does **not** work yet.  Each UI logout is independent.  I'll create a new ticket to address this related, but separate, problem.

### DIFF
--- a/Products/ZenUtils/Auth0/Auth0.py
+++ b/Products/ZenUtils/Auth0/Auth0.py
@@ -15,11 +15,10 @@ from Products.PluggableAuthService.interfaces.plugins import (IExtractionPlugin,
                                                               IRolesPlugin)
 from Products.PluggableAuthService.plugins.BasePlugin import BasePlugin
 from Products.PluggableAuthService.utils import classImplements
-from Products.ZenUtils.AuthUtils import getJWKS, publicKeysFromJWKS, getBearerToken
+from Products.ZenUtils.AuthUtils import getJWKS, publicKeysFromJWKS
 from Products.ZenUtils.CSEUtils import getZenossURI
 from Products.ZenUtils.GlobalConfig import getGlobalConfiguration
 from Products.ZenUtils.PASUtils import activatePluginForInterfaces, movePluginToTop
-from Products.ZenUtils.Utils import getQueryArgsFromRequest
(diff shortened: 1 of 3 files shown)


Products/ZenUtils/Auth0/Auth0.py
```diff
@@ -118,10 +118,12 @@ def storeIdToken(id_token, session, conf):
                              audience=conf['clientid'],
                              issuer=conf['tenant'])
 
-        sessionInfo = session.setdefault(Auth0.session_key, SessionInfo())
+        sessionInfo = SessionInfo()
```
jafcrocker: I don't think you need to construct a new SessionInfo object every time; as long as you set the refresh_token, the old code would have worked.  If we are making changes to the SessionInfo object elsewhere that we expect to persist, then this change could break things.

Products/ZenUtils/Auth0/Auth0Callback.py
```diff
@@ -36,40 +35,53 @@ def __call__(self):
...
+            # We need to make a new request to authorize without the
+            # prompt parameter so that auth0 presents a login screen.
+            uri = "%sauthorize?" % conf['tenant'] + \
+                  "response_type=code&" + \
+                  "client_id=%s&" % conf['clientid'] + \
+                  "state=%s&" % state_arg + \
+                  "scope=openid offline_access&" + \
+                  "redirect_uri=%s/zport/Auth0Callback" % zenoss_uri
+            return self.request.response.redirect(uri)
+        else:
```
jafcrocker: This `else` is not necessary.  The last line of the `if` is a return, and the rest of the code is all in the else clause.  It would make the code simpler and easier to review if we took it out.

PR #3132, opened 2018-04-30
ZEN-29896: Solr: zeneventd eventContext solr timeout

olst:


zedieck:
Closing this PR since the fix here is not desired.

PR #3135, opened 2018-05-01
ZEN-29827: Add CZAdmin role; Update permissions for actions we want to restrict

iancmcc:
This changeset moves access to several screens/actions under the Advanced tab to be based on individual permissions, then creates a CZAdmin role (to represent the SaaS administrator) and gives it all those permissions. They're then selectively removed from the ZenManager role (representing the customer admin).

Details of the permissions are in this sheet: https://docs.google.com/spreadsheets/d/1zjHuNsnW9O3z3KzOaS9eKN-CB0GuCCUZScRb0OhHWEA/edit#gid=0


Products/ZenModel/skins/zenmodel/editUserSettings.pt
```diff
@@ -322,24 +287,12 @@
                         </td>
```
rphillipsz: Should we just remove the change-password elements from the ui for CZ?

iancmcc: That's intentional, until we fully divest ourselves of local users. For the moment, that's how we have to develop.

iancmcc: But only CZ_ADMIN and up have it, so users wouldn't see it even if it did go out.

PR #3140, opened 2018-05-08
ZEN-30008 Drop-down menu doesn't use selected GUI theme

rzino:
Demo: 
* https://jira.zenoss.com/secure/attachment/50903/ZEN-30008-result.mp4
* https://jira.zenoss.com/secure/attachment/50902/result.PNG


Products/ZenUI3/browser/resources/js/zenoss/ExtOverrides.js
```diff
@@ -93,6 +93,10 @@ Ext.override(Ext.util.Sorter, {
         border: false
     });
 
+    Ext.override(Ext.menu.Menu, {
```
carlcamera: Why has this been added?

rzino:
.

Products/ZenUI3/browser/resources/js/zenoss/ExtOverrides.js
rzino: The reason is that regarding the style guide attached in ZEN-29844 our drop-down menu should have borders. 
The `Ext.menu.Menu` inherits `Ext.panel.Panel` and on 92 line of file `ExtOverrides.js` we have an expression that disabled borders for `Ext.panel.Panel` in accordance `Ext.menu.Menu` won't have borders as well.
I did not want to create a chain when we disable borders via javascript and then enable it via CSS `!important` so I decided to do that in this way.

PR #3145, opened 2018-05-11
ZEN-30013 Move "Cannot to connect to server" popup to a flare message in CZ

rzino:
Demo: https://jira.zenoss.com/secure/attachment/50995/zenapi.ip-10-111-23-31_cse_zport_dmd_itinfrastructure%20%284%29.png

### DIFF
--- a/Products/ZenUI3/browser/resources/js/zenoss/zenoss.js
+++ b/Products/ZenUI3/browser/resources/js/zenoss/zenoss.js
@@ -392,7 +392,7 @@ Ext.Direct.on('exception', function(e) {
         return;
     }
     else if (e.code === Ext.direct.Manager.exceptions.TRANSPORT) {
-        Zenoss.flares.Manager.error('Unable to connect to the server. Will retry in a few moments.');
+        Zenoss.message.error('Unable to connect to the server. Will retry in a few moments.');
 
         setTimeout(function () {
             window.location.reload();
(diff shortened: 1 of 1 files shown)


Products/ZenUI3/browser/resources/js/zenoss/zenoss.js
```diff
@@ -391,6 +391,15 @@ Ext.Direct.on('exception', function(e) {
         window.location.reload();
         return;
     }
+    else if (e.code === Ext.direct.Manager.exceptions.TRANSPORT) {
+        Zenoss.message.error('Unable to connect to the server. Will retry in a few moments.');
+
+        setTimeout(function () {
+            window.location.reload();
```
iancmcc: I think you'll also need to cancel this timeout if a request comes back successful within 30 seconds.

rzino: Thanks. Good idea.
I have one more proposal: what if we will change value of `30` seconds to the value that is selected in `Refresh Every` drop-down? 

The reason I am asking is that in `Refresh Every` component value `1 minute` can be selected and accordingly we have to wait exactly such period of time (not `30` seconds) before making any conclusions. Otherwise, the page will be refreshed even if we already able to connect to the server. I mean that no request was made during this 30 seconds so we have not any grounds to determine if everything ok.

zenapi ip-10-111-23-31_cse_zport_dmd_itinfrastructure 5

rzino: Thanks. Good idea.

PR #3151, opened 2018-05-17
ZEN-30073 - Device class does not fill correctly in a form when adding a single device.

bmoskalyk:


Products/ZenUI3/browser/resources/js/zenoss/EventPanels.js
```diff
@@ -1756,12 +1756,15 @@
             }
         },
         refresh: function(){
-            this.directFn({uid:this.uid, keys:['events']}, function(result){
-                if (Zenoss.env.contextUid && Zenoss.env.contextUid !== this.uid) {
-                    return;
-                }
-                this.updateRainbow(result.data.events);
-            }, this);
+            // we always need uid - so no uid no refresh;
```
jplouis: What does this have to do with add device?

Products/ZenUI3/browser/resources/js/zenoss/itinfrastructure.js
```diff
@@ -117,20 +117,17 @@ Ext.onReady(function () {
...
-                        else if (isclass) {
-                            var path = selnode.data.path;
-                            path = path.replace(/^Devices/, '');
-                            if (this.findRecordByValue(path)) {
-                                this.setValue(path);
-                            }
+                        // no needs to take care of "CSE_VIRTUAL_ROOT" in devices uid's
+                        // simply use device path and strip everithing with "Devices" from it;
+                        var selnode = getSelectionModel().getSelectedNode(),
+                            path = selnode && selnode.data.path,
```
jplouis: The comment above says that the root node doesn't have the `path` attr. How doesn't this blow up?

bmoskalyk: Each ExtJs tree has one main root node that is invisible for users and handle all other nodes.
It looks like this
    `root node
         device node
             AWS node
             Azure node
.......................................`
On store callback we select same device class in combo as in tree.
Before my changes it check if selected node is "Device" root or hidden (main root) of tree if so we select first device class otherwise we strip selected node path and search in combo store same.
I simplify it to use only selected node path and to search by it.

Products/ZenUI3/browser/resources/js/zenoss/EventPanels.js
bmoskalyk: While reproducing this issue i notice that when there is nothing selected or selected main root node and "Rainbow" component refreshes we get server error that some function needs two arguments, and i decide that this could be related and easy fixed with this issue.

bmoskalyk: @jplouis  please finish review of this ticket.

jplouis: I'm not seeing this as an issue on alphacorp or betacorp sites. Is this still an issue?

bmoskalyk: it could be an issue on first page load when nothing is selected in device tree.

bmoskalyk: this was changed in scope of ticket ZEN-30131

PR #3155, opened 2018-05-17
Feature/zen 30131

bmoskalyk:
Add Zenoss.env.maybeAddVirtualRoot function to use in other places where we need to handle "CSE_VIRTUAL_ROOT" variable.


bmoskalyk:
Directly in case of this ticket we need CSE_VIRTUAL_ROOT  in "uid" for expanding (directFn) tree nodes in infrastructure/devices, loading devices to device grid, and because of that i can't use Zenoss.render.link fn there.
We need handle all "uid"'s (append CSE_VIRTUAL_ROOT ) on server side or on client, as i did now.
I can use Zenoss.render.link fn in places where we display link and navigate to it on click.

PR #3180, opened 2018-05-29
Feature/zen 30102

bmoskalyk:
Using margins instead off paddings to get right width/height of graph owner panel on init/resize;


Products/ZenUI3/browser/resources/js/zenoss/form/graphPanel.js
```diff
@@ -488,21 +493,20 @@
...
             config.autoScroll = true;
             delete config.html;
 
             var win = Ext.create('Zenoss.dialog.BaseWindow', {
                 cls: 'white-background-panel',
                 layout: 'fit',
-                width: config.width * 1.15,
-                height: config.height * 1.05,
+                // ZEN-30103
+                width: '90%',
```
carlcamera: rather than 90% could we use calc(100% - 40px) or something similar?  Also, apostrophes not needed around value.

Products/ZenUI3/browser/resources/js/zenoss/devdetail.js
```diff
@@ -498,7 +498,8 @@ device_graphs.on('resize', function(panel, width) {
     // always redraw the graphs completely when we resize the page,
     // this way the svg's are the correct size.
     panel.setContext(panel.uid);
-});
+    // add resize buffer to allow user finish resize action and to not resize graphs to often;
```
carlcamera: "too" often. :-)

Products/ZenUI3/browser/resources/js/zenoss/form/graphPanel.js
bmoskalyk: extjs width config apply or percentage value according parent dimensions or number value, we can't use there css calc functionality.

bmoskalyk: but we can calculate -40px of available screen width

PR #3181, opened 2018-05-29
ZEN-30116 - style changes for panel headers and codemirror field when …

bmoskalyk:
…using dark theme


Products/ZenUI3/browser/resources/css/zen-cse.css
```diff
@@ -199,6 +199,10 @@ darker input fields:           #505050
         width: 14px;
     }
 
+    /* ZEN-30116 background changes for panel header */
+    .z-cse .x-panel-header-default {
```
carlcamera: **x-panel-header-default** is used in many places.  if we are targeting a specific header, then we should be more specific in targeting it.

Products/ZenUI3/browser/resources/css/zen-cse.css
```diff
@@ -199,6 +199,10 @@ darker input fields:           #505050
...
+        background-color: #f3f3f3;
```
carlcamera: this line overrides the previous rule for **x-panel-header-default** on line 152.  should we remove line 152?

bmoskalyk: this line override only **background-color**, and we have no need to remove line 152 because there we override original ExtJs css **background-image: none;**

Products/ZenUI3/browser/resources/css/zen-cse.css
bmoskalyk: i suppose that all panels in application should have same header styles that are described in some css rules, and if we want different styles in header in some panel we should write or specific css or add **style** property to panel header

Products/ZenUI3/browser/resources/css/zen-cse.css
carlcamera: Fair enough. I'll accept that answer.

PR #3182, opened 2018-05-29
Update login/logout to use ZC

rphillipsz:
## NOT READY

Remove clientid/secret from CZ and use ZC for login/logout.


Products/ZenUtils/Auth0/Auth0.py
```diff
@@ -189,17 +178,24 @@ def extractCredentials(self, request):
...
+        cookies = parseCookies(request)
+        token = cookies.get(Auth0.zc_token_key, None)
+        if not token:
+            log.debug('No Auth0 token found in cookies')
+            return {}
+
         sessionInfo = request.SESSION.get(Auth0.session_key)
         if not sessionInfo:
-            log.debug('No Auth0 session - not directing to Auth0 login')
+            if not token:
```
jplouis: The method returns if not token on line 185, what is this checking?

PR #3194, opened 2018-06-04
ZEN-29964: zenhub: refactor event handling to improve performance

lundybernard:
Send Events directly to the rawEvents queue


Products/ZenHub/interceptors.py
```diff
@@ -0,0 +1,89 @@
...
+from twisted.spread import pb
+from twisted.internet import defer
+
+
+class WorkerInterceptor(pb.Referenceable):
+    """Redirect service requests to one of the worker processes. Note
+    that everything else (like change notifications) go through
+    locally hosted services."""
+
+    callTime = 0.
```
jpeacock-zenoss: This can probably be removed.  Does anything still try to access this variable?

PR #3202, opened 2018-06-06
Fix infrastructure tree navigation

rphillipsz:
https://jira.zenoss.com/browse/ZEN-30176

Use the recursive flag for the tree findChild() call when searching for children.

### DIFF
--- a/Products/ZenUI3/browser/resources/js/zenoss/HierarchyTreePanel.js
+++ b/Products/ZenUI3/browser/resources/js/zenoss/HierarchyTreePanel.js
@@ -444,7 +444,7 @@
         update: function (data) {
             function doUpdate(root, data) {
                 Ext.each(data, function (datum) {
-                    var node = root.findChild('id', datum.id);
+                    var node = root.findChild('id', datum.id, true);
                     if (node) {
                         node.data = datum;
                         node.setText(node.data.text);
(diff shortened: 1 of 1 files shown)


jpeacock-zenoss:
I don't know why setting the `deep` parameter to `true` is necessary.

Products/ZenUI3/browser/resources/js/zenoss/HierarchyTreePanel.js
```diff
@@ -518,7 +518,7 @@
                 if (keys[index] === undefined) {
                     return;
                 }
-                var node = current.findChild(field, keys[index]);
+                var node = current.findChild(field, keys[index], true);
```
jpeacock-zenoss: I'm not sure this will do the right thing in all cases.  The `expandNodeByPath` is already written to traverse down through the children.  Setting `deep` to `true` may cause a node to be returned that's not part of the requested path.

Products/ZenUI3/browser/resources/js/zenoss/HierarchyTreePanel.js
```diff
@@ -556,10 +556,10 @@
...
                         var lastNode = node;
                         if (success && node) {
-                            node = node.findChild(field, last);
-                            if (node) {
-                                me.getSelectionModel().select(node);
-                                Ext.callback(callback, scope || me, [true, node]);
+                            child = node.findChild(field, last, true);
+                            if (child) {
+                                me.getSelectionModel().select(child);
+                                Ext.callback(callback, scope || me, [true, child]);
```
jpeacock-zenoss: Similar issue here.  This depends on `expandNodeByPath` to traverse down a path and then selects the desired child of the returned node.  Doing a deep search may yield a node that wasn't requested.

PR #3203, opened 2018-06-07
ZEN-30104: Select Component Graph, Click All on one Graph, click gear - Link to …

drevak:
…graph. Fails because the URL is too long. Advertises feature that doesnt work.


Products/ZenUI3/browser/resources/js/zenoss/form/graphPanel.js
```diff
@@ -197,6 +197,7 @@
                                          '     </div> ' +
                                          '    <img id="{buttonId}" class="europaGraphGear" src="/++resource++zenui/img/gear.png"  />' +
                                          '</div>'),
+        maxLinkLength: 4096,
```
ssOlexBaiko: please add the appropriate comment to describe the reason of maxLinkLength value

dmytrorevak: added

PR #3212, opened 2018-06-14
ZPS-3796: Fix the issues with cz prefix in one place

rzino:
Fixes the issues with CZ prefix in one place (hopefully)

### Benefits:
- we handle all /cz/ prefix problems in all application and all ZenPacks in one place
- we can remove (I think) all changes we have done all over the product so far to fix the issue with missed prefix (including zenpacks)

### Disadvantage:
- longer "uid" display in grid columns;

I am not sure what other side effects we can get, that is why I would like to show this solution for further discussion.


Products/ZenUI3/browser/resources/js/zenoss/ExtOverrides.js
```diff
@@ -1098,4 +1098,50 @@ Ext.override(Ext.util.Sorter, {
...
+
+                if (uid) {
+                    record.set({
+                        uid: Zenoss.render.link(null, uid)
+                    });
+                }
+            });
+        },
+
+        extractData: function (root) {
```
rzino: Override `extractData` of `Ext.data.reader.Reader`:
Deal with the problem with `/cz/` at the stage when the data only come from the server to any store we have intercept returned results from zope server, and before they are populated to store verify if we have there "uid" field, if so add `/cz/` prefix to it;

Products/ZenUI3/browser/resources/js/zenoss/ExtOverrides.js
```diff
@@ -1098,4 +1098,50 @@ Ext.override(Ext.util.Sorter, {
         }
     });
 
+    Ext.data.Field.override({
+        constructor: function () {
```
rzino: Override constructor of Ext.data.Field for Ext.data.field.Auto:
Write wraper fn for convert functions defined in model definitions. This will be done before data pass through Ext.data.reader.Reader.

- check if data in field is `object` and have `uid`'s if so add /cz/ prefix;
- if there custom `convert` function then call it;

Products/ZenUI3/browser/resources/js/zenoss/ExtOverrides.js
bmoskalyk: Since all data that are used in trees/grids/combos/forms in ExtJs are handled via store functionality it's good to intercept them in store reader and add missing prefix before we show these data in components. This override doesn't break ExtJs source functionality only extend it.

StevePC:
I've made a change to ZPL (ZPS-3801 brnach) that appear to fix all the component grid URLs in vSphere.  I think you should think hard about this change, as it may cause other unforeseen issues.

joshw:
I think StevePC's fix should be put into ZPL regardless.. It looks consistent with current best practices around links and will fix any broken ones in the ZPL-rendered components.   We'll want to QA it against older RM versions, but I think it will be fine.

Roman's change makes me slightly uneasy, but I can't really think of any cases where it would actually cause a problem that didn't need to be fixed anyway.  But I am curious how many bad links would remain after the ZPL fix.. it might not be enough to justify this (slightly risky) fix.

rvykunta:
How about ZenPacks that doesn't use ZPL? How can the URLs get fixed without this change in the platform?

StevePC:
The ZenPacks not using ZPL should either be migrated to ZPL or have their JS updated to prodeuce correct links.  changing the uid field sent to the UI means that every router call we make back to the product will be using modified uids which will not work internally, and won't be translated by any external proxy on the way in.

rzino:
Closing PR based on the above discussion

PR #3215, opened 2018-06-14
ZING-534 Smartview link for component leads to error ZC page.

carlcamera:


carlcamera:
Javascript was creating path which, under certain circumstances was interpreted as a relative path leading to the problems described in the ticket.  Appending `https://` to the path when assigning `window.location` provides proper url interpretation.

PR #3217, opened 2018-06-18
ZEN-30022: Adding migrate script for memcached MAXCONN change

sxanness:


Products/ZenModel/migrate/updateMemcachedMaxConn.py
```diff
@@ -0,0 +1,50 @@
...
+##############################################################################
+
+import logging
+log = logging.getLogger("zen.migrate")
+
+import re
+import Migrate
+from Products.ZenModel.ZMigrateVersion import (
+SCHEMA_MAJOR, SCHEMA_MINOR, SCHEMA_REVISION
+)
```
jpeacock-zenoss: These need to be written as
```
from Products.ZenModel.ZMigrateVersion import SCHEMA_MAJOR, SCHEMA_MINOR, SCHEMA_REVISION
```
All on one line, otherwise, there will be errors when the prodbin component is released.

sxanness: This change was made.

PR #3226, opened 2018-06-20
ZEN-30301 - hierarchy tree selectfirst on load functionality changes

bmoskalyk:


Products/ZenUI3/browser/resources/js/zenoss/HierarchyTreePanel.js
```diff
@@ -406,16 +406,6 @@
             var store =  this.getStore();
             Zenoss.HierarchyTreePanel.superclass.initEvents.call(this);
 
-            if (this.selectRootOnLoad && !Ext.History.getToken()) {
```
bmoskalyk: we should listen store load event instead of rootNode expand to select selectRootOnLoad;

Products/ZenUI3/browser/resources/js/zenoss/HierarchyTreePanel.js
```diff
@@ -666,17 +656,38 @@
...
                 listeners.change = this.filterTree;
             }
             if (this.searchField && this.ownerCt.regSearchListeners) {
                 this.ownerCt.regSearchListeners(listeners);
             }
-            this.getRootNode().expand(false, true, function (node) {
-                node.expandChildNodes();
-            });
+            // check if root node is loaded and expand it;
+            // if not loaded we should listen store load and than expand root node;
```
bmoskalyk: on afterrender we check if tree root node is loaded and than expand and select first node on expand if there are no cached selection, else we listen store load event and than expand root node and select visible root;

Products/ZenUI3/browser/resources/js/zenoss/itinfrastructure.js
```diff
@@ -1142,6 +1142,7 @@ Ext.onReady(function () {
         }),
         router: REMOTE,
         nodeName: 'Device',
```
bmoskalyk: now selectRootOnLoad functionality works as expected and in this case we should prevent selectRootOnLoad in device trees, because trees are rendered one by one and on page init device tree is already rendered and select root node on load but other trees are not rendered yet and this breaks page selection functionality;
instead we select root node in device tree manually when everything is rendered.

Products/ZenUI3/browser/resources/js/zenoss/itinfrastructure.js
```diff
@@ -2073,8 +2074,9 @@ Ext.onReady(function () {
 
     // if there is no history, select the top node
     if (!Ext.History.getToken()) {
-        var node = Ext.getCmp('devices').getRootNode();
-        node.fireEvent('expand', node);
```
bmoskalyk: select root node instead of fire "expand" event, it wasn't working before and didn't select anything;

PR #3227, opened 2018-06-21
ZEN-29542: global name 'log' is not defined

dmytrorevak:
Add the `is_dev_moved variable` in the `_moveDevice` method of `DeviceClass`.
Set the `is_dev_moved` as True at the end of the method because if there are no exceptions above this line of code the device changes its class. So we return the success result regardless of the device was exported or only moved to another class.

Rename `exports` variable to the `moved_dev_amount` on the `devicefacade.py`

Fix undefined log variable in the `facade.py`

Add the appropriate result structure of the `_moveDevices` method of `DeviceFacade`.

### DIFF
--- a/Products/Jobber/facade.py
+++ b/Products/Jobber/facade.py
@@ -77,5 +77,5 @@ class FacadeMethodJob(Job):
             except FacadeMethodJobFailed:
                 raise
             except (TypeError, KeyError):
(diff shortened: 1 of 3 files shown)


webhookbotuser:

webhookbotuser:

webhookbotuser:

webhookbotuser:

PR #3228, opened 2018-06-21
ZEN-29542: global name 'log' is not defined

drevak:
Add the `is_dev_moved variable` in the `_moveDevice` method of `DeviceClass`.
Set the `is_dev_moved` as True at the end of the method because if there are no exceptions above this line of code the device changes its class. So we return the success result regardless of the device was exported or only moved to another class.

Rename `exports` variable to the `moved_dev_amount` on the `devicefacade.py`

Fix undefined log variable in the `facade.py`

Add the appropriate result structure of the `_moveDevices` method of `DeviceFacade`.

### DIFF
--- a/Products/ZenModel/DeviceClass.py
+++ b/Products/ZenModel/DeviceClass.py
@@ -220,7 +220,6 @@ class DeviceClass(DeviceOrganizer, ZenPackable, TemplateContainer):
 
         notify(DeviceClassMovedEvent(dev, dev.deviceClass().primaryAq(), target))
 
-        exported = False
 
         if dev.__class__ != targetClass:
             from Products.ZenRelations.ImportRM import NoLoginImportRM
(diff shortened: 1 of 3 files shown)


Products/ZenModel/DeviceClass.py
```diff
@@ -357,12 +356,12 @@ def moveDevices(self, moveTarget, deviceNames=None, REQUEST=None):
...
         targetClass = target.getPythonDeviceClass()
-        numExports = 0
+        moved_devices_amount = 0
         for devname in deviceNames:
             devicewasExported = self._moveDevice(devname, target, targetClass)
             if devicewasExported:
-                numExports += 1
-        return numExports
+                moved_devices_amount += 1
+        return moved_devices_amount
```
jpeacock-zenoss: Just a nitpick, but please use "count" instead of "amount".  Both words have basically the same meaning, but "count" is the more common (and recognizable) word to use.

drevak: Changed the `moved_devices amount` to the `moved_devices_count` according to your suggestion.

Products/Zuul/facades/devicefacade.py
```diff
@@ -581,8 +581,13 @@ def _moveDevices(self, uids, target):
...
                 notify(ObjectAddedToOrganizerEvent(dev, target))
         elif isinstance(target, DeviceClass):
-            exports = self._dmd.Devices.moveDevices(targetname,[dev.id for dev in devs])
-        return exports
+            moved_devices_count = self._dmd.Devices.moveDevices(targetname,[dev.id for dev in devs])
+
+        result = {
+            'success': bool(moved_devices_count),
+            'message': 'The %s devices have been moved' % moved_devices_count
+        }
```
jpeacock-zenoss: What's the reason for the changing the result from a number to a `dict` object?  As far as I can tell, callers to this function either ignore the result or only care that `bool(result) is True`.

PR #3236, opened 2018-06-27
Change From address parsing in MailProcessor

ed-:


Products/ZenEvents/MailProcessor.py
```diff
@@ -80,6 +80,16 @@ def __init__(self, zem, defaultSeverity = 2):
...
 
+    def parse_email_address(self, address):
+        two_part = re.compile(r'(.*) <(.*?)>$')
+        match = two_part.match(address)
+        if match:
+            return self.parse_email_address(match.group(2))
+        one_part = re.compile(r'''^[<"']?(.*?)[">']?$''')
+        match = one_part.match(address)
+        if match:
+            return match.group(1)
```
jpeacock-zenoss: Since these regular expressions are not dynamic, please perform the compilation once and store the result as a class or module variable.  Probably not a huge performance improvement in this situation, but it is a habit to acquire (e.g. idiom) when using regular expressions.

PR #3239, opened 2018-06-29
Add new impact filter for pagespeed cache

achow-zenoss:


Products/ZenModel/migrate/addZproxyNginxImpactConfig.py
```diff
@@ -0,0 +1,85 @@
...
+                break
+
+        for cfg in zproxy.originalConfigs:
+            if cfg.filename == self.config_file:
+                orig_config = cfg
+                break
+
+        if orig_config:
+            new_content = orig_config.content.replace(org_statement, new_statement)
+            if orig_config != new_content:
```
StevePC: orig_config will never be the same as new_content, do you mean orig_config.content?

Products/ZenModel/migrate/addZproxyNginxImpactConfig.py
```diff
@@ -0,0 +1,85 @@
...
+                orig_config.content = new_content
+                commit = True
+
+        if current_config:
+            new_content = current_config.content.replace(org_statement, new_statement)
+            if current_config != new_content:
```
StevePC: you probably want current_config.content != new_content

Products/ZenModel/migrate/addZproxyNginxImpactConfig.py
```diff
@@ -0,0 +1,85 @@
...
+
+        try:
+            if commit:
+                log.info("Updated zproxy configuration to the latest version.")
+                ctx.commit()
+        finally:
+            if wrote:
+                log.info(("A copy of your existing configuration has been saved to %s."
+                        " If you had made changes to zproxy-nginx.conf in the Control Center UI, "
+                        "please reapply them manually, then restart the service.") % self.save_file)
```
StevePC: the migration only adds 2 lines and removes nothing.  Aside from if the user had removed/modified these two lines and then re-run the migration script, there should be no modifications to re-apply.  You didn't replace the entire contents.  Also the save file path is a container-local path, so not valid outside the container.  you might want to give the path where it will be located on the CC master (eg. /opt/serviced/var/volumes/<tenantid>/...)

Products/ZenModel/migrate/addZproxyNginxImpactConfig.py
achow-zenoss: If the migration already occurred or the orig_config was not of the expected pattern, then new_content will be the same as the orig_config.. In that case, we don't commit

Products/ZenModel/migrate/addZproxyNginxImpactConfig.py
achow-zenoss: Yes, I'll make the info more precise.. 
The host dir zenoss-var-ext is actually mounted into the container as /opt/zenoss/var/ext by Zope service-def.

Products/ZenModel/migrate/addZproxyNginxImpactConfig.py
achow-zenoss: You are right.. Thank you

PR #3244, opened 2018-06-29
ZEN-21935 - Drag-n-drop report organizer into another report organize…

bmoskalyk:
…r throws UI error


ssOlexBaiko:
According to @zedieck message:
Hi All, for now, We should not merge any changes to platform develop branches.  Chris Smouse will let us know when we can continue merging.  For now, finish up what you are working on, PR, but do no merge.  Then please shift over to help with 6.2.1 and 6.3 as you are able. (c)

PR #3251, opened 2018-07-03
ZEN-30379: Fix request url for IP Services

rzino:
We have to add cz prefix to request URL since otherwise, we will get 404 error.


ssOlexBaiko:
lgtm

PR #3262, opened 2018-07-05
ZEN-29646 Re-identify job can fail when there is no perf data, user n…

romanbakaleyko:
…ot informed other than by job failure

### DIFF
--- a/Products/Zuul/facades/metricfacade.py
+++ b/Products/Zuul/facades/metricfacade.py
@@ -772,9 +772,16 @@ class MetricFacade(ZuulFacade):
                         log.info(jsonLine['content'])
                         joblog.info(jsonLine['content'])
                 elif jsonLine['type'] == 'error':
-                    log.error(jsonLine['content'])
-                    joblog.error(jsonLine['content'])
-                    nFails += 1
+                    # ZEN-29646 openTSDB API considers renaming of something what doesn't exist
+                    # as an error with status code 400, handle this as it is not an error for 
+                    # us in this particular case. 
+                    if "does not exist" not in jsonLine['content']:
+                        log.error(jsonLine['content'])
+                        joblog.error(jsonLine['content'])
+                        nFails += 1
+                    else:
+                        log.info("No metrics found for %s", oldId)
+                        joblog.info("No metrics found for %s", oldId)
                 else:
                     log.error(
                         'Unknown log msg type received from central query: '


romanbakaleyko:
In case we query opentsdb to rename somethings what doesn't exist it responses as < HTTP/1.1 400 Bad Request >

(zenoss) [zenoss@bce01496e42c ~]$ curl -v -k https://opentsdb.ip-10-111-23-193/api/uid/rename -H 'Content-Type: application/json' -d '{"tagv": "192.168.56.6", "name":"192.168.56.2"}'
* About to connect() to opentsdb.ip-10-111-23-193 port 443 (#0)
*   Trying 10.111.23.193...
* Connected to opentsdb.ip-10-111-23-193 (10.111.23.193) port 443 (#0)
* Initializing NSS with certpath: sql:/etc/pki/nssdb
* skipping SSL peer certificate verification
* SSL connection using TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
* Server certificate:
*       subject: O=Zenoss,L=Austin,ST=Texas,C=US
*       start date: Jan 18 21:06:04 2016 GMT
*       expire date: Jan 17 21:06:04 2021 GMT
*       common name: (nil)
*       issuer: O=Zenoss,L=Austin,ST=Texas,C=US
> POST /api/uid/rename HTTP/1.1
> User-Agent: curl/7.29.0
> Host: opentsdb.ip-10-111-23-193
> Accept: */*
> Content-Type: application/json
> Content-Length: 47
>
* upload completely sent off: 47 out of 47 bytes
< HTTP/1.1 400 Bad Request
< Content-Type: application/json; charset=UTF-8
< Strict-Transport-Security: max-age=31536000
< Date: Wed, 04 Jul 2018 08:45:09 GMT
< Content-Length: 66
<
* Connection #0 to host opentsdb.ip-10-111-23-193 left intact
{"result":"false","error":"Name(\"192.168.56.6\") does not exist"}

In our particular case it is not a problem for us so I added a handler for this particular case. 

Another thing I have to mention here. Looking to the code base I found weird that when we get an error during renaming of metrics we anyway re-identify a device, however we leave it in the state when collector doesn't collect data for it because we leave it in the "renaming is in progress" state. https://github.com/zenoss/zenoss-prodbin/blob/develop/Products/Zuul/facades/metricfacade.py#L873-L879 

It looks for me wrong, I don't see any reason why would we need to set dev.renameInProgress to False only if there were no errors during metrics re-identifying as we anyway set new id for a device. Looking for a discussion about that, I would change that behavior to changing  dev.renameInProgress to false regardless renaming metrics result unless somebody knows why it is necessary.

StevePC:
I assume that is was left this way so that the reson for the failed moving of metrics could be determined and corrected prior to allowing new data in.  Does the moving of existing data from an old id overwrite or remove data already added to the new id?  If so, then we might want ot ensure that the old data is moved before any new data is added...

Products/Zuul/facades/metricfacade.py
```diff
@@ -772,9 +772,16 @@ def processRename():
...
+                    # ZEN-29646 openTSDB API considers renaming of something what doesn't exist
+                    # as an error with status code 400, handle this as it is not an error for 
+                    # us in this particular case. 
+                    if "does not exist" not in jsonLine['content']:
+                        log.error(jsonLine['content'])
+                        joblog.error(jsonLine['content'])
+                        nFails += 1
+                    else:
+                        log.info("No metrics found for %s", oldId)
+                        joblog.info("No metrics found for %s", oldId)
```
jplouis: I don't think "oldId" is correct in this log. It is possible only certain metrics and or tagv don't have entries.  It is part of the reason that I suggested making the change to look at the response string in central query so that it could send an info or warn message instead of the error.

PR #3267, opened 2018-07-09
minimize number of facts sent to zing-connector. use model catalog to…

fraibacas:
… IndexingEvent hanlders to trigger zing model updates


Products/Zing/model_updates.py
```diff
@@ -0,0 +1,133 @@
...
+            tx_state.need_deletion_fact[uuid] = ZFact.deletion_fact(uuid)
+
+    def update_object(self, obj, idxs=None):
+        """
+        ModelCatalog calls this method when an object needs to be updated
+        """
+        try:
+            self._update_object(obj, idxs)
+        except Exception as ex:
+            log.exception(traceback.format_exc())
```
iancmcc: This will double-log the exception. `log.exception("message")` is the same as `log.error("%s\nmessage", ex)`. Just do the former.

PR #3272, opened 2018-07-10
ZEN-29469: move ip to new realm on device collector change

rderkach:
Do not log the trace when handling exception but log the
exception message and log a traceback only in debug mode.

[JIRA&DEMO](https://jira.zenoss.com/browse/ZEN-29469?focusedCommentId=132013&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-132013)

### DIFF
--- a/Products/Zuul/routers/device.py
+++ b/Products/Zuul/routers/device.py
@@ -14,7 +14,6 @@ Operations for Device Organizers and Devices.
 Available at:  /zport/dmd/device_router
 """
 import logging
-import traceback
 from cgi import escape
 from collections import OrderedDict
 from itertools import islice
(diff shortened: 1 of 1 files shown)


Products/Zuul/routers/device.py
```diff
@@ -948,7 +949,8 @@ def setCollector(self, uids, collector, hashcheck, uid=None, ranges=(),
                 return DirectResponse.succeed('Changed collector to %s for %s devices.' %
                                       (collector, len(uids)))
         except Exception, e:
-            log.exception(e)
+            log.debug(traceback.format_exc())
+            log.error(e)
```
jpeacock-zenoss: For a slightly more efficient (i.e. avoid formatting the traceback which happens regardless of whether it's logged) and keeping the error to a single entry in the log, I suggest:
```
    if log.isEnabledFor(logging.DEBUG):
        log.exception(e)
    else:
        log.error(e)
```
And it also doesn't log a traceback as a DEBUG log entry.

PR #3284, opened 2018-07-16
ZEN-30400: SiteError after following links in Links section on NetApp…

drevak:
… device overview page.

### DIFF
--- a/Products/ZenUI3/browser/resources/js/zenoss/Renderers.js
+++ b/Products/ZenUI3/browser/resources/js/zenoss/Renderers.js
@@ -29,6 +29,10 @@ var ipInterfaceStatusTemplate = new Ext.Template(
     '<span title="Administrative / Operational">{adminStatus} / {operStatus}</span>');
 ipInterfaceStatusTemplate.compile();
 
+// Array with URL prefixes which should be resolved as an external URLs and
+// should not concatenate as a part of /sport/dmd/ path.
+var externalUrlPrefixes = ['http', 'https'];
+
 function convertToUnits(num, divby, unitstr, places){
     unitstr = unitstr || "B";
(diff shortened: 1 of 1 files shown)


Products/ZenUI3/browser/resources/js/zenoss/Renderers.js
```diff
@@ -291,6 +294,11 @@ Ext.apply(Zenoss.render, {
             url = Zenoss.env.CSE_VIRTUAL_ROOT + url.replace(/^\/+/g, '');
         }
         if (url && name) {
+            for (var i=0; i < rootUrlPrefixes.length; i++) {
+                if (url.includes(rootUrlPrefixes[i])) {
+                    url = url.slice(url.indexOf(rootUrlPrefixes[i]))
```
jpeacock-zenoss: I'm not convinced this does anything useful, e.g.
```
>> url = "http://zport/dmd/Devices"
>> url.indexOf("http")
0
>> url.slice(url.indexOf("http"))
"http://zport/dmd/Devices"
>>
```

drevak: The reason why I've added this changes was the format of URL string we received from the server.
In the scope of the ticket's issue, the received URL was "/http://now.netapp.com/". Because of `/` symbol at the beginning of URL, it was rendered as a part of `zport/dmd/` path. For example: `https://zenoss5.ip-10-111-23-161/http://now.netapp.com/` and it caused an error.

With my changes, this issue is resolved.
```
>> url = "/http://now.netapp.com/"
>> url.indexOf("http")
1
>> url.slice(url.indexOf("http"))
"http://now.netapp.com/"
```

PR #3289, opened 2018-07-17
Feature/zen 30384

bmoskalyk:


Products/ZenUI3/browser/resources/js/zenoss/ComponentPanel.js
```diff
@@ -217,16 +217,24 @@ Zenoss.nav.register({
...
+                    id: 'edit_panel',
+                    layout: 'fit'
                 });
-            } else {
-                target.layout.setActiveItem('edit_panel');
-                target.layout.activeItem.setContext(uid);
             }
+            target.layout.setActiveItem('edit_panel');
+            editPanel.setLoading(true);
+            Zenoss.form.getGeneratedForm(uid, function(config){
```
bmoskalyk: we should always add new form when context changes because we receive from server bunch of form fields with data in it, if we will not remove old and add new we will always have same data in details view.

PR #3292, opened 2018-07-19
ZPS-3987: remove cse prefix from object id

rderkach:
Id of the object may come with cz# prefix and we would not
be able to get the object itself.
Remove cz# prefix from the object id

[JIRA&DEMO](https://jira.zenoss.com/browse/ZPS-3987?focusedCommentId=132733&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-132733)

### DIFF
--- a/Products/ZenModel/ZenPacker.py
+++ b/Products/ZenModel/ZenPacker.py
@@ -9,8 +9,10 @@
 
 
 import Globals
+from zope.component import getUtility
 from AccessControl import ClassSecurityInfo
 from Products.ZenWidgets import messaging
+from Products.ZenUtils.virtual_root import IVirtualRoot
 
 class ZenPacker(object):
     security = ClassSecurityInfo()
@@ -49,6 +51,8 @@ class ZenPacker(object):
 
     def findObject(self, id):
         "Ugly hack for inconsistent object structure accross Organizers"
+        # remove cz# prefix if there is one in the object id
+        id = getUtility(IVirtualRoot).strip_virtual_root(id)
         result = []
         try:
             result.append(self._getOb(id))


rderkach:
@joshw this issue is only presented in 7.x line so why it is not backward compatible?

And regarding url. It's not sending an url, the root problem is in 
https://github.com/zenoss/zenoss-prodbin/blob/develop/Products/ZenUI3/browser/templates/devdetail.pt#L20 
where we set up `Zenoss.env.device_uid` which is then used as a parameter for some routers call. 

After rendering that `pt` we get something like
`Zenoss.env.device_uid = /zport/dmd/Devices/vSphere/devices/qa-vcenter1.zenoss.loc`
that is in `html page` then as far as I know some nginx/zproxy config add to that string `cz#` prefix so we get `Zenoss.env.device_uid = /cz0/zport/dmd/Devices/vSphere/devices/qa-vcenter1.zenoss.loc`
on the web page. You can check that in Chrome dev tool.

joshw:
Products.ZenUtils.virtual_root doesn't exist in pre-7.x versions of RM, so that import would fail.

As far as device_uid goes, that seems like a bug to me- if it's being turned into a url by that nginx thing when it's meant to be a uid, stripping it back off again on the server side seems like a silly way to fix it- but if that's what we're doing elsewhere in the product, that's fine with me..  

It just needs to be done in a way that works on RM 6.x  (and it would be best if it works on 5.x and 4.x, just in case we end up backporting this into a maintenance release, or change our plans re: supporting those versions.. unlikely, but I generally try to not go out of my way to break support for those versions, just in case)

rderkach:
Changes are in prodbin so there will be no import fail.

PR #3293, opened 2018-07-19
Feature/zen 30491

bmoskalyk:


Products/ZenUI3/browser/resources/css/zen-cse.css
```diff
@@ -374,6 +373,10 @@ BYE BYE BORDERS
     .z-cse #serviceForm-body textarea
     { background-image: none; background-color: #f4f4f4; border: 0; }
 
+    .z-cse #contentPaneContainer input[type=button] { /*ZEN-30221*/
```
bmoskalyk: we should not clear background-image in input buttons.

Products/ZenUI3/browser/resources/js/zenoss/BackCompatPanel.js
```diff
@@ -108,7 +108,7 @@ Ext.define("Zenoss.IFramePanel", {
     },
     setSrc: function(url) {
         this.frameLoaded = false;
-        if (url === 'about:blank' || url === '') {
+        if (url === 'about:blank' || !url) {
```
bmoskalyk: url can be undefined/null/false/'' - in this case we should load blank source.

PR #3297, opened 2018-07-23
ZEN-30522: Fix usage of input[type=file]

rzino:
Use stylized `input[type=file]` provided by Ext.JS instead of default unstyled.

### DIFF
--- a/Products/ZenUI3/browser/resources/js/zenoss/process.js
+++ b/Products/ZenUI3/browser/resources/js/zenoss/process.js
@@ -864,7 +864,7 @@ Ext.define("Zenoss.SequenceGrid2", {
 });
 
 function readBlob() {
-    var files = document.getElementById("files").files;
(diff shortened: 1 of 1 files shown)


Products/ZenUI3/browser/resources/js/zenoss/process.js
```diff
@@ -1132,25 +1132,36 @@ Ext.define("Zenoss.TestRegexDialog", {
...
+                            {
+                                layout: {
+                                    type: 'hbox',
+                                    pack: 'start',
+                                    align: 'stretch'
+                                },
+                                items: [
+                                    {
+                                        fieldLabel: 'File',
+                                        xtype: 'filefield',
```
rzino: Since Ext.JS has component `Ext.form.field.File` it is better to use it than create some custom file input using `html` property as it was before.

PR #3303, opened 2018-07-25
ZEN-30531: Site Error after navigating to a device with a re-named Ma…

drevak:
…nufacturer entity

### DIFF
--- a/Products/Zuul/facades/manufacturersfacade.py
+++ b/Products/Zuul/facades/manufacturersfacade.py
@@ -249,6 +249,19 @@ class ManufacturersFacade(TreeFacade):
         manufacturer = self._getObject(
                 "/zport/dmd/Manufacturers/" +
                 params.get('oldname', params['name']))
+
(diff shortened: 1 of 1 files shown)


Products/Zuul/facades/manufacturersfacade.py
```diff
@@ -261,6 +274,11 @@ def editManufacturer(self, params):
         if (manufacturer.id != params['name']):
             manufacturer.rename(params['name'])
 
+        # Set updated instances to the product
+        for pair in product_intance_pairs:
+            obj, prod = pair
+            obj.setProductClass(prod)
```
jpeacock-zenoss: Since `product_instance_pairs` contains pairs, this can be written as:
```
for obj, prod in product_instance_pairs:
    obj.setProductClass(prod)
```

drevak: I Implemented the changes you suggested.

PR #3322, opened 2018-08-05
ZEN-29558 Add support of offsets to getDevices method of ModelQuery r…

romanbakaleyko:
…outer.


romanbakaleyko:
also added order and reverse order.

PR #3353, opened 2018-08-28
ZEN-30684 CZ RBAC does not allow ZenManagers to define Administered D…

romanbakaleyko:
…evices for ZenUsers


romanbakaleyko:
It reverts back an ability to edit users for ZenManagers what was changed by https://github.com/zenoss/zenoss-prodbin/pull/3135

I am not sure it is correct to set version Migrate.Version(300, 0, 0) same like the migration in the mentioned commit pull/3135

I confirm that ZenManager has an access to edit fields which are described here https://docs.google.com/spreadsheets/d/1zjHuNsnW9O3z3KzOaS9eKN-CB0GuCCUZScRb0OhHWEA/edit#gid=0

Products/ZenModel/migrate/UpdateEditUserPermission.py
```diff
@@ -0,0 +1,52 @@
...
+    MANAGER_ROLE,
+    CZ_ADMIN_ROLE,
+    OWNER_ROLE,
+    ZEN_EDIT_USER
+)
+
+
+class UpdateEditUserPermission(Migrate.Step):
+    
+    version = Migrate.Version(300, 0, 0)
```
jpeacock-zenoss: I'm thinking the migrate version should be higher than the version given in #3135.
The ordering of migrate scripts having the same version is undefined.

Products/ZenModel/UserSettings.py
```diff
@@ -721,9 +721,10 @@ def iseditable(self):
         if currentUser == thisUser:
             return True
 
-        # CZAdmins can edit any users' settings except for Managers.
-        if (currentUser.has_role(CZ_ADMIN_ROLE) and not
-                thisUser.has_role(MANAGER_ROLE)):
+        # CZAdmins and ZEN_MANAGERs can edit any users' settings except for Managers.
+        if (currentUser.has_role(CZ_ADMIN_ROLE) or \
```
iancmcc: This isn't true, I think. ZenManagers can't edit CZAdmins.

PR #3354, opened 2018-08-29
add tests for BasePublisher class

rderkach:


Products/ZenHub/tests/testMetricPublisher.py
```diff
@@ -8,22 +8,75 @@
 ##############################################################################
 
 import unittest
+from mock import Mock, create_autospec, patch, call
 
 import ujson as json
 
-from Products.ZenHub.metricpublisher.publisher import RedisListPublisher, HttpPostPublisher, BasePublisher
+from Products.ZenHub.metricpublisher.publisher import RedisListPublisher, HttpPostPublisher, BasePublisher, deque, bufferHighWater, defer
```
lundybernard: PEP8: breakup long imports using parentheses:

```python
from Products.ZenHub.metricpublisher.publisher import (
    RedisListPublisher,
    HttpPostPublisher,
    BasePublisher,
    deque,
    bufferHighWater,
    defer,
)
```

Products/ZenHub/metricpublisher/tests/test_publisher.py
```diff
@@ -0,0 +1,441 @@
...
+# This content is made available according to terms specified in
+# License.zenoss under the directory where your Zenoss product is installed.
+#
+##############################################################################
+
+from unittest import TestCase, TestSuite, makeSuite
+from mock import Mock, MagicMock, create_autospec, patch
+from zope.interface.verify import verifyObject
+
+import ujson as json
```
lundybernard: The module under test imports already imports json, so it should be imported from Products.ZenHub.metricpublisher.publisher

Products/ZenHub/metricpublisher/tests/test_publisher.py
```diff
@@ -0,0 +1,441 @@
...
+
+METRIC_BUILDED = [
+    '{"timestamp":1535460634.26479,"metric":"eventQueueLength",\
+    "value":0.0,"tags":{\
+        "instance":"0","daemon":"zenpython","monitor":"localhost",\
+        "metricType":"GAUGE","tenantId":"b0e4t72hrole5z1xv88djc1hb"\
+        }\
+    }'
+]
+METRIC = 'testMetricName'
```
lundybernard: BUFFER_LEN is only used in BasePublisherTest, and should be defined within its scope until it is needed by another class.

Products/ZenHub/metricpublisher/tests/test_publisher.py
```diff
@@ -0,0 +1,441 @@
...
+BUFFER_LEN = 10
+PUBLISHER_FREQ = 10
+SCHEDULED = True
+
+
+class PublishError(Exception):
```
lundybernard: Any time we need to define a class, function, or object just for testing, it is a good candidate for a Mock.

for example, we can replace this class with
```
PublishError = Mock(Exception, name='PublishError', autospec=True)
```

This object is only used in a single test, `BasePublisherTest.test_publish_failed` and should be moved into that test.

Products/ZenHub/metricpublisher/tests/test_publisher.py
```diff
@@ -0,0 +1,441 @@
...
+            self.metric,
+            self.pub.build_metric('m', 1.0, 1, {}))
+        self.assertEqual(
+            self.metric,
+            self.pub.build_metric('m', "1.0", 1, {})
+        )
+
+    def test_publish_failed(self):
+        pub_error = PublishError()
+        self.pub._publish_failed(pub_error, METRIC_BUILDED)
```
lundybernard: _publish_failed returns a value, which should be captured and checked.
ex:
```
ret = self.pub._publish_failed(pub_error, METRIC_BUILDED)
self.assertEqual(ret, 'some value')
```

Products/ZenHub/metricpublisher/tests/test_publisher.py
```diff
@@ -0,0 +1,441 @@
...
+        super(PublishError, self).__init__("Some error in publishing")
+
+
+class BasePublisherTest(TestCase):
+
+    def setUp(self):
+        self.pub = BasePublisher(BUFFER_LEN, PUBLISHER_FREQ)
+        self.metric = {'metric': 'm', 'value': 1.0, 'timestamp': 1, 'tags': {}}
+
+    def test_init(self):
```
lundybernard: Naming convention.  change this to `test___init__`, and tests for functions with a leading _ should be `test__function_name`.  this prevents issues if we add a `def init():` method, and where we have `def function_name()` and a `def _function_name()` which is a fairly common pattern.

Products/ZenHub/metricpublisher/tests/test_publisher.py
rderkach: I also use BUFFER_LEN  in RedisPublisherTest.test_init

PR #3360, opened 2018-09-04
ZEN-29856 Use pagination to run zencatalogscan

ssOlexBaiko:
[Jira](https://jira.zenoss.com/browse/ZEN-29856)


Products/Zuul/catalog/model_catalog_tool.py
```diff
@@ -270,6 +271,28 @@ def getValue(obj):
         return sorted((unbrain(brain) for brain in queryResults),
                       key=getValue, reverse=reverse, cmp=natural_compare)
 
+    def get_all_docs(self):
```
jpeacock-zenoss: There are some extra requirements needed here.  It needs to accept the `types`, `limit`, and `fields` parameters.  I plan to use this in the ZenETL zenpack, but I need this API to support these three parameters.

Also, with the addition of those parameters, `search_with_cursor` or `cursor_search` will probably be a better name for the method.

Products/Zuul/catalog/model_catalog_tool.py
```diff
@@ -270,6 +271,28 @@ def getValue(obj):
...
+        need_result = True
+        while need_result:
+            search_params = SearchParams(query, limit=limit, order_by=order_by, next_cursor=cursor)
+            catalog_results = self.model_catalog_client.cursor_search(search_params, self.context)
+            if cursor == catalog_results.next_cursor:
+                need_result = False
+                break
+            cursor = catalog_results.next_cursor
+            for result in catalog_results.results:
+                yield result
```
jpeacock-zenoss: This looks _really_ similar to the `_wrap_results_gen` method in modelindex's `SolrSearcher` class.  Perhaps `SolrSearcher` should have the generator method instead.

I do not think the cursors should be treated identically to pagination.  Conceptually, cursors are very much like iterators and generators.  I think all the implementation detail of setting `cursorMark` and copying the cursor mark from the response into the next query should be kept in the `modelindex` package.  As far as `ModelCatalogTool` is concerned, it's receiving a iterator (or generator) that it has to iterate over to produce the brains.

Products/Zuul/catalog/model_catalog_tool.py
```diff
@@ -270,6 +271,17 @@ def getValue(obj):
...
 
+    def cursor_search(self, types=(), limit=DEFAULT_SEARCH_LIMIT, filterPermissions=False, fields=None):
+        """
+        Get all results from catalog by using solr deep pagination.
+        return generator.
+        """
+        init_cursor = INITIAL_CURSOR_MARK
+        order_by = INDEX_UNIQUE_FIELD
+        query, _ = self._build_query(filterPermissions=filterPermissions, types=types)
+        search_params = SearchParams(query, limit=limit, order_by=order_by, next_cursor=init_cursor, fields=fields)
```
jpeacock-zenoss: I think we can further constrain (or simplify) the API by having modelindex provide a `CursorSearchParams` (having `SearchParams` as a base class) object that fills in the `order_by` and `next_cursor` fields with the correct values.  This will complete the symmetry of the search methods and strongly indicate that they are different.  The `search` method uses `SearchParams` and the `cursor_search` method uses `CursorSearchParams`.

Products/Zuul/catalog/model_catalog.py
```diff
@@ -433,6 +436,23 @@ def _get_fields_to_return(self, fields):
...
+        try:
+            search_params.fields = self._get_fields_to_return(search_params.fields)
+            catalog_results = self.model_index.cursor_search(search_params)
+        except SearchException as e:
+            log.error("EXCEPTION: {0}".format(e.message))
+            self.raise_model_catalog_error("Exception performing search")
+        else:
+            for result in catalog_results:
+                brains = iter(self._parse_catalog_results(result, context))
+                count = result.total_count
```
jpeacock-zenoss: Question: is `total_count` the count of items in the specific result or the total count of all the items in the catalog?

E.g. If there are 1000 items in the catalog and I request a cursor search with a limit of 10 items.  Does `total_count` have the value 1000 or 10?

jpeacock-zenoss: OK, good.  I think we can now take the next step to make the `search` and `cursor_search` methods achieve parity.  

Currently, the `search` method returns an object that is implements the iterator protocol *and* provides the metadata fields like `total_count`.  The `cursor_search` method returns a generator.  I think we can return the same kind of object from `cursor_search` as is returned from `search`.

You won't be able to use the `SearchResults` class as is.  A `CursorSearchResults` class is needed to handle the nested iteration required to produce the catalog results.  Here's an example of the functionality I want to see from `ModelCatalogTool`'s `cursor_search` method:
```
tool = IModelCatalogTool(dmd)
query = "Products.ZenModel.Device.Device"

results = tool.cursor_search(query, limit=100)
log.info("Found %s devices in the catalog", results.count)
for brain in results:
    log.info("Found device %s", brain.getPath())
```
I can already do this using the `search` method.  Let's make `cursor_search` support the same functionality.

ssOlexBaiko:
retest this

ssOlexBaiko:
retest this

ssOlexBaiko:
Retest this

ssOlexBaiko:
retest this

PR #3373, opened 2018-09-11
ZEN-30738 - CZ: Expanding component graph puts close button off screen

bmoskalyk:


Products/ZenUI3/browser/resources/js/zenoss/form/graphPanel.js
```diff
@@ -487,20 +487,27 @@
             if (graphPanel && Ext.isNumber(graphPanel.drange)) {
                 drange = graphPanel.drange;
             }
+            // ZEN-30758
```
bmoskalyk: cross browser screen size.

PR #3384, opened 2018-09-19
ZEN-27680: Don't allow min value of process threshold to be greater than max value

rzino:


rzino:
Closing PR [according to comment in Jira](https://jira.zenoss.com/browse/ZEN-27680?focusedCommentId=136721&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-136721)

PR #3389, opened 2018-09-21
ZEN-30684 CZ RBAC does not allow ZenManagers to define Administered D…

romanbakaleyko:
…evices for ZenUsers.


romanbakaleyko:
after discussion with Ian it appeared that current ticket is invalid, user management going to be implemented in the cloud side, reverting changes https://github.com/zenoss/zenoss-prodbin/pull/3353/files

PR #3398, opened 2018-09-25
ZEN-30795: Include SNMP trap source IP address in event detail

drevak:


drevak:
fix the title and the commit message

PR #3415, opened 2018-10-02
ZEN-30344

ssOlexBaiko:
[Jira](https://jira.zenoss.com/browse/ZEN-30344)
port from https://github.com/zenoss/zenoss-prodbin/pull/3342


dougsyer:
Id like to say i never release a modeler that fails to compile but my test coverage fails me sometimes and if i need to tshoot with non administrative access thats another thing that would help us get there.

PR #3416, opened 2018-10-02
Move zenhubworker into its own service.

jpeacock-zenoss:
Implements ZEN-30832.

A new service, zenhubworker, is added to each collection hub.  This new service replaces the zenhubworker subprocess that zenhub previously used. zenhub no longer spawns zenhubworker subprocesses.

The zenhubworker service is organized as a sibling to the zenhub service primarily due to aesthetics.  If zenhubworker service were a child of zenhub, zenhub's healthcheck information would no longer be visible from the service tree display in Control Center.

After service migration:
* There will be one instance of zenhub.
* The number of zenhubworker instances will equal the original number of zenhub instances multiplied by the 'workers' count found in zenhub's pre-migrate config file.
* The 'invalidationworkers' option in zenhub's config file will be set to the original number of zenhub instances multiplied by the original 'invalidationworkers' count found in zenhub's config file.


Products/ZenModel/migrate/addZenHubWorkerService.py
```diff
@@ -0,0 +1,222 @@
...
+        content = configFile.content
+
+        content, _ = editConfig(workers_entry, content)
+        content, _ = editConfig(worker_priority_entry, content)
+        content, _ = editConfig(call_limit_entry, content)
+        content, _ = editConfig(priority_set, content)
+
+        # Retrieve workers config, if it was set
+        content, groups = editConfig(workers_set, content)
+        workers = groups[0] if groups else None
```
jplouis: Should the new worker count be the number of zenhub instances * worker

jpeacock-zenoss: I chatted with Jake and StevePC and they said yes.  They also recommended that the invalidationworker count should also be adjusted if there were multiple `zenhub` instances.  I added adjusting the invalidationworker count to the migration.

PR #3428, opened 2018-10-11
ZEN-30841: Relationship Between Network and Device Not Removed When D…

olst:
…evice is Deleted


olst:
We need to restore `IpAddress` objects on interfaces, not on nets.
Also, generator () doesn't work for `iface_ips_bkup ` var.

PR #3429, opened 2018-10-11
ZEN-30880: make a request to server to get Browser state

rderkach:
There is nginx filtering rules that are adding /cz# prefix
to all the html that is passed to UI and this is breaking
things that do not require /cz# prefix.
To fetch the browser state we are now making an api call.

[JIRA&DEMO
](https://jira.zenoss.com/browse/ZEN-30880?focusedCommentId=138066&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-138066)

### DIFF
--- a/Products/ZenUI3/browser/javascript.py
+++ b/Products/ZenUI3/browser/javascript.py
@@ -271,15 +271,16 @@ class BrowserState(JavaScriptSnippet):
     """
     Restores the browser state.
     """
+    # return the js snippet that will make a request for browser state
+    # and than set it
     def snippet(self):
-        try:
-            userSettings = self.context.ZenUsers.getUserSettings()
-        except AttributeError:
-            # We're on a backcompat page where we don't have browser state
-            # anyway. Move on.
-            return ''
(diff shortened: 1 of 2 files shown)


Products/ZenUI3/browser/javascript.py
```diff
@@ -271,15 +271,16 @@ class BrowserState(JavaScriptSnippet):
...
-            # We're on a backcompat page where we don't have browser state
-            # anyway. Move on.
-            return ''
-        state_container = getattr(userSettings, '_browser_state', {})
-        if isinstance(state_container, basestring):
-            state_container = {}
-        state = state_container.get('state', '{}')
-        return 'Ext.state.Manager.getProvider().setState(%r);' % state
+        snippet = """
+            Ext.onReady(function(){
```
cluther: Do we want to ignore failures here? Or is there something more appropriate that could be done?

rderkach:
@cluther There is no guarantee `setState(response)` will be executed before all the other that are using `state` and the problem, as far as I can tell, is not is `Ext.onReady()` more in requests that are asynchronous.

PR #3432, opened 2018-10-12
ZEN-29654 Reopened- zenmail fails to parse hostname from an email address (the patch for this broke zenmail)

romanbakaleyko:
Change From address parsing in MailProcessor


romanbakaleyko:
originally change went to the develop, cherry-picking it to support.

PR #3433, opened 2018-10-15
 ZEN-30673 Zenmanager role is missing options from the Advanced tab

romanbakaleyko:


romanbakaleyko:
updated zodbsql to have changes in the brand new deployment, also fixed copy/paste typo in the migration script.

PR #3434, opened 2018-10-15
RRDDataPoint: add store, forward, tags properties

cluther:
Add support for all datapoints to have three new capabilities related to
metric tagging.

- store
- forward
- tags

The store property will result in the {"no-store": "true"} tag being
added to published metrics for the datapoint when set to False, and no
additional tags set when set to True. True is the default to retain
backwards compatibility for all existing datapoints, and all new
datapoints by default.

The forward property will result in the {"no-forward": "true"} tag being
added to published metrics for the datapoint when set to False, and no
additional tags set when set to True. True is the default to retain
backwards compatibility for all existing datapoints, and all new
datapoints by default.

The tags property allows arbitrary additional tags to be published for
the datapoint's metrics. Multiple tags can be specified in the following
format (one per line):

    name1:value2
    name2:value2

These tags will be evaluated as TALES expressions so examples like the
following will also work.

    deviceId:${device/id}
    hereId:${here/id}
    interfaceName:${here/interfaceName}
    totalMemory:${here/hw/totalMemory}

These changes only immediately add this support for metrics collected by
"platform" (non-ZenPack) collector services. Specifically the following
collectors.

- zencommand
- zenperfsnmp
- zenping
- zenprocess

All collector services that are part of ZenPacks must be updated to use
this new capability individually. Specifically by updating their
HubService to pass the per-datapoint tags down to the collector service,
and their collector services' calls to writeMetricWithMetadata to
include those tags as the extraTags parameter. See the
CommandPerformanceConfig.py and zencommand.py changes as a reference for
the kind of updates that are required.

Implements ZEN-30906.


StevePC:
What are the performance implications of adding additional per-datapoint tales expressions to the config generation?

cluther:
> What are the performance implications of adding additional per-datapoint tales expressions to the config generation?

The minimal you'd expect. If you take a look at the RRDDatapoint.getTags() method that gets called for each datapoint during config generation for datapoint collector daemons (zenping, zenprocess, zenperfsnmp, zencommand, +ZenPacks in the future) you will see that there are many "early-exit" scenarios that prevent any TALES evaluation overhead at all. Under the default datapoint configuration with no extra tags specified there will be no TALES evaluation, or even loading of the context necessary for TALES evaluation.

So there will only be significant overhead if people use TALES expressions in the datapoint extra tags.

Products/ZenCollector/daemon.py
```diff
@@ -468,6 +482,9 @@ def writeMetric(self, contextKey, metric, value, metricType, contextId,
 
         # check for threshold breaches and send events when needed
         if value is not None:
+            if extraTags:
+                tags.update(extraTags)
```
jplouis: Do we want the extraTags to possibly overwrite the default tags?

cluther: Yes. That was intentional.

Products/ZenHub/services/ProcessConfig.py
```diff
@@ -73,29 +74,36 @@ def _filterDevice(self, device):
...
+
             # Find out which datasources are responsible for this process
             # if SNMP is not responsible, then do not add it to the list
             snmpMonitored = False
             for rrdTpl in p.getRRDTemplates():
-                if len(rrdTpl.getRRDDataSources("SNMP")) > 0:
+                for datasource in rrdTpl.getRRDDataSources("SNMP"):
                     snmpMonitored = True
+                    for datapoint in datasource.getRRDDataPoints():
+                        tags = datapoint.getTags(p)
```
jplouis: This loop will only get the tags from first datapoint for the last datasource since the break only exits the inner loop. Is that the desired behavior? Maybe a comment on what this is doing.

cluther: I'll add a comment. This is the case in PingTask too. They both essentially ignore the datapoint configurations and assume a static list. So I didn't go to the trouble of changing the overall strategy of how their configs are built.

Products/ZenModel/RRDDataPoint.py
```diff
@@ -199,6 +216,95 @@ def isCounter(self):
...
+
+        This includes tags that are the result of setting the store, forward,
+        and other properties that should influence the metric tags for this
+        datapoint.
+
+        """
+        tags = {}
+
+        if self.store is False:
+            tags[NO_STORE_KEY] = TRUTH_STRING
```
jplouis: Just to be clear, the metric-consumer impl just looks for the presence of the tag key, the value is never looked at.  It will work with this, but won't work if someone adds `no-store=False`

cluther: Yes, I agree that entering "no-store:False" or something like that into the extra tags will result in it being stored, but I don't think that's likely to happen since there's a checkbox right about it saying "Store in CZ" that handles that without doing extra tags at all.

PR #3440, opened 2018-10-16
Fix typo in WorkerInterceptor.

jpeacock-zenoss:
Also fix serialization problem when sending data to sendEvent/sendEvents.

Fixes ZEN-30920.

### DIFF
--- a/Products/ZenHub/tests/test_interceptors.py
+++ b/Products/ZenHub/tests/test_interceptors.py
@@ -8,7 +8,7 @@
 ##############################################################################
 
 from unittest import TestCase
-from mock import Mock, MagicMock, patch
+from mock import Mock, MagicMock, call, patch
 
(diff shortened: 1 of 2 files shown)


Products/ZenHub/tests/test_interceptors.py
```diff
@@ -47,9 +60,18 @@ def setUp(self):
         self.zenhub.deferToWorker.return_value = 'deferred'
         self.wi = WorkerInterceptor(zenhub=self.zenhub, service=self.service)
         self.broker = Mock(name='pb.Broker', spec_set=pb.Broker)
+        self.broker.unserialize = Mock()
```
lundybernard: this will work better as:
and unserialize will be included in the spec
```suggestion
        self.broker = Mock(pb.Broker, name='pb.Broker')
```

PR #3448, opened 2018-10-22
Updated logging for applyDataMaps user priority.

jpeacock-zenoss:
* The worker stats logging includes better information for the new higher priority jobs from user requested device modeling.
* Model pause checking has been abstracted into its own class
* Job prioritization is now handled by its own class.

Fixes ZEN-30892.

### DIFF
--- a/Products/ZenHub/tests/test_zenhub.py
+++ b/Products/ZenHub/tests/test_zenhub.py
@@ -36,6 +36,7 @@ from Products.ZenHub.zenhub import (
     DefaultConfProvider, IHubConfProvider,
     DefaultHubHeartBeatCheck, IHubHeartBeatCheck,
     IEventPublisher,
+    ModelingPaused
 )
 
 PATH = {'src': 'Products.ZenHub.zenhub'}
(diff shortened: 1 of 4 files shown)


Products/ZenHub/tests/test_zenhub.py
```diff
@@ -1235,3 +1236,28 @@ def test_check(t):
...
+
+    def test_paused(self):
+        dmd = Mock()
+        dmd.getPauseADMLife.return_value = 100
+        pause_timeout = 200
+        paused = ModelingPaused(dmd, pause_timeout)
+
+        actual = paused()
+
+        dmd.getPauseADMLife.assert_has_calls([call()])
```
lundybernard: more concise way to say 'called with no parameters'
```suggestion
        dmd.getPauseADMLife.assert_called_with()
```

Products/ZenHub/tests/test_zenhub.py
```diff
@@ -1235,3 +1236,28 @@ def test_check(t):
...
+
+    def test_not_paused(self):
+        dmd = Mock()
+        dmd.getPauseADMLife.return_value = 300
+        pause_timeout = 200
+        paused = ModelingPaused(dmd, pause_timeout)
+
+        actual = paused()
+
+        dmd.getPauseADMLife.assert_has_calls([call()])
```
lundybernard:
```suggestion
        dmd.getPauseADMLife.assert_called_with()
```

Products/ZenHub/tests/test_zenhub.py
jpeacock-zenoss: Hmm... I was hoping to simulate `assert_called_once` (not available in the version of mock we're using), but neither `assert_has_calls([call()])` nor `assert_called_with()` appear to do that.

PR #3450, opened 2018-10-23
ZEN-30704 User with ZenUser role can't manage events where it should

romanbakaleyko:


romanbakaleyko:
It appeared that for global role ZenUser should be an ability to add admin objects with rights to manage events. [DEMO](https://jira.zenoss.com/secure/attachment/56483/demo-ZEN-30704.mov)

PR #3451, opened 2018-10-23
ZEN-30840: Add InvalidationManager class

lundybernard:
* move zenhub invalidation functionality into InvalidationManager
* Fix failing legacy integration tests

### DIFF
--- a/Products/ZenHub/invalidationmanager.py
+++ b/Products/ZenHub/invalidationmanager.py
@@ -1,0 +1,269 @@
+##############################################################################
+#
+# Copyright (C) Zenoss, Inc. 2018, all rights reserved.
+#
+# This content is made available according to terms specified in
+# License.zenoss under the directory where your Zenoss product is installed.
+#
+##############################################################################
+
+import logging
+from time import time
+from itertools import chain
+
+from twisted.internet.defer import inlineCallbacks
+from zope.component import getUtility, getUtilitiesFor, subscribers
+from ZODB.POSException import POSKeyError
+
+from zenoss.protocols.protobufs.zep_pb2 import (
+    SEVERITY_CRITICAL, SEVERITY_CLEAR
+)
+
+import Globals  # required to import zenoss Products
+from Products.ZenUtils.Utils import unused
+
+from Products.ZenRelations.PrimaryPathObjectManager import (
+    PrimaryPathObjectManager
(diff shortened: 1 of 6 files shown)


lundybernard:
jenkins retest this please

Products/ZenHub/invalidationmanager.py
```diff
@@ -0,0 +1,281 @@
...
+
+
+@coroutine
+def iterate(target):
+    """iterates over the iterable, sending each produced item to the target.
+    """
+    while True:
+        iterable = (yield)
+        for item in iterable:
+            target.send(item)
```
jpeacock-zenoss: The `iterate` coroutine isn't used and so can be removed.  It's essentially dead code.

Products/ZenHub/invalidationmanager.py
```diff
@@ -0,0 +1,270 @@
...
+        except Exception:
+            log.exception('error in _poll_invalidations')
+
+    @inlineCallbacks
+    def _send_event(self, event):
+        yield self.__send_event(event)
+
+
+class InvalidationPipeline(object):
+
```
jpeacock-zenoss: I think this class could use a docstring to explain what it does.

Products/ZenHub/invalidationmanager.py
```diff
@@ -0,0 +1,270 @@
...
+            filter_obj(
+                filters,
+                transform_obj(
+                    processor
+                )
+            )
+        )
+
+    def run(self, invalidations):
+        self.__pipeline.send(invalidations)
```
jpeacock-zenoss: The parameter should be written in the singular form rather than plural.  I read 'invalidation**s**' and think this is being passed an iterable rather than a single item.

Products/ZenHub/invalidationmanager.py
lundybernard: Added doc string

Products/ZenHub/invalidationmanager.py
lundybernard: Changed these to singular 'invalidation'

Products/ZenHub/invalidationmanager.py
lundybernard: it has been removed

PR #3459, opened 2018-10-26
Add object references to the serialized model change event

achow-zenoss:


Products/ZenMessaging/ChangeEvents/events.py
```diff
@@ -62,8 +62,9 @@ class MessagePrePublishingEvent(object):
     """
     implements(IMessagePrePublishingEvent)
 
-    def __init__(self, msgs, maintWindowChanges):
+    def __init__(self, msgs, maintWindowChanges, refs={}):
```
fraibacas: using a dict as default param might have undesirable side effects. Default it to None and create the dict in the constructor if refs is None.
https://stackoverflow.com/questions/26320899/why-is-the-empty-dictionary-a-dangerous-default-value-in-python

PR #3467, opened 2018-11-01
Feature/zen 30977

achow-zenoss:
The references to the zing related code is not included..


Products/ZenMessaging/ChangeEvents/events.py
```diff
@@ -63,6 +63,8 @@ class MessagePrePublishingEvent(object):
 
     def __init__(self, msgs, maintWindowChanges):
         self.msgs = msgs
+        self.refs = refs
```
fraibacas: refs does not seem to be a parameter in __init__

achow-zenoss: Thank you!.. updated

PR #3470, opened 2018-11-05
ZEN-24368 Add component counts to callhome data

romanbakaleyko:


Products/ZenCallHome/ZenossAppData.py
```diff
@@ -45,9 +46,19 @@ def callHomeData(self, dmd):
...
                  self.user_count,
                  self.product_count,
-                 self.product_name)
+                 self.product_name,
+                 self.components)
         return chain.from_iterable(map(lambda fn: fn(), stats))
 
+    def components(self):
+        brains = self._catalog.search(types=(DeviceComponent,), facets_for_field=["meta_type"])
+        facets = brains.facets.itervalues().next()
```
fraibacas: protect this in case we get an unexpected response from solr
if brains.facets and brains.facets.get("meta_type"):
    facets = brains.facets["meta_type"]

Products/ZenCallHome/ZenossAppData.py
```diff
@@ -45,9 +46,19 @@ def callHomeData(self, dmd):
...
+        comps = facets.get_values()
+        for comp, count in comps.items():
```
fraibacas: use iteritems

PR #3471, opened 2018-11-05
ZEN-30929: Add "Model Change Handler time" measurement to RM ZenPack

drevak:
Added the Metrology timers to check the execution of commit hooks.

### DIFF
--- a/Products/ZenMessaging/queuemessaging/publisher.py
+++ b/Products/ZenMessaging/queuemessaging/publisher.py
@@ -296,6 +296,7 @@ class PublishSynchronizer(object):
             log.debug("ModelEventList batch size %s" % count)
         return msgs
 
+    @Metrology.utilization_timer('zen.queuepublisher.beforeCompletionHookTimer')
     def beforeCompletionHook(self, tx):
         try:
             log.debug("beforeCompletionHook on tx %s" % tx)
@@ -316,6 +317,7 @@ class PublishSynchronizer(object):
             if hasattr(tx, '_synchronizedPublisher'):
                 tx._synchronizedPublisher = None
 
+    @Metrology.utilization_timer('zen.queuepublisher.afterCompletionHookTimer')
     def afterCompletionHook(self, status, tx):
         try:
             log.debug("afterCompletionHook status:%s for tx %s" % (status, tx))


drevak:
retest this

Products/ZenMessaging/queuemessaging/publisher.py
```diff
@@ -31,6 +31,8 @@
 log = logging.getLogger('zen.queuepublisher')
 
 MODEL_TYPE = ProtobufEnum(modelevents_pb2.ModelEvent, 'model_type')
+BEFORE_COMMIT_HOOK_TIMER = Metrology.utilization_timer("zen.queuepublisher.beforeCompletionHookTimer")
```
cluther: I don't think these meters should be created in the module's global scope. This leads to them being created in process that don't even have access to ZODB. See what happens when I put a pdb.set_trace() on the same line, then run zenpython.

```
(zenoss) [zenoss@5b5f49ae7072 ~]$ zenpython run
> /opt/zenoss/Products/ZenMessaging/queuemessaging/publisher.py(40)<module>()
-> class ModelChangePublisher(object):
(Pdb)
```

If you want them in the module's scope, you could set them to None here, then initialize them on demand when they're used.

Products/ZenMessaging/queuemessaging/publisher.py
```diff
@@ -232,8 +234,10 @@ def getModelChangePublisher():
         tx._synchronizedPublisher = ModelChangePublisher()
         # create new PublishSynchronizer also add after completion hook so that client/channel can be closed
         pSync = PublishSynchronizer()
-        tx.addBeforeCommitHook(pSync.beforeCompletionHook, [tx])
-        tx.addAfterCommitHook(pSync.afterCompletionHook, [tx])
+        with BEFORE_COMMIT_HOOK_TIMER:
+            tx.addBeforeCommitHook(pSync.beforeCompletionHook, [tx])
```
cluther: I believe this only causes us to measure how long it takes to add the hook. Not how long the hooks take to execute on each commit.

Products/ZenMessaging/queuemessaging/publisher.py
drevak: Added the Metrology timers as a decorator to the appropriate methods.
The Metrology package has his own registry package so we only can use it without writing additional getter function.

Products/ZenMessaging/queuemessaging/publisher.py
drevak: Added the timers as a decorator to `beforeCompletionHook` and `afterCompletionHook` functions.

PR #3478, opened 2018-11-12
ZEN-31009 Moving devices to/from Groups, Locations and Systems result…

romanbakaleyko:
… in 'Remodel Required' modal


romanbakaleyko:
[https://jira.zenoss.com/secure/attachment/57128/demo.mov](DEMO)

PR #3481, opened 2018-11-14
ZEN-30592: remove buildTextLabel functionality

rderkach:
When a user uses page size filter on the table view and sorts
the column that has int value in it will fall with a traceback.
Also, values from the sorted column are used as a page name on
pagination select element, which is not logically right, there
should be only the page numbers 1,2,3 ....
Remove that functionality since it is confusing and not logical.

We are using `_buildTextLabel`  only in this module.

[JIRA&DEMO](https://jira.zenoss.com/browse/ZEN-30592?focusedCommentId=140258&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-140258)


rderkach:
@jpeacock-zenoss Ok thanks for explaining this.  I will fix that with `str` approach.

PR #3484, opened 2018-11-26
Feature/ZING-1627

corntoole:
ZING-1423/ZING-1627: CZ should block access of users without any roles on that CZ by redirecting them to Zenoss Cloud UI.

### DIFF
--- a/Products/ZenUtils/Auth0/Auth0.py
+++ b/Products/ZenUtils/Auth0/Auth0.py
@@ -27,6 +27,7 @@ import jwt
 import logging
 import re
 import time
+import traceback
 
 log = logging.getLogger('Auth0')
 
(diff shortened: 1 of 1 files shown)


Products/ZenUtils/Auth0/Auth0.py
```diff
@@ -337,6 +341,10 @@ def challenge(self, request, response):
         # sessionInfo doesn't persist the auth_expiration property when modified outside of the initial storage,
         # and a local dictionary isn't available to all zopes.
         if sessionInfo:
+            if len(sessionInfo.roles) < 1:
+                request['RESPONSE'].redirect('/?errcode=1', lock=1)
```
jplouis: Can you put a comment here on how it interacts with the zing UI and the meaning of the error code number.

PR #3485, opened 2018-11-27
ZEN-31092 [RCE] Remote Code Execution with weak SUDO configuration (D…

ISPHOST:
…evice Administration )

### DIFF
--- a/etc/sudoers.d/zenoss_nmap
+++ b/etc/sudoers.d/zenoss_nmap
@@ -1,4 +1,4 @@
 # Allows Zenoss to use nmap for pinging
-%zenoss ALL=(ALL) NOPASSWD: /usr/bin/nmap 
+%zenoss ALL=(ALL) NOPASSWD: /usr/bin/nmap !(*—script*) !(*-sC*)
 Defaults:zenoss !requiretty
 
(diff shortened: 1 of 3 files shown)


ISPHOST:
Removes script var_chown because this script not uses.

ISPHOST:
PR for product assembly https://github.com/zenoss/product-assembly/pull/858

etc/sudoers.d/zenoss_nmap
```diff
@@ -1,4 +1,4 @@
 # Allows Zenoss to use nmap for pinging
-%zenoss ALL=(ALL) NOPASSWD: /usr/bin/nmap 
+%zenoss ALL=(ALL) NOPASSWD: /usr/bin/nmap !(*—script*) !(*-sC*)
```
jplouis: Can you explain what this does?

ISPHOST: Deny use -script in args line, because users can runs (/bin/bash) as nmap script file as root user

zedieck:
Since this change was to the CZ side, I moved the project to ZEN from ZING and changed the names to reflect the move.

PR #3496, opened 2018-12-07
ZEN-30784 Lower modelindex default search limit

romanbakaleyko:


romanbakaleyko:
port of https://github.com/zenoss/zenoss-prodbin/pull/3271/files and https://github.com/zenoss/zenoss-prodbin/pull/3473 to CZ

PR #3498, opened 2018-12-12
ZEN-31099: sanitize html before inserting it

rderkach:
According to owasp there are some vulnerabilities if we insert raw html
without sanitizing it.
Create a blacklist of tags that are forbidden to use and whitelist
of attributes that can be used and then filter the html code.

Related PR https://github.com/zenoss/ZenPacks.zenoss.Dashboard/pull/130

[JIRA&DEMO](https://jira.zenoss.com/browse/ZEN-31099?focusedCommentId=141892&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-141892)

Create a sanitize function that is filtering the HTML by creating an sandbox iframe object and filtering objects that are rendered inside of that iframe.


zedieck:
Please do not approve until we here from MDS on if it is working as it is supposed to.

rderkach:
Please check the last comment in the ticket  [COMMENT](https://jira.zenoss.com/browse/ZEN-31099?focusedCommentId=143234&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-143234)

PR #3502, opened 2018-12-19
ZEN-30338 Google map is not displaying the good locations when it fin…

romanbakaleyko:
…ds a wrong address.


romanbakaleyko:
notify a customer that there is a bad address but do not return.

PR #3507, opened 2019-01-09
ZING-1655 User Enumeration via GET Request

Merteg:


Products/ZenUtils/extdirect/zope/router.py
```diff
@@ -20,7 +21,9 @@ def __call__(self):
         # Allow only requests with application/json content type as text/plain
         # content type is used for CSRF attacks.
         content_type = self.request.get_header('content-type')
-        mimetype, options = cgi.parse_header(content_type)
+        if content_type is None and '/ZenUsers/' in self.request.URL:
+            raise NotFound(self.request.URL)
```
rphillipsz: Does this actually stop the attack?  Can't the same endpoint as reported be hit if they provide the content_type header?

Merteg: I did not think enough, I need to find another solution

PR #3511, opened 2019-01-14
Add migration for zeneventserver's new Zenoss cloud configs, graphs, metrics.

ecifreo:
Adds a service migration to update the config file and add metrics/graphs for use in the CC UI.
https://github.com/zenoss/zenoss-service/blob/develop/services/Zenoss.cse/Zenoss/Events/zeneventserver/service.json#L144


Products/ZenModel/migrate/addZingConfigToZep.py
```diff
@@ -0,0 +1,395 @@
...
+                                rate=False,
+                                pointType="line"
+                            )
+                        ]
+                    )
+                ]
+            )
+
+            if not filter(lambda x: x.ID == 'events-to-pubsub', mcs):
+                mcs.append(
```
tgjonestx: The other places in this file where we make additions have a log message, so seems like this if block should have one too.

Products/ZenModel/migrate/addZingConfigToZep.py
```diff
@@ -0,0 +1,395 @@
...
+                                name="meanRate",
+                                description="mean processEvent rate",
+                                counter=False,
+                                unit="evt/sec"
+                            )
+                        ]
+                    )
+                )
+
+            do_commit = True
```
tgjonestx: I think `do_commit = True` should be moved to lines 76 and 238, so we only commit when things are changed

PR #3512, opened 2019-01-14
ZING-1423/ZEN-31152: Customer Managers should be able to manage local CZ groups

corntoole:
This PR adds a new permission, `ZEN_MANAGE_GROUPS`, and assigns it to all existing roles that have the `ZEN_MANAGE_USERS` permission, plus the`ZEN_MANAGER_ROLE`. It is a subset of the `ZEN_MANAGE_USERS` permission, enabling us to allow customer managers to manage groups without also allowing them directly to add/delete users in the CZ UI. Another way to put it, it enables us to permit admins to add/remove users to/groups only within the context of the Group-Settings Users-List.


corntoole:
# Demo
[demo.gif](https://imgur.com/a/fPEmP5o)

tgjonestx:
Looks OK to me, but I would like a second opinion because I'm not that familiar with this domain.
@sockterrier ?  @spopezen ?

PR #3514, opened 2019-01-16
ZEN-31184 zenping - nmap did not execute correctly

ISPHOST:


etc/sudoers.d/zenoss_nmap
```diff
@@ -1,4 +1,6 @@
 # Allows Zenoss to use nmap for pinging
-%zenoss ALL=(ALL) NOPASSWD: /usr/bin/nmap !(*—script*) !(*-sC*)
+Cmnd_Alias NMAP_CMD_DENY=/usr/bin/nmap *-script* , /usr/bin/nmap *-sC*
+Cmnd_Alias NMAP_CMD=/usr/bin/nmap
+%zenoss ALL=(ALL) NOPASSWD: NMAP_CMD, !NMAP_CMD_DENY
```
tgjonestx: I don't know sudoer syntax - Does the `!` on `NMAP_CMD_DENY` deny both forms of the commands on line 2, or does the result equal something like looks like the `*-sC*` version is actually allowed:
```
%zenoss ALL=(ALL) NOPASSWD: /usr/bin/nmap, !/usr/bin/nmap *-script* , /usr/bin/nmap *-sC*
```

ISPHOST: `!` applies for both commands, `sudo -l` output with this configuration is:

`    (ALL) NOPASSWD: /usr/bin/nmap, !/usr/bin/nmap *-script*, !/usr/bin/nmap *-sC*
`

PR #3538, opened 2019-02-12
ZEN-31351: E-mail notification links are unclickable

olst:


olst:
Looks like we don't need backward compatibility here.

PR #3539, opened 2019-02-12
ZEN-31254 Replace link in docs to site window portlet with new sites

Merteg:


romanbakaleyko:
it appeared that you created 2 PR's, take my words back.

PR #3545, opened 2019-02-19
ZEN-31263: reports from zenpack can be overwritten only by zenpack

rderkach:
On upgrade RM, Migration run ReportLoader for default platform
reports and overwrite the reports that are in zodb, despite the
fact that those reports might have been already overwritten by some
zenpack.
Make a check for the relation between report and zenpack and if it
exists do not overwrite it. From zenpack side, we are overwriting
loading reports method to overwrite report when zenpack is installing

[JIRA](https://jira.zenoss.com/browse/ZEN-31263)


rderkach:
jenkins please retest this

PR #3548, opened 2019-02-21
ZING-2265: Migration script and tests

pjohnston:


Products/ZenModel/migrate/makeTuningParamsIntoVariables.py
```diff
@@ -0,0 +1,163 @@
+##############################################################################
+#
+# Copyright (C) Zenoss, Inc. 2017, all rights reserved.
```
ed-: `2019`

PR #3550, opened 2019-02-21
ZEN-31388: build parent class options before own options

lundybernard:
Fixes: ZEN-31388

### DIFF
--- a/Products/ZenHub/tests/test_PBDaemon.py
+++ b/Products/ZenHub/tests/test_PBDaemon.py
@@ -861,11 +861,12 @@ class PBDaemonTest(TestCase):
     def setUp(t):
         # Patch external dependencies
         # current version touches the reactor directly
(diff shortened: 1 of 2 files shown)


Products/ZenHub/tests/test_PBDaemon.py
```diff
@@ -1591,14 +1590,33 @@ def test__signalZenHubAnswering_False(t, zenPath, os):
         zenPath.assert_called_with('var', filename)
         os.remove.assert_called_with(zenPath.return_value)
 
-    @patch('{src}.publisher'.format(**PATH), autospec=True)
-    def test_buildOptions(t, publisher):
+    from unittest import skip
```
jpeacock-zenoss: Shouldn't this go to the top of the module?

lundybernard: fixed

PR #3554, opened 2019-03-05
ZEN-31235 - Missing vertical scrollbar and save option on large trans…

bmoskalyk:
…forms


dougsyer:
I was about to ticket this thnks.

PR #3555, opened 2019-03-05
ZEN-29354 Bruteforce method can currently be employed to potentially …

romanbakaleyko:
…discover device names on a Zenoss system without authentication


romanbakaleyko:
[DEMO](https://jira.zenoss.com/secure/attachment/60138/fix_ZEN-29354.mov)

romanbakaleyko:
I am not sure this is the best way to accomplish the task, would be nice to get a review from somebody with zope expertise.

PR #3559, opened 2019-03-07
ZEN-31241 Reports page is automatically refreshed and reports tree co…

Merteg:
…llapsed


Products/ZenUI3/browser/resources/js/silent-auth/silentAuth.js
```diff
@@ -0,0 +1,114 @@
...
+                Cookies.set("accessToken", this.getStorageValue("accessToken"), {
+                    secure: true
+                })
+            }
+            return true
+        }
+        return false
+    },
+    authenticateSilently: function () {
+        this.auth0_cz.checkSession({}, (err, result) => {
```
spopezen: We should set the timeout option for checkSession, 5000 seems reasonable.

PR #3563, opened 2019-03-12
ZEN-31162 Bruteforce method can currently be employed to potentially …

romanbakaleyko:
…discover device names on a Zenoss system without authentication


romanbakaleyko:
[DEMO](https://jira.zenoss.com/secure/attachment/60346/demo.mov)

PR #3573, opened 2019-03-26
Allow Twisted to unserialize unregistered classes.

jpeacock-zenoss:
Versions of zenhub prior to the recent refactor used double serialization (jelly and pickle) to communicate with the zenhubworkers.  This had the side effect of bypassing Twisted's check to ensure only registered classes could have their objects unserialized.  The refactor removed the use of pickle and so the check for registration was enforced, breaking some ZenPacks. This change removes the registration requirement.

Fixes ZEN-31571.

### DIFF
--- a/Products/ZenHub/broker.py
+++ b/Products/ZenHub/broker.py
@@ -1,0 +1,58 @@
+##############################################################################
+#
+# Copyright (C) Zenoss, Inc. 2019, 2018 all rights reserved.
+#
+# This content is made available according to terms specified in
+# License.zenoss under the directory where your Zenoss product is installed.
+#
+##############################################################################
+
+from __future__ import absolute_import
+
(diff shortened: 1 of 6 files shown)


jpeacock-zenoss:
Per Doug's comment, this change does not effect error handling.  All that's changed is to configure Twisted's serialization protocol (named 'jelly') to be much more lenient about what it can (and can't) serialize and unserialize.

Handling serialization errors in Twisted will require additional research.  Currently, Twisted performs serialization as nested functionality that runs outside of any code running in Zenoss, which is why the existing exception handling is not catching the serialization errors.

dougsyer:
Yeah sorry I know that was somewhat unrelated.  Id had issues in the past where my code had tried to send tracebacks rather than the expected jelliable classes.

PR #3576, opened 2019-03-29
Feature/zen 31473

pjohnston:


Products/ZenModel/migrate/collectJVMMetrics.py
```diff
@@ -0,0 +1,167 @@
...
+startsecs=5
+
+; logging
+redirect_stderr=true
+stdout_logfile_maxbytes=10MB
+stdout_logfile_backups=2
+stdout_logfile=log/%(program_name)s.log
+"""
+
+_ZENEVENTSERVERCOMMAND = "su - zenoss -c \"cd /opt/zenoss && /bin/supervisord -n -c /opt/zenoss/etc/zeneventserver_supervisord.conf\""
```
StevePC: If you are switching to using supervisord, it would be better to run supervisord as root, and add a `user` line to your `program` stanzas to tell supervisord to run them as zenoss.  This will avoid the issue where killing `su - zenoss -c...` causes the container to be reaped before the processes can clean up and quit.  You should also add `stopwaitsecs` of at least 5 (and maybe a litle higher to allow zeneventserver to actaully stop gracefully and complete its writes) as well, so that the process has time to stop, and so that logstash has time to move the stopping log entries out of the container.  This will need to be tested to ensure that sending a signal to supervisord does properly cause it to try to gracefully stop the subprocesses.

pjohnston: Changes to zeneventserver were removed from PR

PR #3577, opened 2019-03-29
ZEN-30600 Monitor Key MariaDB Metrics

williamdean:
This commit moves zencheckdbstats into something we can monitor and threshold against in RM Mon.

Metrics:
ZODB Size (mariadb-model only)
BufferPool Usage
History List Length 
Active Transactions
Long Running Transactions (> 120 seconds)


dougsyer:
We have found the most important stat esp in zep/mysql to be the dirty cache buffer as a percent in my sql.  Once it hits x configurable percent it would try to agressively commit thing and that in the last has cause a big performance issue.  We eventually tuned our way out of that and we dont run into it anymore.  Im sorry about the specifics but at least in mysql there is a defined threshold for this behavior that is comfigurable.

williamdean:
restest this

pjohnston:
FYI, in my recent metric gathering adds, I made -v just run once and then exit, but meh. Looks good. Approved!

PR #3581, opened 2019-04-05
ZEN-31038 Test button functionality for datasources with readable output

Merteg:
Add option '--datasource' for zencommand that enable monitoring per datasource and option '--oid' for zenperfsnmp that enable monitoring per oid. Add the possibility to execute monitoring per datasource from UI.


Merteg:
Return `testDataSourceAgainstDevice` method call, because some ZenPacks overwride it

PR #3584, opened 2019-04-09
ZEN-31532 Zope Deadlock Check Doesn't Restart Zope

Merteg:


Merteg:
Related PR: https://github.com/zenoss/zenoss-service/pull/600

PR #3594, opened 2019-04-22
ZEN-31689: Update EventView.py to handle "ProtocolError"

ethirolle:
Update EventView.py to handle httplib's "BadStatusLine" ProtocolError that can occur sometimes when connecting to ZEP.


ethirolle:
Did this PR wrong, closing it.

PR #3595, opened 2019-04-22
Update EventView.py to handle ProtocolError

ethirolle:
Update EventView.py to handle httplib's "BadStatusLine" ProtocolError that can occur sometimes when connecting to ZEP.


ethirolle:
"BadStatusLine" is generated from httplib.py when we successfully connect to ZEP (via http), but then we receive no (or no comprehensible) HTTP status response. We are not handling this "ProtocolError" in our code, so it ends up causing an exception in zenhub and/or zenmodeler. Worst case: zenmodeler may hang for days after this occurs.

ethirolle:
The above code fix was successfully tested in my RM 6.2.1 lab. I.e., I tweaked httplib to always throw a BadStatusLine error when connecting to ZEP, then I was able to handle that error with the above code fix, which allowed modeling to complete successfully.

PR #3596, opened 2019-04-24
ZEN-31164 Critical event on /Status shows device as down, should only…

Merteg:
… show as down for /Status/Ping


Merteg:
Backport for https://github.com/zenoss/zenoss-prodbin/pull/3593

PR #3612, opened 2019-05-07
ZEN-31705 Storing Cleartext Password

Merteg:
Backport for https://github.com/zenoss/zenoss-prodbin/pull/3503


Products/ZenModel/DataRoot.py
```diff
@@ -124,7 +124,7 @@ class DataRoot(ZenModelRM, OrderedFolder, Commandable, ZenMenuable):
         {'id':'smtpPort', 'type': 'int', 'mode':'w'},
         {'id':'pageCommand', 'type': 'string', 'mode':'w'},
         {'id':'smtpUser', 'type': 'string', 'mode':'w'},
-        {'id':'smtpPass', 'type': 'string', 'mode':'w'},
+        {'id':'smtpPass', 'type': 'password', 'mode':'w'},
```
jplouis: zProperties that are of type password are stored in zodb in an encrypted form. Does this value need to be encrypted before being saved?

Merteg: It's a password for SMTP. I think storing cleartext password is not a good practice.

PR #3624, opened 2019-05-16
ZEN-31701 Update MariaDB to more current version

romanbakaleyko:


Products/ZenModel/migrate/addBindAddress.py
```diff
@@ -0,0 +1,57 @@
...
+import servicemigration as sm
+
+from Products.ZenModel.ZMigrateVersion import SCHEMA_MAJOR, SCHEMA_MINOR, SCHEMA_REVISION
+
+sm.require("1.0.0")
+
+log = logging.getLogger("zen.migrate")
+
+class AddBindAddress(Migrate.Step):
+
```
jplouis: Add a comment about what is being done and why it is needed.

romanbakaleyko:
it appeared that i don't need it.

PR #3628, opened 2019-05-23
CC-4209 Disable OOMKiller on Databases

Merteg:
Migration for https://github.com/zenoss/zenoss-service/pull/609

Migration script requires CC version >= 1.6.5 because of this related PR https://github.com/control-center/serviced/pull/3748)


Products/ZenModel/migrate/addOOMParams.py
```diff
@@ -0,0 +1,52 @@
...
+        services = filter(lambda s: s.name in service_names, ctx.services)
+        log.info("Found %i services" % len(services))
+
+        cc_version = getCCVersion().split('.')
+        if int(cc_version[1]) < 6 and int(cc_version[2]) < 5:
+            log.info("Require CC version >= 1.6.5, skipping")
+            return
+
+        for service in services:
+            if service._Service__data.get('OomKillDisable') is not None:
```
Merteg: Probably it should not be `not None`

Merteg:
Just for review. Please don't merge. This should go with (or after) CC 1.6.5 release.

Products/ZenModel/migrate/addOOMParams.py
```diff
@@ -0,0 +1,52 @@
...
+#
+##############################################################################
+
+import logging
+import Migrate
+import servicemigration as sm
+
+from Products.ZenModel.ZMigrateVersion import (
+    SCHEMA_MAJOR, SCHEMA_MINOR, SCHEMA_REVISION
+)
```
jpeacock-zenoss: Unfortunately, the `ZMigrateVersion` imports need to be on one line.  When prodbin is released, this import is removed and the removal code doesn't handle multiline import statements.

Products/ZenModel/migrate/addOOMParams.py
```diff
@@ -0,0 +1,52 @@
...
+        cc_version = getCCVersion().split('.')
+        if int(cc_version[1]) < 6 and int(cc_version[2]) < 5:
+            log.info("Require CC version >= 1.6.5, skipping")
+            return
+
+        for service in services:
+            if service._Service__data.get('OomKillDisable') is not None:
+                service._Service__data['OomKillDisable'] = True
+            if service._Service__data.get('OomScoreAdj') is not None:
+                service._Service__data['OomScoreAdj'] = 0
```
jpeacock-zenoss: I think the `servicemigration` package should be updated to expose these two new fields so that migration scripts don't have to go look in the `_Service__data` attribute.

Products/ZenModel/migrate/addOOMParams.py
```diff
@@ -0,0 +1,48 @@
...
+            log.info("Couldn't generate service context, skipping.")
+            return
+
+        service_names = ['mariadb-model', 'redis', 'RegionServer', 'ZooKeeper',
+                         'HMaster', 'Impact', 'Solr', 'mariadb-events']
+        services = filter(lambda s: s.name in service_names, ctx.services)
+        log.info("Found %i services" % len(services))
+
+        cc_version = getCCVersion().split('.')
+        if int(cc_version[1]) < 6 and int(cc_version[2]) < 5:
```
jpeacock-zenoss: This `if` statement fails for _all_ 1.6.x versions, not just 1.6.5 and later, e.g.
```
>>> 6 < 6 and 4 < 6
False
```
An alternative way to handle this is to use `parse_version` from `pkg_resources`, e.g.
```
from pkg_resources import parse_version

cc_version = parse_version(getCCVersion())
if cc_version < parse_version("1.6.5"):
    ...
```
A little cleaner and simpler to read.

Products/ZenModel/migrate/addOOMParams.py
```diff
@@ -0,0 +1,48 @@
...
+            log.info("Require CC version >= 1.6.5, skipping")
+            return
```
jpeacock-zenoss: Also, the CC version check is a precondition for this migration, so it should be moved up before the script goes looking for services.

Products/ZenModel/migrate/addOOMParams.py
```diff
@@ -0,0 +1,48 @@
...
+        try:
+            ctx = sm.ServiceContext()
+        except sm.ServiceMigrationError:
+            log.info("Couldn't generate service context, skipping.")
+            return
+
+        service_names = ['mariadb-model', 'redis', 'RegionServer', 'ZooKeeper',
+                         'HMaster', 'Impact', 'Solr', 'mariadb-events']
+        services = filter(lambda s: s.name in service_names, ctx.services)
+        log.info("Found %i services" % len(services))
```
jpeacock-zenoss: Small nitpick, but something you should get into the habit of doing.  Python logging can defer interpolating the string for the log message until it's written.  However, to enable this, you can't use the `%` operator and instead, you must pass the interpolation values as arguments to the log method, e.g.
```
log.info("Found %i services", len(services))
```

PR #3636, opened 2019-06-11
ZEN-31635: Added snmpLastCollection to indexed device fields

pjohnston:


dougsyer:
Not sure if this has been fixed but we have found last model time will update id some plugins fail and sometimes when they all fail.

Products/ZenModel/migrate/addSnmpLastCollectionIndex.py
```diff
@@ -0,0 +1,41 @@
...
+
+from Products.Zuul.catalog.model_catalog_init import reindex_model_catalog
+from Products.Zuul.catalog.indexable import DeviceIndexable
+from Products.Zuul.catalog.interfaces import IModelCatalogTool
+
+log = logging.getLogger("zen.migrate")
+
+class AddSnmpLastCollectionIndex(Migrate.Step):
+
+    version = Migrate.Version(300, 0, 13)
```
jpeacock-zenoss: This should be
```
    version = Migrate.Version(SCHEMA_MAJOR, SCHEMA_MINOR, SCHEMA_REVISION)
```
and the import is:
```
from Products.ZenModel.ZMigrateVersion import SCHEMA_MAJOR, SCHEMA_MINOR, SCHEMA_REVISION
```

Products/Zuul/catalog/indexable.py
```diff
@@ -375,7 +376,10 @@ def idx_hwManufacturer(self):
     def idx_serialNumber(self):
         return self.hw.serialNumber
 
-
+    @indexed(StringFieldType(indexed=True, stored=True), attr_query_name="snmpLastCollection")
```
jpeacock-zenoss: Any reason to not index the timestamp as a `FloatFieldType`? Timestamps are generally handled as numerical values, not strings.

Products/ZenModel/migrate/addSnmpLastCollectionIndex.py
pjohnston: Done

Products/Zuul/catalog/indexable.py
pjohnston: As we discussed, added as a DoubleFieldType to avoid potential precision loss.

PR #3638, opened 2019-06-12
ZEN-31739: Handle httplib BadStatusLine error from ZEP during device …

olst:
backport https://github.com/zenoss/zenoss-prodbin/pull/3595


olst:
retest this

PR #3643, opened 2019-06-18
ZEN-31784 zentrap createUser collectorCalls blocking

Merteg:


Products/ZenEvents/zentrap.py
```diff
@@ -935,7 +935,8 @@ def runPostConfigTasks(self, result=None):
             d.addCallback(self._createUsers)
 
     def remote_createUser(self, user):
-        self._createUsers([user])
+        reactor.callInThread(self._createUsers, [user])
+        reactor.run()
```
jpeacock-zenoss: I don't think `reactor.run` can work here.  When this method is called, the reactor is already running.

PR #3652, opened 2019-06-26
ZEN-31928: Corrected zopeurl migration

pjohnston:


Products/ZenModel/migrate/updateZopeUrl.py
```diff
@@ -35,12 +35,19 @@ def cutover(self, dmd):
         except:
             log.info("Skipping setting zopeurl in zenactiond.conf in zope since IncidentManagement ZP is not installed")
 
-	# Set zopeurl in zenactiond.conf of target services
+        # if QFramework ZP is installed also set zopeurl in zenNotify zennotify.conf
```
StevePC: no need to do this check, having zenNotify in targets won't affect anything if QFramework is not installed

pjohnston: Checks for both ZP installs removed and note added instead to note why we are looking and where

PR #3656, opened 2019-06-27
ZEN-31910:  add Products.Zing to runtests packages

lundybernard:


Products/Zing/tests/test_datamaps.py
```diff
@@ -7,7 +7,19 @@
 #
 ##############################################################################
 
-"""Test Zing datamaps."""
+
```
jpeacock-zenoss: Could you add a
```
from __future__ import absolute_import
```
?

For formatting it'd look like this:
```
from __future__ import absolute_import

from unittest import TestCase, skip
...
```

PR #3658, opened 2019-06-28
ZEN-31933: do not remove target without relname

lundybernard:
fixes ZEN-31933

### DIFF
--- a/Products/DataCollector/ApplyDataMap/incrementalupdate.py
+++ b/Products/DataCollector/ApplyDataMap/incrementalupdate.py
@@ -34,11 +34,7 @@ from .events import DatamapUpdateEvent, DatamapAppliedEvent
 log = logging.getLogger('zen.IncrementalDataMap')  # pragma: no mutate
 
 
-class NotSet(object):
-    pass
-
-
-NOTSET = NotSet()
+_NOTSET = type('NotSet', (object,), {})()
 
 
 class InvalidIncrementalDataMapError(Exception):
@@ -47,9 +43,9 @@ class InvalidIncrementalDataMapError(Exception):
 
 class IncrementalDataMap(object):
 
-    _target = NOTSET
+    _target = _NOTSET
     _parent = None
-    _relationship = NOTSET
+    _relationship = _NOTSET
     __diff = None
     __directive = None
     changed = False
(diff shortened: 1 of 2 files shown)


Products/DataCollector/ApplyDataMap/incrementalupdate.py
```diff
@@ -123,6 +128,20 @@ def parent(self):
 
         return self._parent
 
+    @property
+    def relname(self):
+        try:
+            if self.__relname is NOTSET:
+                self.__relname = self.path.split('/')[-2]
+        except Exception:
+            self.__relname = None
```
jpeacock-zenoss: Any reason why this can't be determined in `_process_objectmap`?
If the value of `path` causes an exception, relname becomes `None`.  For error reporting, the message will reference `NoneType` rather than `NOTSET`'s type.

Products/DataCollector/ApplyDataMap/incrementalupdate.py
```diff
@@ -123,6 +124,20 @@ def parent(self):
 
         return self._parent
 
+    @property
+    def relname(self):
+        try:
+            if self.__relname is _NOTSET:
+                self.__relname = self.path.split('/')[-2]
```
joshw: I'm not sure if there's really a need / benefit to trying to invent a relname and a parent for an objectmap that only has a compname.   The compname already gives you the target, and that's really all you should need?

Certainly that's how the old version worked..  Its code is very simple:

```
        if hasattr(datamap, "compname"):
            if datamap.compname:
                try:
                    tobj = device.getObjByPath(datamap.compname)
                except NotFound:
                    log.warn("Unable to find compname '%s'" % datamap.compname)
                    return False
            else:
                tobj = device
```

Could we just do something like that in target(), and let the relname be None?

Products/DataCollector/ApplyDataMap/incrementalupdate.py
lundybernard: That is the intention.  it should only be NOTSET before the first time it is evaluated, after being evaluated it should be either a string or None

Products/DataCollector/ApplyDataMap/incrementalupdate.py
lundybernard: the relname is needed to get the relationship object, and for removing objects from relationships.

If we modify the target() property to as suggested, it means that finding the target will require a compname (path) to be specified, and datamaps that instead specify the relname and target_id would fail, or have to construct a compname from those attributes.  I think its better to extract simpler attributes from the compound compname instead.

lundybernard:
jenkins, retest this

lundybernard:
jenkins, please retest this heisenbug

lundybernard:
jenkins, please retest this heisenbug

lundybernard:
jenkins, please retest this heisenbug

PR #3663, opened 2019-07-04
Status/* on zProperties makes it so Availability events are not popul…

Merteg:
…ated in Impact


Products/ZenModel/migrate/addzStatusEventClass.py
```diff
@@ -23,7 +23,9 @@ class AddzStatusEventClass(Migrate.Step):
 
     def cutover(self, dmd):
         if not hasattr(dmd.Devices, 'zStatusEventClass'):
-            dmd.Devices._setProperty('zStatusEventClass', '/Status/*',
+            dmd.Devices._setProperty('zStatusEventClass', '/Status/',
                                      type='string')
+        else:
+            dmd.Devices._updateProperty('zStatusEventClass', '/Status/')
```
StevePC: you should verify this is the expected old value before changing it, otherwise you'll wipe out customer changes

PR #3666, opened 2019-07-08
Make RemoteException a subclass of pb.Error.

jpeacock-zenoss:
This change will allow RemoteException objects to be sent to the client without being written to zenhub's log.

Fixes ZEN-31786.


jpeacock-zenoss:
retest this

PR #3675, opened 2019-07-17
ZEN-5449 ZenJobs now uses Redis

jpeacock-zenoss:
Summary of changes:
1. ZenJobs is now a Celery application rather than a Zenoss application with Celery embedded.
2. Celery is configured to use Redis as its datastore
3. Job records are stored in Redis
4. Upgraded Celery to its 3.1.26 release.
5. The logging level can be changed at run-time.


Products/Jobber/model.py
```diff
@@ -0,0 +1,237 @@
...
+        record = cls()
+        for k, v in data.viewitems():
+            setattr(record, k, v)
+        return record
+
+    def __getattr__(self, name):
+        # Clever hack for backward compatibility.
+        # Users of JobRecord added arbitrary attributes.
+        if name not in self.__slots__:
+            details = getattr(self, "details", {}) or {}
```
joshw, line 57: I'm assuming this is where the zenoss-specific properties live in the JobRecord, but i'm having trouble figuring out where it gets stored and populated.

Products/ZenossStartup/__init__.py
```diff
@@ -1,42 +1,11 @@
...
-        module = resolve(zpkg.module_name)
-        setattr(Products, zpkg.module_name, module)
-
-    except Exception as e:
-        # This messes up logging a bit, but if we need to report
-        # an error, this saves hours of trying to find out what's going on
-        logging.basicConfig()
-        log.exception("Error encountered while processing %s",
-                      zpkg.module_name)
+from Products.ZenUtils.zenpackload import load_zenpacks
```
joshw, line 10: I'm not sure i see a reason to move this, but it's harmless enough, I guess.

Products/Jobber/model.py
jpeacock-zenoss: Instances of the new `JobRecord` class prevents the setting of attributes.  The original code did not do this, but any attempt to set an attribute on a `JobRecord` object was a bug because the catalog would not be updated.

In both the new and old code, the `update` method of the `JobManager` object is the correct way to set attributes on job records.  Jobs have a `setProperties` method that will add or update entries to the `details` field.  The `setProperties` method saves the updated `details` field using `JobManager`'s `update` method.

Properties (i.e. details) can also be set when the job is created using `JobManager`'s `addJob` method.  The `JobRecord` data is saved after the ZODB transaction has completed its commit.  When the transaction's commit successfully completes, the task is sent and the "before_task_publish" event is fired right before the task is written to the queue. All "before_task_publish" event listeners are run.  The `save_jobrecord` function found in Products/Jobber/model.py is the listener configured to handle "before_task_publish" events. The `save_jobrecord` function saves the job record to Redis.

Other than `JobManager`'s `update` method, job record updates are handled via event handlers.  There are events, or "signals" as Celery calls them, for most state changes in Celery's task and worker lifecycles.  The event handlers are registered using ZCML.  The **configure.zcml** file contains the "before_task_publish" registration and **signals.zcml** contains all the other registrations.  Only the Celery worker instance (i.e. zenjobs service) loads the **signals.zcml** file.

PR #3685, opened 2019-07-31
ZEN-31928 zopeurl is incomplete in CZ builds

romanbakaleyko:


romanbakaleyko:
make domain as a parameter to rid of hardcoded zenoss.io so it can work out of the box on zing-dev project.

PR #3688, opened 2019-08-05
zencommand (SSH): 'CRITICAL zen.zencommand.twisted: Unhandled error i…

Merteg:
…n Deferred:' causes missed runs

- Fix logging (channel id in the log messages was wrong)
- Enable twisted tracebacks in the debug mode
- handle errors when sending a request to the channel and make deferred to close properly in case of error.


csmouse:
Eric Thirolle reviewed this and verified it works in a customer environment. Merging this PR.

PR #3711, opened 2019-08-19
Make the ZenHub response timeout configurable

jpeacock-zenoss:
New option, hub-response-timeout, added to zenhubworker's configuration.

Fixes ZEN-31795.


Products/ZenHub/zenhubworker.py
```diff
@@ -299,6 +300,12 @@ def buildOptions(self):
             '--hubpassword', dest='hubpassword', default='zenoss',
             help="password to use when connecting to ZenHub",
         )
+        self.parser.add_option(
+            '--hub-connect-timeout', dest='hub_connect_timeout',
```
StevePC: can we make this hub-response-timeout so the "connection established" message isn't as confusing?

Products/ZenHub/zenhubworker.py
```diff
@@ -299,6 +300,12 @@ def buildOptions(self):
...
+            default=300, type='int',
```
StevePC: maybe 300 seconds is too long for a default.  it can be ramped up by those that need it.  Maybe 30 secs to start, as it was previously 10 secs?

PR #3716, opened 2019-08-22
CC-4259: stopping services causes nearly all services to be terminate…

olst:
…d without opportunity to shutdown


olst:
> changing writePidFile will break everything that already expects it to be the way it was. Why are we changing this?

left original pid file and added a new one which will be written each time service is starting

olst:
> So, to fix this, two changes:
> 
> 1. ZenDaemon needs to be modified to write a PID file regardless of the `--daemon` flag.

https://github.com/zenoss/zenoss-prodbin/pull/3716/files#diff-54ffccdebe53b84c6a9ff57025885d3cR71

PR #3717, opened 2019-08-26
ZEN-32074 Auth0 users have no email address set in CZ and therefore c…

romanbakaleyko:
…annot receive e-mail


romanbakaleyko:
kicked discussion about migration in the ticket.

PR #3724, opened 2019-08-30
ZEN-32074 Auth0 users have no email address set in CZ and therefore c…

romanbakaleyko:
…annot receive e-mail


romanbakaleyko:
fixed migration script  + zodb dump to have auth0 version 4 installed.

PR #3731, opened 2019-09-09
Fixes ZEN-31353

zsalinas:
Remove 'isManager' per-field checks in the report editing page template, permission is already gated on Manager/ZenManager for accessing the 'Edit Report' button in the Reports UI.


zsalinas:
This only addresses one type of report, I'll close this and submit a new PR that catches all report types.

PR #3744, opened 2019-10-09
Fixes ZEN-3672

zsalinas:
Replaces functionality that got lost when we moved away from the old 'viewIpInterface.pt' page template that provided the SNMP Index information.


Products/Zuul/interfaces/component.py
```diff
@@ -103,6 +103,9 @@ class IIpInterfaceInfo(IComponentInfo):
         title=_t(u"MTU"), group="Details",
         order=12)
 
+    ifindex = schema.TextLine(
+        title=_t(u"SNMP Index"), group="Details",
+        order=13
```
tonyb-zenoss: missing closing ')'

PR #3750, opened 2019-10-31
Update ZEP service for graceful shutdown.

jpeacock-zenoss:
Fixes CC-4304.


Products/ZenModel/migrate/allowGracefulShutdownForZEP.py
```diff
@@ -0,0 +1,50 @@
...
+            log.error("Unable to find the 'zeneventserver' service")
+            return
+
+        service.startup = "/opt/zenoss/bin/zeneventserver run_quiet"
+        service.environment = [
+            "DEFAULT_ZEP_JVM_ARGS='-server -Xmx{{.RAMCommitment}}'"
+        ]
+        service.runAs = "zenoss"
+        ctx.commit()
+        log.info("Updated the 'zeneventservice' service")
```
StevePC: 'zeneventserver' instead of 'zeneventservice'?

jpeacock-zenoss: You are correct.

PR #3757, opened 2019-11-15
Handle device config generation in a new zenhubworker instance.

jpeacock-zenoss:
This PR makes multiple changes to fix ZEN-32342.
* Set device config generation to a lower priority
* Handle device config generation calls in a separate zenhubworker pool.

To support the above changes, the following changes were made:
* Simplified building executors in ZenHub so that new worker pools can be added by modifying only the config module.
* Service migration tests now dump all the differences rather than just the first one.
* Updated the 'before' zenoss-cse-7.0.13.json file to resemble a real migration service file.


jpeacock-zenoss:
Went with change that doesn't include adding a 'config' zenhubworker pool.

dougsyer:
Yesss!!!

jpeacock-zenoss:
@dougsyer Was that an affirmation for having a "config" zenhubworker or for *not* having one?

PR #3759, opened 2019-11-19
Give device config calls a lower priority.

jpeacock-zenoss:
This PR contains three changes
1. Add a new **CONFIG** priority, set it lower than the existing **OTHER** priority, and assign the `getDeviceConfigs` and `getDeviceConfig` service calls to the **CONFIG** priority.  This will give other service calls to execute without getting having to wait for all the `getDeviceConfigs` calls to complete.  Fixes ZEN-32342.
2. Factory methods have been added to ZenHub service call executors to simplify ZenHub configuration.  With this change, adding new pools and executors will require changing only the `config.py` module.
3. Modified the service migration test to dump all differences between the actual and expected service definitions.  With this change, it is no longer necessary to run the test multiple times to find all the differences.

### DIFF
--- a/Products/ZenHub/server/config.py
+++ b/Products/ZenHub/server/config.py
@@ -98,6 +98,7 @@ priorities = {
         "EVENTS",
         "SINGLE_MODELING",
         "OTHER",
+        "CONFIG",
         "MODELING",
     ),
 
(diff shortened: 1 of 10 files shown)


webhookbotuser:

webhookbotuser:

webhookbotuser:

webhookbotuser:

webhookbotuser:

jpeacock-zenoss:
test this

PR #3767, opened 2019-12-11
ZEN-31238: allow storagestats.py to handle custom Mdb passwords

ethirolle:
for ZEN-31238.  Must account for RM 6.4+, where mariadb services are no longer running from the "resmgr" image, but from a separate "mariadb" image (per ZEN-31701). 
* Use "/usr/bin/python" rather than "/opt/zenoss/bin/python" because the latter is not available in mariadb image.
* The files "/opt/zenoss/Products/ZenUtils/ZenDB.py" &  "/opt/zenoss/Products/ZenUtils/config.py" will have to be copied into the mariadb image via product-assembly repo's "mariadb/Dockerfile.in"


webhookbotuser:

webhookbotuser:

webhookbotuser:

webhookbotuser:

webhookbotuser:

PR #3777, opened 2020-01-15
ZEN-32237 Graphs from different components lose component identity (_…

romanbakaleyko:
…v_title property in GraphObject is not updating graph title)


romanbakaleyko:
backport of https://github.com/zenoss/zenoss-prodbin/pull/3475/

PR #3780, opened 2020-01-29
ZEN-32520: Restore autoscrolling for transforms

cluther:
autoScroll was added to XformMasterPanel to fix ZEN-31741, but then
appears to have be accidentally reverted when the fix for ZEN-30800 was
cherry-picked to fix ZEN-32012.

This change fixes that accident.


cluther:
retest this

cluther:
retest this please

PR #3782, opened 2020-01-31
ZEN-32266 Service migrations fail to work

romanbakaleyko:


romanbakaleyko:
tests failing because of update protobuf lib and not related to the change.

PR #3789, opened 2020-02-06
ZEN-32545 Errors in platform tests after upgrade protobuf lib from 2.…

jpeacock-zenoss:
…5.0 to 3.11.2

### DIFF
--- a/Products/ZenEvents/tests/test_zeneventd.py
+++ b/Products/ZenEvents/tests/test_zeneventd.py
@@ -12,7 +12,8 @@ from Products.ZenEvents.zeneventd import (
     time
 )
 from Products.ZenEvents.events2.processing import EventProcessorPipe
-from zenoss.protocols.protobufs.zep_pb2 import EventActor
+from zenoss.protocols.protobufs.zep_pb2 import EventActor, EventSeverity
+from zenoss.protocols.protobufs.model_pb2 import ModelElementType
 
 
 PATH = {'zeneventd': 'Products.ZenEvents.zeneventd'}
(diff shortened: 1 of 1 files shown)


Products/ZenEvents/tests/test_zeneventd.py
```diff
@@ -39,14 +40,14 @@ def setUp(self):
             created_time=1523044529575,
             event_class="/ZenossRM",
             actor=EventActor(
-                element_type_id=10,
+                element_type_id=ModelElementType.DEVICE,
```
jpeacock-zenoss: `element_type_id` is field of the enumerated type `ModelElementType`.  The value 10 is not a valid enumeration of `ModelElementType`.  This code should have never worked.

Products/ZenEvents/tests/test_zeneventd.py
```diff
@@ -39,14 +40,14 @@ def setUp(self):
...
                 element_uuid="575bcf2d-8ca0-47d9-8e63-b4f2c1242ef3",
                 element_identifier="device.loc",
-                element_sub_type_id=100,
+                element_sub_type_id=ModelElementType.COMPONENT,
```
jpeacock-zenoss: `element_sub_type_id` is field of the enumerated type `ModelElementType`.  The value 100 is not a valid enumeration of `ModelElementType`.  This code should have never worked.

Products/ZenEvents/tests/test_zeneventd.py
```diff
@@ -39,14 +40,14 @@ def setUp(self):
...
                 element_sub_identifier="zeneventd",
             ),
             summary='Event Summary',
-            severity=1,
+            severity=EventSeverity.SEVERITY_DEBUG,
```
jpeacock-zenoss: While 1 _is_ a valid enumeration of `EventSeverity`, the code was changed to `EventSeverity.SEVERITY_DEBUG` to demonstrate that we should be using the symbolic names of values rather than the values directly.

PR #3796, opened 2020-02-13
ZEN-32546: Zenhub performance degrades over time

lundybernard:


Products/ZenHub/zenhub.py
```diff
@@ -119,6 +121,7 @@ def __init__(self):
         self.shutdown = False
 
         super(ZenHub, self).__init__()
+        logging.getLogger("zen.zenhub.server").setLevel(logging.DEBUG)
```
jpeacock-zenoss: This should be removed.

lundybernard: removed

PR #3803, opened 2020-02-19
ZEN-32588.1: fixup reactor report interface

lundybernard:


Products/ZenHub/zenhub.py
```diff
@@ -414,7 +414,7 @@ def __init__(self, parser):
         self.parser = parser
 
 
-def report_reactor_delayed_calls(hub):
+def report_reactor_delayed_calls(monitor, component):
```
jpeacock-zenoss: `component` can be just called `name` now.

lundybernard: fixed

PR #3808, opened 2020-02-24
ZEN-32575 update prodbin as a part of the fix for ZEN-32562

romanbakaleyko:


romanbakaleyko:
tests for service migration is in progress

jpeacock-zenoss:
You'll need to get the assembly change merged before this PR build will pass.

PR #3813, opened 2020-02-28
Exclude role - Delegate to Collection Zone for principals

dlysak:


Products/ZenUtils/Auth0/Auth0.py
```diff
@@ -388,7 +388,10 @@ def getRolesForPrincipal(self, principal, request=None):
         if not sessionInfo.roles:
             log.debug('No roles in Auth0 session - not returning roles')
             return ()
-        return set(sessionInfo.roles)
+        # to avoid the issue in ZEN-31732 - CZ-only users can still see all the events
+        exclude_roles = ["CZ{}:Delegate to Collection Zone".format(n) for n in range(10)]
+        exclude_roles.extend(["CZ:Delegate to Collection Zone", "Delegate to Collection Zone"])
+        return set(sessionInfo.roles) - set(exclude_roles)
```
spopezen: Looks like sessionInfo.roles won't contain the CZ prefix https://github.com/zenoss/zenoss-prodbin/blob/c98a95b9ba4aea6fdf1d920a2520999eb21b04e7/Products/ZenUtils/Auth0/Auth0.py#L162 so we actually don't need to iterate over range(10).
We should just be able to return;
`return set(sessionInfo.roles) - set(["Delegate to Collection Zone"])`

dlysak: @spopezen Updated

spopezen:
Reverts https://github.com/zenoss/zenoss-prodbin/pull/3809

PR #3818, opened 2020-03-03
ZEN-32038 Made rabbitmq.sh executable

DeKapito:


DeKapito:
@joshw No, on develop everything is fine

PR #3821, opened 2020-03-16
ZEN-32629 allow monitoring MariaDB using custom passwords

DeKapito:


bin/metrics/mysqlmetrics.sh
```diff
@@ -1,6 +1,13 @@
 #!/bin/bash
 
-mysql -s --skip-column-names -B <<SQL
+if [ $1 = "mariadb-model" ]
+  then
+     DB="zodb"
```
ISPHOST: please fix white spaces in this line.

PR #3832, opened 2020-03-27
ZEN-32596 Upgrade fails if custom database passwords are in use

romanbakaleyko:


romanbakaleyko:
it also requires to prepare a doc to run migration script by hand before upgrade or reset password, this will be only the case for upgrades prior to 6.5.0, for upgrades from 6.5.0 to further versions extra step is not required.

PR #3836, opened 2020-04-01
Add the zen_schema_tags metadata to the ZING data maps

rphillipsz:
https://jira.zenoss.com/browse/ZING-8700

In order to properly type devices and device components, we're adding a new metadata field to the ZING entities for devices and device components.  This tag will be used to add additional schemas to the entities in model-id.

### DIFF
--- a/Products/Zing/datamaps.py
+++ b/Products/Zing/datamaps.py
@@ -317,13 +317,12 @@ class ZingDatamapHandler(object):
         f.metadata[ZFact.DimensionKeys.META_TYPE_KEY] = om_context.meta_type
         f.data[ZFact.MetadataKeys.NAME_KEY] = om_context.name
 
-        if om_context.is_device:
+        if om_context.is_device_component:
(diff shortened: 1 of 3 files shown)


Products/Zing/datamaps.py
```diff
@@ -313,9 +318,13 @@ def apply_extra_fields(self, om_context, f):
         f.data[ZFact.MetadataKeys.NAME_KEY] = om_context.name
 
         if om_context.is_device:
+            f.data[ZFact.MetadataKeys.ZEN_SCHEMA_TAGS_KEY] = "Device"
```
bharatZen, line 321: Should we check to make sure it is not also a DeviceComponent? If the latter, we should only add DeviceComponent to the schema-tags, not Device

rphillipsz: It would only happen if the object was an instance of both Device and DeviceComponent. Is that possible?  If so, yes we should I think.

rphillipsz: The next commit handles this, to ensure it won't get both schemas from this addition.

PR #3843, opened 2020-04-22
ZEN-32713 Need option to disable adding maintenance windows at the root device class

romanbakaleyko:


romanbakaleyko:
backport of https://github.com/zenoss/zenoss-prodbin/pull/3825 and https://github.com/zenoss/zenoss-prodbin/pull/3831

PR #3845, opened 2020-04-24
Add support for RelStorage 3.0.

jpeacock-zenoss:
Fixes ZEN-32709.


romanbakaleyko:
old style repr is back!

PR #3853, opened 2020-05-08
ZEN-32772 Zenpack installation fails with 'FATA[0408] Unable to commi…

romanbakaleyko:
…t container' error

### DIFF
--- a/bin/quiesce-mariadb.sh
+++ b/bin/quiesce-mariadb.sh
@@ -66,8 +66,6 @@ function getMysqlConnectionArgs()
     local arg
     arg+=" "$(getConfValue "$config" "$dbname-admin-user" "--user=")
     arg+=" "$(getConfValue "$config" "$dbname-admin-password" "--password=")
-    arg+=" "$(getConfValue "$config" "$dbname-host" "--host=")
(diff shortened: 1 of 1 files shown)


bin/quiesce-mariadb.sh
```diff
@@ -66,8 +66,6 @@ function getMysqlConnectionArgs()
     local arg
     arg+=" "$(getConfValue "$config" "$dbname-admin-user" "--user=")
     arg+=" "$(getConfValue "$config" "$dbname-admin-password" "--password=")
-    arg+=" "$(getConfValue "$config" "$dbname-host" "--host=")
-    arg+=" "$(getConfValue "$config" "$dbname-port" "--port=")
```
jpeacock-zenoss, line 70: The root user doesn't have a password, so I think the admin-password option should be excluded.

jpeacock-zenoss: If the script is smart enough to exclude the `--password` option if the admin password is blank, then it's fine.

romanbakaleyko: it works even if you put empty password, I checked that before pushing changes

PR #3854, opened 2020-05-12
ZEN-32752 zentrap fails silently while processing traps

romanbakaleyko:


romanbakaleyko:
fixing regression caused by https://github.com/zenoss/pynetsnmp/commit/be19c6123680b35f74a86ef4e0e5821da498f734#diff-9bc2b2dd326f0172ac610569d41f36dc

PR #3868, opened 2020-06-11
ZEN-32798 zenossdbpack doesn't work -- regression after update relsto…

romanbakaleyko:
…rage

### DIFF
--- a/Products/ZenUtils/patches/mysqladaptermonkey.py
+++ b/Products/ZenUtils/patches/mysqladaptermonkey.py
@@ -19,6 +19,7 @@ import logging
 import traceback
 import time
 from Products.ZenUtils.Utils import monkeypatch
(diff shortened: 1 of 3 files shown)


romanbakaleyko:
For now running zenossdbpack with workers doesn't work, I need to create a ticket for that. Along with that atching of relsotorage to monitor connections by adding records for each connection to connection_info is also not working properly as we populate connection_connection info using original connection and in the new relstorage it can be read only what raises exception when attempt to update table via such connection. I catch that exception for now and going to create a ticket for that as well.

romanbakaleyko:
Regarding how I verified what is done, I run zenossdbpack in legacy mode and without it and compared results, for the following scenarios I got same results

zenossdbpack -d1 
zenossdbpack -d1 -t
zenossdbpack -d1 -m

PR #3869, opened 2020-06-12
ZEN-32882 Maintenance windows for ComponentGroups can leave component…

romanbakaleyko:
…s in incorrect production state


romanbakaleyko:
back port https://github.com/zenoss/zenoss-prodbin/pull/3743/files

PR #3873, opened 2020-06-23
ZEN-32662 Cloud ZenOperator users cannot Ack/Close Events from Event …

romanbakaleyko:
…Console

### DIFF
--- a/Products/ZenUI3/browser/resources/js/zenoss/eventDetail.js
+++ b/Products/ZenUI3/browser/resources/js/zenoss/eventDetail.js
@@ -467,6 +467,7 @@ Ext.onReady(function() {
             this.addSection(eventDetailsSection);
 
             this.checkCustomizations();
+            Zenoss.Security.onPermissionsChange(this.updateEventActions, this);
         },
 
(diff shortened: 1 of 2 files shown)


Products/ZenUI3/browser/resources/js/zenoss/eventDetail.js
```diff
@@ -467,6 +467,7 @@ Ext.onReady(function() {
             this.addSection(eventDetailsSection);
 
             this.checkCustomizations();
+            Zenoss.Security.onPermissionsChange(this.updateEventActions, this);
```
jpeacock-zenoss, line 470: I don't understand what's accomplished by using `updateEventActions` as a callback when `Zenoss.Security.setContext` is called (the `onPermissionsChange` callback is called when `setContext` is called).

It seems like the `update` function would handle all the needed permission settings.

romanbakaleyko: setContext pulls permissions from the backend, I need to make sure I get permissions before I call updateEventActions

PR #3874, opened 2020-06-25
ZEN-32661 Cloud ZenOperator users cannot run commands on Events

romanbakaleyko:


romanbakaleyko:
make common permissions for CommandView to be able to run commands by user without global role, check permissions inside CommandView

PR #3877, opened 2020-07-03
zencommand handles command timeouts.

jpeacock-zenoss:
When a command's execution time exceeds the timeout value, the connection to the device will be closed.  The connection is closed when the collection cycle has completed.

Fixes ZEN-32226.

### DIFF
--- a/Products/ZenRRD/runner.py
+++ b/Products/ZenRRD/runner.py
@@ -7,6 +7,8 @@
 #
 ##############################################################################
 
+from __future__ import absolute_import
+
 import logging
 
 from twisted.internet import reactor, defer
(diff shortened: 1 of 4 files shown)


jpeacock-zenoss:
retest this

jpeacock-zenoss:
please test this.

Products/ZenRRD/runner.py
```diff
@@ -166,26 +170,28 @@ def processEnded(self, reason):
...
-        d, self.command_defer = self.command_defer, None
+        d, self._deferred = self._deferred, None
         if not d.called:
             d.callback(self)
 
 
+@implementer(IRunner)
 class SshRunner(object):
-    implements(IRunner)
+    """Execute commands on the remote end of an SSH connection."""
```
jeff-powell, line 186: This looked like an important change, implements() to @implementer(), what is the difference?

jpeacock-zenoss: They both accomplish the same thing.

`implements` is old-skool.  However, it implies "magic" because it somehow just knows what the class is.

`@implementer` is a class decorator and is the idiomatic way to apply changes to a class.  It also signals to anyone reading the code that something's being done to the class.  The `implements` function is just a function call and signals confusion to anyone unfamiliar with zope.interface APIs.

Products/ZenUtils/Executor.py
```diff
@@ -1,84 +1,392 @@
...
+
+    Setting the DynamicDeferredSemaphore to a zero limit effectively
+    creates a semaphore with no limit.
+
+    @ivar limit: The total number of tokens available for acquisition.
+    @type limit: int
+
+    @ivar tokens: The number of tokens available for acquisition.
+    @type tokens: int
     """
```
jeff-powell, line 318: Might want to expand the comments to reference the origin of this semaphore

PR #3878, opened 2020-07-06
ZEN-32847 Component / Device fields are not always reflected in cloud via zing-connector

dlysak:
These changes also help to solve the issue in https://jira.zenoss.com/browse/ZEN-32825


Products/Zing/tests/test_datamaps.py
```diff
@@ -187,6 +177,7 @@ def test_generate_facts_handles_missing_parent_uuid(t, subscribers):
             components.append(
                 Mock(getUUID=Mock(return_value=uuid), getComponentGroupNames=Mock(return_value='getComponentGroupNames')))
```
StevePC, line 178: making the components have the same uuid as the device makes this test invalid, as you can't tell if we recorded the correct uuid or not.  having all components return the same set of component groups also makes this invalid as we can't tell if we are correctly recording the component groups for individual components.

Products/Zing/tests/test_datamaps.py
```diff
@@ -383,6 +380,17 @@ def getDeviceClassName(self):
     def getProductionStateString(self):
         return 'test_production_state_string'
 
+    def getPrimaryParent(self):
```
StevePC, line 383: this doesn't allow testing of nested components, or testing of handling relations on the 'os' and 'hw' attributes of devices.

Products/Zing/tests/test_datamaps.py
```diff
@@ -47,6 +47,7 @@ def setUp(t):
             isLockedFromUpdates=Mock(return_value=False),
             isLockedFromDeletion=Mock(return_value=False),
         )
+        t.target.device.return_value = t.target
```
StevePC, line 50: We need test cases for datamaps that are against components directly, not just for datamaps against devices.  We also need test cases for datamaps against components that are children of components.  We aren't testing much of the functionality here.

PR #3879, opened 2020-07-08
Remove relative import statement.

jpeacock-zenoss:
Apparently, relative imports do not work in modules that are not imported using absolute import notation.

Fixes ZEN-32923.


jpeacock-zenoss:
Don't worry about the build failure.  It's due service migration script test failing and has nothing to do with the issue this PR is addressing.

PR #3885, opened 2020-07-14
Refactor zencommand's event management.

jpeacock-zenoss:
Add Clear events, when necessary, for a datasource during result processing, not during result storage.  Also, datasource timeout events have the component field set to the name of the datasource.

Fixes ZEN-32925.

### DIFF
--- a/Products/ZenRRD/zencommand.py
+++ b/Products/ZenRRD/zencommand.py
@@ -45,8 +45,7 @@ from Products.ZenCollector.tasks import (
     SubConfigurationTaskSplitter,
     TaskStates,
 )
-from Products.ZenEvents import Event
-from Products.ZenEvents.ZenEventClasses import Clear, Cmd_Fail
+from Products.ZenEvents.ZenEventClasses import Clear, Cmd_Fail, Error, Info
 from Products.ZenRRD import runner
 from Products.ZenRRD.CommandParser import ParsedResults
 from Products.ZenUtils.Executor import makeExecutor
@@ -459,7 +458,7 @@ class SshPerformanceCollectionTask(BaseTask):
         try:
             if not self._manageIp and self._useSsh:
                 self._eventService.sendEvent(
-                    self.manage_ip_event, severity=Event.Info
+                    self.manage_ip_event, severity=Info
                 )
             else:
                 self._eventService.sendEvent(
@@ -505,7 +504,7 @@ class SshPerformanceCollectionTask(BaseTask):
                 device=self._devId,
                 summary=e.message,
                 component=COLLECTOR_NAME,
-                severity=Event.Error,
+                severity=Error,
             )
             raise
         else:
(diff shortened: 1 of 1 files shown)


Products/ZenRRD/zencommand.py
```diff
@@ -947,21 +932,25 @@ def makeCmdTimeoutEvent(deviceId, datasource, severity=None):
         deviceId,
         datasource,
         "Datasource %s command timed out" % datasource.name,
-        severity=severity,
+        severity=Error if severity is None else severity,
         eventKey="CommandTimeout",
+        component=datasource.name,
```
StevePC, line 937: worried about re-using component here to represent the datasource, as it is possible for a device-level datasource to have the same name as an actual device component.  Could pass "" instead to have this be a device-level event for a component template.

Products/ZenRRD/zencommand.py
```diff
@@ -657,20 +657,22 @@ def _parse_error(self, datasources, failure):
...
-                results.append((ds, failure, parsedResults))
-            return results
+            results = [(ds, ParsedResults()) for ds in datasources]
+        else:
+            log.error(
+                "Command failed  device=%s datasource=%s error=%s reason=%s",
+                self._devId, datasources[0].name, failure.type, failure.value,
+            )
+            results = [(ds, ParsedResults()) for ds in datasources]
+            severity = Clear
```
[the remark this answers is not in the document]
jpeacock-zenoss: I have avoided adding new events.  If the released code didn't send an event for a particular error, these changes will continue that behavior.

The only behavior change I did make was the one requested in the ticket: timeout events for one datasource do not get cleared by the timeout clear events of another datasource.

Products/ZenRRD/zencommand.py
jpeacock-zenoss: Having the component field be empty is the reason why timeout events are getting cleared.  Two different datasources will fire events that look identical.  The event identity is based on the device, event class, component, and event key.  For all the datasources, the device, event class, and event key are identical.  The only way to differentiate is to use the datasource for the component.

Products/ZenRRD/zencommand.py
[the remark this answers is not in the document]
StevePC: The code didn't send an event before because users had access to logs.  With cloud, we need to be *much* more proactive about sending events for things that impact users and that users might be able to do something about.  Even without that argument, our code is far from perfect.  Assuming it is doing (or not doing) something for a valid reason we can't see (or find after some reasonable investigation) as an excuse to not fix a valid issue we do see is a bad idea, IMHO.

Products/ZenRRD/zencommand.py
StevePC: Since you are setting a specific event key for timeouts, if you need to differentiate one datasource's timeout from another's, just prepend the timeout event key with the datasource?

jpeacock-zenoss: I can do that.

Looking at other examples of event key values, I see:

interval_exceeded
snmp_v3_error
snmp_error
partial_oids_collected
no_oid_results
agent_down
cycle_interval
correlation_execution
nmap_execution
ping_lag
transform_failed
transform_disabled
zCommandUsername

so why not \<datasource>\_timeout (or timeout\_\<datasource>).

Products/ZenRRD/zencommand.py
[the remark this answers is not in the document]
jpeacock-zenoss: This sounds like a bigger issue than addressing ssh timeout issues.  It just not zencommand that would need these changes.

I think adding additional events to signal different collection error situations should probably go into a different ticket.

PR #3890, opened 2020-07-21
ZEN-32836 Application Audit Log Missing Initial Login for Brightstar

romanbakaleyko:


romanbakaleyko:
In 7x world we don't have login page, all authentication performs in auth0, that is why code we have in patched validate method doesn't work as login page was a marker to identify when customer logs in. Therefor I don't see a reason to keep validate patched, I removed it and put audit logging in the place where I think it should be.

PR #3892, opened 2020-07-30
ZEN-32956 Cloud ZenOperatpor users can't create an event from Event c…

romanbakaleyko:
…onsole while can on device level


romanbakaleyko:
Check whether we are in device level context and if not disable button for creating of events if the user has no global role. Per conversation with Rajan decided that if user doesn't have global role he should create events from the device level context he has access to.

PR #3902, opened 2020-09-08
ZEN-33021 ZenOperator can save device Priority when changing Prod. State

romanbakaleyko:


romanbakaleyko:
* disabled rows on the device overview user doesn't have access to
* check role explicitly for changing production state
* set context in the device grid for user without global role
* removed some code in the Action.js as it is broken and causes a problem because once externallyDisabled is set to true there is no way to set it back to false. Since I couldn't figure out what role this thing plays instead of fixing it I removed it.

Did some some and for me things work fine.

Products/ZenUI3/browser/resources/js/zenoss/Action.js
```diff
@@ -22,9 +21,6 @@
             if (!this.permitted) {
                 return true;
             }
-            if (this.externallyDisabled) {
-                return true;
-            }
             if (!this.permitted && !this.filtered) {
```
jpeacock-zenoss, line 24: This `if` statement will never succeed or never execute because the `if` statement on line 21 will return when `this.permitted` is false.

jpeacock-zenoss (changes requested):

jpeacock-zenoss (approved):

PR #3903, opened 2020-09-09
Fix log formatter context usage.

jpeacock-zenoss:
The log formatter context must be called when used in a `with` statement.

Fixes ZEN-33032.

### DIFF
--- a/Products/Jobber/tests/test_logformattercontext.py
+++ b/Products/Jobber/tests/test_logformattercontext.py
@@ -1,0 +1,49 @@
+##############################################################################
+#
+# Copyright (C) Zenoss, Inc. 2019, all rights reserved.
+#
+# This content is made available according to terms specified in
+# License.zenoss under the directory where your Zenoss product is installed.
+#
+##############################################################################
+
+from __future__ import absolute_import, print_function
+
+import logging
+
+from unittest import TestCase
+
+# from .utils import subTest
+from ..jobs.subprocess import (
+    LogFormatterContext, getLogFormattingContext, null_context,
+)
+from ..utils.log import TaskLogFileHandler
+from .utils import LoggingLayer
+
+
+class LogFormatterContextTest(TestCase):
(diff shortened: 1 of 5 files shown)


Products/Jobber/tests/test_logformattercontext.py
```diff
@@ -0,0 +1,49 @@
...
+#
+##############################################################################
+
+from __future__ import absolute_import, print_function
+
+import logging
+
+from unittest import TestCase
+
+# from .utils import subTest
```
romanbakaleyko, line 16: why do we need this commented import ?

romanbakaleyko (changes requested):

Products/Jobber/tests/test_logformattercontext.py
jpeacock-zenoss: we don't need it.

Products/Jobber/tests/test_logformattercontext.py
```diff
@@ -0,0 +1,48 @@
+##############################################################################
+#
+# Copyright (C) Zenoss, Inc. 2019, all rights reserved.
+#
+# This content is made available according to terms specified in
+# License.zenoss under the directory where your Zenoss product is installed.
+#
+##############################################################################
+
+from __future__ import absolute_import, print_function
```
romanbakaleyko, line 10: I don't see we use print anywhere, do we need print_function ?

romanbakaleyko (changes requested):

Products/Jobber/tests/test_logformattercontext.py
jpeacock-zenoss: The print function was used when I was debugging the code.

It's OK to leave the `print_function` reference in place.  It's not actually importing anything since __future__ is a built-in.  My linter (flake8) is not complaining.

romanbakaleyko (approved):

PR #3907, opened 2020-09-14
added method

sockterrier:


jpeacock-zenoss (approved):
I'm not convinced that the device router is the best place for this API, but it's better than all the other routes.

PR #3913, opened 2020-09-21
ZEN-33051 'Send event' form is missing values for event class field f…

romanbakaleyko:
…or a user without global role.

### DIFF
--- a/Products/ZenUI3/browser/resources/js/zenoss/EventPanels.js
+++ b/Products/ZenUI3/browser/resources/js/zenoss/EventPanels.js
@@ -30,6 +30,10 @@
         var device;
         if (Zenoss.env.device_uid) {
             device = Zenoss.env.device_uid.split("/").reverse()[0];
+        } else {
+            if (!_has_global_roles()) {
+                device = Ext.getCmp(gridId).getSelectionModel().getSelected().data.device.text
+            }
         }
         var collectors = new Ext.data.ArrayStore({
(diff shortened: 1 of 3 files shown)


romanbakaleyko:
fixed behavior:
* list of event classes names is available for user without global role, fixes ZEN-33051 and  ZEN-32957
* 'add event' button is disabled in the event console context by default. For user without global role it becomes active after clicking on any event If such user has "Manage Event" permission for a device the event belongs to. Add event form will have filled device field automatically without permission to change it (field is read only)
* on a device context for user without global role 'add event' button is enabled if such user has a device as administrative object with role which has "Manage Event" permissions, in this case it works the same as in event console context, field 'device' is read only and will be autofilled with device name.

jpeacock-zenoss (approved):

Products/ZenUI3/browser/resources/js/zenoss/EventPanels.js
```diff
@@ -1146,7 +1151,8 @@
                             });
                         }
                         // Disable add_event button for event console context if no global role
-                        if (!Zenoss.env.device_uid) {
+                        // In some contexts we don't have add event button so check it first
+                        if (!Zenoss.env.device_uid && tbar && tbar.getComponent('add_event_main_button')) {
```
StevePC, line 1155: more efficient to do getComponent once to a local variable

PR #3916, opened 2020-09-22
ZEN-33049 Can't delete a Group on the INFRASTRUCTURE page

romanbakaleyko:


romanbakaleyko:
the issue is a regression after this fix  https://github.com/zenoss/zenoss-prodbin/commit/1b9001f96070096b09b534812f553454a416a03d which had to be a special case only for Device Class organizer but appeared to be for any contexts. Later it was fixed for "Dynamic Service" organizer, now it is working as expected: special handling for Device Class and default behavior for everything else.

Deer-WarLord (approved):

PR #3928, opened 2020-10-07
ZEN-33097 Can't create an event using a user without global role (Zen…

DeKapito:
…Operator)

### DIFF
--- a/Products/Zuul/routers/zep.py
+++ b/Products/Zuul/routers/zep.py
@@ -648,6 +648,17 @@ class EventsRouter(DirectRouter):
             return False
         manage_events_for = (r.managedObjectName() for r in user.getAllAdminRoles() if r.role in READ_WRITE_ROLES)
         return organizer_name in manage_events_for
+    
+    def add_events(self, summary, device, component, severity, evclasskey,
+                  evclass=None, monitor=None, **kwargs):
+        ctx = self.context.dmd.Devices.findDevice(device.strip())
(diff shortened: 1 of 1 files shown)


Products/Zuul/routers/zep.py
```diff
@@ -648,6 +648,17 @@ def manage_events(self, evids=None, excludeIds=None, params=None, uid=None, asof
...
+    def add_events(self, summary, device, component, severity, evclasskey,
+                  evclass=None, monitor=None, **kwargs):
+        ctx = self.context.dmd.Devices.findDevice(device.strip())
+        if not ctx:
+            ctx = self.context
+
+        if Zuul.checkPermission(ZEN_MANAGE_EVENTS, ctx):
+            return True
+        else:
+            return False
```
jpeacock-zenoss, line 661: While using an `if` statement is explicit, I believe it is redundant because `Zuul.checkPermission` already returns a boolean (or boolean like) value. So the return can be written more succinctly as:
```
return Zuul.checkPermission(ZEN_MANAGE_EVENTS, ctx)
```

jpeacock-zenoss (changes requested):

Products/Zuul/routers/zep.py
DeKapito: Yeah, it is a little bit stupid

jpeacock-zenoss (approved):

PR #3931, opened 2020-10-08
Expire jobs based on status.

jpeacock-zenoss:
Fixes ZEN-33101.

### DIFF
--- a/Products/Jobber/tests/test_storage.py
+++ b/Products/Jobber/tests/test_storage.py
@@ -374,6 +374,13 @@ class ExpireKeysTest(TestCase):
         t.assertIsNotNone(ttl)
         t.assertLess(ttl, t.expires)
 
+    def test_ttl_for_aborted(t):
+        t.store.update(t.jobid, status="ABORTED")
+        time.sleep(1.0)
+        ttl = t.store.ttl(t.jobid)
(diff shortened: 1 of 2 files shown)


Products/Jobber/tests/test_storage.py
```diff
@@ -307,6 +308,73 @@ def test_update_with_only_Nones(t):
...
+        t.assertIsNotNone(ttl)
+        t.assertLess(ttl, t.expires)
+
+    def test_ttl_for_failure(t):
+        t.store.update(t.jobid, status="FAILURE")
+        time.sleep(1.0)
+        ttl = t.store.ttl(t.jobid)
+        t.assertIsNotNone(ttl)
+        t.assertLess(ttl, t.expires)
+
```
StevePC, line 376: Should have a test for "ABORTED" as well, as that was the specific ask in the Jira

StevePC (changes requested):

Products/Jobber/tests/test_storage.py
jpeacock-zenoss: You are right, I should.  The whole reason why I was making this change in the first place.

StevePC (approved):

PR #3940, opened 2020-10-16
ZEN-32912 zentrap fails silently while processing traps

romanbakaleyko:


romanbakaleyko:
This is forward port 
https://github.com/zenoss/zenoss-prodbin/pull/3854/files
https://github.com/zenoss/zenoss-prodbin/pull/3866/files
Assembly is already updated to use recent version of pynetsnmp

jpeacock-zenoss (approved):

PR #3943, opened 2020-10-19
JobManager now mimics transactions.

jpeacock-zenoss:
When submitting jobs via JobManager, jobs are not submitted to zenjobs until the current database transaction is committed.  JobManager's other methods for retrieving jobs should return jobs that have been submitted, but not committed yet.

Fixes ZEN-33084.


StevePC (approved):

jpeacock-zenoss:
This PR was missing features.

PR #3945, opened 2020-10-20
ZEN-32653 Process memory-threshold patch should be applied to prodbin

romanbakaleyko:


jpeacock-zenoss (changes requested):
Since this is adding a pair of new attributes to ZODB objects.  There should probably a migration script to add these in on upgrades.

romanbakaleyko:
> Since this is adding a pair of new attributes to ZODB objects. There should probably a migration script to add these in on upgrades.

It is not required, if there is no attribute for a particular OSProcess it takes value from OSProcessClass

PR #3950, opened 2020-10-22
ZEN-31950 revert patch which was added in order to find the root caus…

romanbakaleyko:
…e of the problem in scope of ZEN-32484


romanbakaleyko:
revert logging we added in order to find the root cause of the ZEN-31950 as it simply useless and doesn't really help. https://github.com/zenoss/zenoss-prodbin/commit/76aae18809522dd946101ad2ebbe3c91167a3cac

DeKapito (approved):

PR #3958, opened 2020-10-27
ZEN-33081:Tracebacks 'Bad task  executor 'in zencommand log for all linux devices

Deer-WarLord:


Products/ZenRRD/zencommand.py
```diff
@@ -492,21 +492,23 @@ def doTask(self):
...
+                log.error(err_msg)
             self._eventService.sendEvent(
                 STATUS_EVENT,
                 device=self._devId,
                 summary=e.message,
                 component=COLLECTOR_NAME,
                 severity=Error,
             )
-            raise
+            defer.fail(e)
```
jpeacock-zenoss, line 511: This can be removed.  All `defer.fail` does is returned a `Deferred` object with its errback called, with `e` as the argument passed to the errback.

jpeacock-zenoss (changes requested):

jpeacock-zenoss (approved):

PR #3961, opened 2020-10-28
ZEN-32255 Add varbindCopyMode option for configuring a way of process…

DeKapito:
…ing varbinds

```
Assume we have the following varbinds:
   someVar.0  Data0
   someVar.1  Data1
Possible copy modes:
0 - the varbinds are copied into event as one
    field. Expected event fields:
        someVar:         Data0,Data1
        someVar.ifIndex: 0,1
1 - the varbinds are copied into event as several
    fields and sequence field is added.
    Expected event fields:
        someVar.0:        Data0
        someVar.1:        Data1
        someVar.sequence: 0,1
2 - the mixed mode. Uses varbindCopyMode=0 behaviour
    if there is only one occurrence of the varbind,
    otherwise uses varbindCopyMode=1 behaviour
```


Products/ZenModel/migrate/addVarbindCopyMode.py
```diff
@@ -0,0 +1,76 @@
...
+                    continue
+                log.info("Appending varbindCopyMode option to %s for '%s'.", current_config, zentrap.name)
+                configfile.content += update_string
+                commit = True
+
+        if commit:
+            ctx.commit()
+
+
+AddVarbindCopyMode()
```
jpeacock-zenoss, line 76: Please implement this service migration in the zenoss-service repository.  Implementing service migrations using the Zenoss Products.ZenModel.Migrate framework is deprecated.

jpeacock-zenoss (changes requested):

Products/ZenEvents/zentrap.py
```diff
@@ -577,9 +600,15 @@ def _process_varbinds(self, varbinds):
...
+                if self.varbindCopyMode == LEGACY_VARBIND_COPY_MODE:
+                    result[base_name].append(value)
+                    if suffix:
+                        result[base_name + ".ifIndex"].append(suffix)
+                else:
+                    result[full_name].append(value)
+                    if suffix:
+                        result[base_name + ".sequence"].append(suffix)          
 
         return {name: ','.join(vals) for name, vals in result.iteritems()}
```
jpeacock-zenoss, line 613: Rather than mixing the various copy modes into the same method, it'll be more clear to future readers of this code if the different copy modes were implemented as distinct functions.  To avoid testing the `varbindCopyMode` variable multiple times, each copy mode deserves its own class which is selected and initialized when zentrap starts.

Here's what the "legacy" class would look like:
```
class _LegacyVarbindProcessor(object):

    def __init__(self, oid2name):
        self.oid2name = oid2name

    def __call__(self, varbinds):
        result = defaultdict(list)
        for oid, value in varbinds:
            base_name = self.oid2name(oid, exactMatch=False, strip=True)
            full_name = self.oid2name(oid, exactMatch=False, strip=False)
            if base_name != full_name:
                result[base_name].append(value)
                suffix = full_name[len(base_name) + 1:]
                result[base_name + ".ifIndex"].append(suffix)
            else:
                result[full_name].append(value)
        return result
```
This is what the "direct" mode class would look like:
```
class _DirectVarbindProcessor(object):

    def __init__(self, oid2name):
        self.oid2name = oid2name

    def __call__(self, varbinds):
        result = defaultdict(list)
        for oid, value in varbinds:
            base_name = self.oid2name(oid, exactMatch=False, strip=True)
            full_name = self.oid2name(oid, exactMatch=False, strip=False)
            result[full_name].append(value)
            if base_name != full_name:
                suffix = full_name[len(base_name) + 1:]
                result[base_name + ".sequence"].append(suffix)
        return result
```
The "mixed" mode class will be similar and its `__call__` method I'll leave as an exercise for you.
```
class _MixedVarbindProcessor(object):

    def __init__(self, oid2name):
        self.oid2name = oid2name

    def __call__(self, varbinds):
        ...
```
To select the appropriate varbind processor, I suggest the following:
```
# defined as a module or TrapTask class variable
_varbind_processors = {
    LEGACY_VARBIND_COPY_MODE: _LegacyVarbindProcessor.
    DIRECT_VARBIND_COPY_MODE: _DirectVarbindProcessor,
    MIXED_VARBIND_COPY_MODE: _MixedVarbindProcessor,
}
```
Then, somewhere in the `TrapTask`'s `__init__` method:
```
    processor_class = _varbind_processors.get(self.varbindCopyMode)
    self._process_varbinds = processor_class(self.oid2name)
```

jpeacock-zenoss (changes requested):

jpeacock-zenoss (approved):

PR #3962, opened 2020-11-02
ZEN-30566: Unable to add user command

Deer-WarLord:


Products/ZenModel/migrate/UpdateRolesForDefineCommands.py
```diff
@@ -0,0 +1,38 @@
...
+    CZ_ADMIN_ROLE,
+    MANAGER_ROLE,
+    ZEN_MANAGER_ROLE,
+    ZEN_DEFINE_COMMANDS_EDIT
+)
+from Products.ZenModel.migrate.UpdateDefineCommandsEditPermission import UpdateDefineCommandsEditPermission
+from Products.ZenModel.ZMigrateVersion import SCHEMA_MAJOR, SCHEMA_MINOR, SCHEMA_REVISION
+
+
+class UpdateRolesForDefineCommands(Migrate.Step, UpdateDefineCommandsEditPermission):
```
jpeacock-zenoss, line 28: The `Migrate.Step` class may be doing something unconventional that prevents more than one class inheriting from it in the same class hierarchy.

You could try fixing `Migrate.Step` (or the metaclass it specifies), but an alternative fix is to not inherit from `UpdateDefineCommandsEditPermission` class and invoke the `addPermission` method manually, e.g.
```
    UpdateDefineCommandsEditPermission.addPermission(
        self,
        dmd.zport,
        ZEN_DEFINE_COMMANDS_EDIT,
        [MANAGER_ROLE, CZ_ADMIN_ROLE, ZEN_MANAGER_ROLE],
        1,
    )
```

jpeacock-zenoss (changes requested):

Products/ZenModel/migrate/UpdateRolesForDefineCommands.py
Deer-WarLord: fixed using alternative approach

jpeacock-zenoss (approved):

PR #3965, opened 2020-11-05
ZEN-33157 remove option to migrate perf data during device reidentifying

romanbakaleyko:


romanbakaleyko:
for the moment this functionality is broken and causes only problems, remove it until we make it work.

jpeacock-zenoss (approved):

PR #3966, opened 2020-11-10
job_end should verify the job has ended.

jpeacock-zenoss:
A job that is set to RETRY is not ended, but will run again later.

Fixes ZEN-33176.


lundybernard (approved):
looks good

PR #3980, opened 2020-11-19
ZEN-33016 New zenhubworker service for flexible per-customer tuning

romanbakaleyko:


Products/ZenHub/server/config.py
```diff
@@ -59,6 +61,34 @@ def xmlrpcport(self):
...
+    def __init__(self, config):
+
+        self.__config = {}
+
+        try:
+            with open(config, 'r') as f:
+                self.__config = json.load(f)
+        except Exception as e:
+            import logging
+            log = logging.getLogger("zen.%s" %  __name__)
```
jpeacock-zenoss, line 76: Please name this logger "zen.zenhub.server.config" to stay consistent with the other loggers defined in zenhub, e.g.
```
    log = logging.getLogger("zen.zenhub.server.config")
```
The `__name__` variable will return "Products.ZenHub.server.config" which is not the desired name.

jpeacock-zenoss (changes requested):

jpeacock-zenoss (changes requested):
I think the configuration file for specifying routes is too complicated.  Configuration files are just as much an API of a service as methods are for a class.

I have written an alternate configuration file for configuring the executors, pools, and routes.  It groups details about a particular executor together in one place.  The alternative format also removes redundant information. I opened a PR on this PR here: https://github.com/zenoss/zenoss-prodbin/pull/3982.

Please take a look.

jpeacock-zenoss (changes requested):
There should be unit tests to cover this new code.

Products/ZenHub/server/config.py
romanbakaleyko: done

jpeacock-zenoss (approved):

PR #3983, opened 2020-11-20
ZEN-33009: fix object comparison in ADM

romanbakaleyko:
(cherry picked from commit 6c1964ae10aab499ae78c058cf57d9371bde30dc)


romanbakaleyko:
the issue is fixed in prod environments by a patch and has to be merged in to develop to not roll back after upgrade.

romanbakaleyko:
oops, looking what is wrong with tests.

romanbakaleyko:
hmmm, locally I don't see the problem, investigating.

StevePC (approved):

PR #3985, opened 2020-11-24
ZEN-33165 adding and searching for IpAddresses/networks can fail due …

romanbakaleyko:
…to improper cache management


romanbakaleyko:
this is a forward port of changes that were patched on 7.0.15 https://github.com/Zenoss/zenoss-prodbin/commit/7e9e9cc356f1eea1aed35c8eab6083011b8ba47c so we don't loose them after upgrade

DeKapito (approved):

PR #3990, opened 2020-11-26
ZEN-33213 Flare message 'TypeError: Cache values must be persistent o…

romanbakaleyko:
…bjects' after Upgrade (Jobs list is blank)


romanbakaleyko:
I thought zope throws an exception when I am trying to get broken old JobManager but that is not true it logs an error but return some wrapper object, so changed migration to check for a new attribute in a new JobManager, if there is no new attr then remove JobManager and add new one.

DeKapito (approved):

PR #3994, opened 2020-12-04
ZEN-33208: Error sending datamaps for netapp devices

Deer-WarLord:


Deer-WarLord:
Unfortunately, there's a bug in zenpack yaml declaration where for `ocum_event_ids` field we apply default type - string, while in code we form dictionary, so to exclude such errors in future fix provides an alternative way to check types, not via propertyMap info but using real variable via isinstance. 

https://github.com/zenoss/ZenPacks.zenoss.NetAppMonitor/blob/develop/ZenPacks/zenoss/NetAppMonitor/zenpack.yaml#L71

Products/Zing/fact.py
```diff
@@ -171,33 +171,27 @@ def device_info_fact(device):
...
-                    log.exception(
-                        "Unable to call property: %s. Exception %s",
-                        device.getProperty(propId),
-                        e,
-                    )
-            else:
-                value = device.getProperty(propId)
-            if value is None:
-                value = ""
+        value = None
```
sockterrier, line 177: in the old code, "device.isLocal(propId)" evaluating to false avoided calling callables, i think we should preserve that.

sockterrier (changes requested):

Products/Zing/fact.py
```diff
@@ -171,33 +171,27 @@ def device_info_fact(device):
...
+                log.exception(
+                    "Unable to call property: %s. Exception %s",
+                    device.getProperty(propId),
+                    e,
+                )
+        else:
+            value = device.getProperty(propId)
+        if value is None:
+            value = ""
+        if device.isLocal(propId) and isinstance(value, valid_types):
```
joshw, line 194: You mention the original problem was triggered by "non-scalar" types.  Was that only dict, or will we have problems with list, tuple, or set?   Do we need to convert tuple and set to list, for instance, or is that handled elsewhere when the fact is processed?

joshw (changes requested):

Products/Zing/fact.py
joshw: Looks like it is, just lower down (line 194)

Deer-WarLord: `device.isLocal(propId)` are preserved, please see the 194 line

Products/Zing/fact.py
Deer-WarLord: I believe the `lines` type that before assumed `list, set, tuple` as valid types for scalar arrays.

Deer-WarLord: Please see FactEncoder https://github.com/zenoss/zenoss-prodbin/blob/develop/Products/Zing/fact.py#L313-L338

Products/Zing/fact.py
```diff
@@ -171,33 +171,27 @@ def device_info_fact(device):
...
-                    )
-            else:
-                value = device.getProperty(propId)
-            if value is None:
-                value = ""
+        value = None
+        # Some of the device properties can be methods,
+        # so we have to call them and get values.
+        if callable(device.getProperty(propId)):
+            log.warn("Callable: %s", device.getProperty(propId))
```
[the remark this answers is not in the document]
Deer-WarLord: Changed to debug.

Products/Zing/fact.py
joshw: Yes, but now it doesn't check it until after calling callable properties.  I think he was suggesting that checking it sooner may avoid unnecessary calls.

Deer-WarLord: yes, that makes sense, fixed

sockterrier (approved):

joshw (approved):

PR #3999, opened 2020-12-14
ZEN-33185: Hide CZ Dashboard

Deer-WarLord:


romanbakaleyko (changes requested):
this change requires to update global.conf in the service template  + migration script

Deer-WarLord:
> this change requires to update global.conf in the service template + migration script

PR with added property for enabling/disabling Dashboard https://github.com/zenoss/zenoss-service/pull/701. 
This config doesn't have any impact on ZODB or other persistent data in the Zenoss system storages, because of that, the migration script is not required. This config edits zcml context for the started zope instances.

romanbakaleyko (approved):

PR #4012, opened 2020-12-30
ZEN-33269 Default device classes can be deleted

romanbakaleyko:


romanbakaleyko:
this is a backport the fix https://github.com/zenoss/zenoss-prodbin/commit/1b9001f96070096b09b534812f553454a416a03d with additional changes that came after original fix

DeKapito (approved):

PR #4047, opened 2021-01-28
ZEN-33398 Some Zenoss.resmgr pages are failing to render

romanbakaleyko:


romanbakaleyko:
make it work with new version of Zope2/zope.tales we updated to in 6.6

jpeacock-zenoss (approved):

PR #4052, opened 2021-02-05
ZEN-33415 don't forward zProps to zing

romanbakaleyko:


romanbakaleyko:
This regression caused by https://github.com/zenoss/zenoss-prodbin/pull/3878/files#diff-e3e284430f8f54c7093b528446d45b55f9eba2b136c7e251db706766e2ba361cR109, my guess is that we don't need send zProperties to zing.

jpeacock-zenoss:
I thought the intent was to exclude only the password fields from being sent to Cloud, not all the z-properties.

If all z-properties _should_ be excluded, maybe custom properties (c-properties) should be excluded as well?

romanbakaleyko:
asking @StevePC to participate this conversation to clarify as he submitted a ticket ZEN-32847 the fix causes behavior we see. Steve, please, share your opinion on it.

StevePC:
In previous meetings, it was decided that all model information would need to be sent to Zing, if we are ever going to be able to replicate the functionality of CZ in cloud.  For instance, many customers use values in cProperties to fill in fields for notifications and incident creation.  If this information is not going to be sent to cloud, we need to design an alternative method, as well as a migration path for all of this data, and implement it before we can deprecate pretty much any features in CZ.

StevePC (approved):

PR #4057, opened 2021-02-22
Added events to zenpackadapter

jeff-powell:
Add event processing to zenpackadapter as part of zminihub.
fixed a several issues causing some events and metrics to throw errors and be skipped
This addresses ZING-12577


webhookbotuser:

webhookbotuser:

webhookbotuser:

Products/ZenHub/PBDaemon.py
```diff
@@ -1210,6 +1217,12 @@ def buildOptions(self):
                                type='int',
                                help='Maximum number of events to queue')
 
+        self.parser.add_option('--maxqueueage',
```
joshw, line 1220: In general I try to keep changes to anything other than Products/ZenPackAdapter out of this branch.    I'd suggest getting this into develop (and merge it in here too), just to make it easier to keep things reconciled.

Products/ZenPackAdapter/utils/tales.py
```diff
@@ -49,11 +49,8 @@ def talesEval(express, context, extra=None, skipfails=False):
         msg = "Error when processing tales expression %s on context %s : Exception Class %s Message: %s" % (express,
                                                                                                             context,
                                                                                                             type(e), e)
-        if skipfails:
-            res = express
-            log.debug(msg)
-        else:
-            raise InvalidTalesException(msg)
+        res = express
+        log.warn(msg)
```
joshw, line 53: Is this still needed?  It looks like you're catching the InvalidTalesExpression and logging it elsewhere.

Products/ZenPackAdapter/zobject.py
```diff
@@ -595,12 +595,14 @@ def device(self):
         return self
 
     def getDeviceComponents(self, collector=None):
+        components = []
         for component in self.getSubComponents():
             if collector is not None:
                 if collector in getattr(component, 'collectors', []):
-                    yield component
+                    components.append(component)
```
joshw, line 602: not that I imagine it matters much, but why the change here?  this is slightly less efficient..

Products/ZenPackAdapter/zminievents.py
```diff
@@ -0,0 +1,51 @@
+##############################################################################
```
joshw, line 1: Is this used anywhere?

joshw (changes requested):

Products/ZenHub/PBDaemon.py
jeff-powell: I can isolate it - I'm just not sure whether or not they'd want an age-limit added to the event queue. I can ask Jason about this too.

Without these changes, two-thirds of the buffer has to be filled before the daemon will send.  (The default buffer size is 5000.)  Event if we set a low (via configs,) a few events could wait around a long time before the daemon sends them.

Products/ZenPackAdapter/utils/tales.py
jeff-powell: I couldn't get zpa or the EMC zenpack to load correctly until i removed the "raise InvalideTalesException(msg)"; the exception was causing the zenpack to fail loading.  Hence, I updated this segment to operate similar to how res_manager operates; returning the expression unmodified.

From there, I never really re-visited this since ZPA has been running smooth for me since (so I could work on events.)

Products/ZenPackAdapter/zminievents.py
jeff-powell: no, I should remove it.

Products/ZenPackAdapter/zobject.py
jeff-powell: when running zendmd on resmgr, a list with the components were provided
> dev1.getDeviceComponents()
[ compont_type_name_id_1,
compont_type_name_id_2,
etc,
] 

When running dmd on ZPA, the list is not generated, instead it's handle is returned
> dev1.getDeviceComponents()
(generator_object)

Products/ZenHub/PBDaemon.py
jpeacock-zenoss: This particular change looks reasonably isolated from the rest of the ZPA project to call it an enhancement for pushing events based on age (in addition to the queue size).

jeff-powell: Interesting, that 5 second loop makes sense though.
I wonder if that particular code isn't running in ZPA, because when I working on events, they never were sent until I added this

jpeacock-zenoss: The `pushEventsLoop` code should have been scheduled into the Twisted reactor when PBDaemon's `run` method is called.

jeff-powell: This is all correct now.  I updated monkey.py to correctly use and backed these changes to pbdaemon. 
`@monkeypatch(PBDaemon)
def pushEventsLoop(self):
    pass`

Products/ZenPackAdapter/zobject.py
joshw: Sure, that's typical of generators, but it's functionally equivalent (you can iterate on it, same as a list, or turn it into one  `list(dev1.getDeviceComponents())`for debugging)

joshw (approved):

PR #4075, opened 2021-04-14
not sending null values to zing-connector

sockterrier:


ecifreo (approved):

Products/Zing/fact.py
```diff
@@ -331,11 +331,11 @@ def _tweak_data(self, data_in):
                             "Casting it to str",
                             x.__class__,
                         )
-                        x = str(x)
+                        x = '' if x is None else str(x)
                     values.append(x)
                 data_out[k] = sorted(values)
             else:
-                data_out[k] = [v]
+                data_out[k] = [''] if v is None else [v]
```
StevePC, line 338: this line should probably follow the same logic as above in terms of what types we send bare and what gets converted to a str.  Just because it isn't in a list doesn't mean it might not be another type, like perhaps datetime.datetime.
Better to just wrap single values in a list at first and then process them all the same.

StevePC (changes requested):

PR #4084, opened 2021-05-27
ZEN-33612: add organizer field to organizer facts

cluther:
This is to allow organizer entities in cloud to be found when searching
for entities in the organizers.


cluther:
retest this please

joshw (approved):

PR #4085, opened 2021-06-09
Updating ZPA to use zobject

jeff-powell:
Fix for Zing-13679
https://jira.zenoss.com/browse/ZING-13679

### DIFF
--- a/Products/ZenPackAdapter/monkey.py
+++ b/Products/ZenPackAdapter/monkey.py
@@ -23,6 +23,7 @@ from Products.ZenHub.PBDaemon import PBDaemon
 from Products.ZenUtils.CmdBase import CmdBase
 from Products.ZenPackAdapter.cloudpublisher import CloudMetricPublisher
 
+import zenwrapt
 LOG = logging.getLogger("zen.monkeypatches")
 
 @monkeypatch(PBDaemon)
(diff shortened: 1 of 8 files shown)


webhookbotuser:

webhookbotuser:

webhookbotuser:

Products/ZenPackAdapter/monkey.py
```diff
@@ -23,6 +23,7 @@
 from Products.ZenUtils.CmdBase import CmdBase
 from Products.ZenPackAdapter.cloudpublisher import CloudMetricPublisher
 
+import zenwrapt
```
joshw, line 26: doesn't appear to be used

Products/ZenPackAdapter/services.py
```diff
@@ -68,6 +67,11 @@ def inner(*args, **kw):
     return inner
 
 
+def get(device=None, component=None):
```
joshw, line 70: probably doesn't need to be here.

joshw (changes requested):

Products/ZenPackAdapter/monkey.py
jeff-powell: zenwrapt auto-initialized when imported, so nothing more needed to be done. Originally, that code was included in monkey.py, but was separated out for readability and maintainability.

I update zenwrapt.py to have an initialize() method that is now called here and should be more intuitive.

Products/ZenPackAdapter/services.py
jeff-powell: removed - good catch
I think this was left over from when I was comparing the new zobject devices to the (previous) db object devices.

joshw (approved):

PR #4086, opened 2021-06-11
Add job to send organizer facts to zing.

jpeacock-zenoss:
Fixes ZEN-33421.

### DIFF
--- a/Products/Zing/send_organizers.py
+++ b/Products/Zing/send_organizers.py
@@ -1,0 +1,48 @@
+##############################################################################
+#
+# Copyright (C) Zenoss, Inc. 2021 all rights reserved.
+#
+# This content is made available according to terms specified in
+# License.zenoss under the directory where your Zenoss product is installed.
+#
+##############################################################################
+
+from __future__ import absolute_import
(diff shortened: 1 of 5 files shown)


Products/Zing/send_organizers.py
```diff
@@ -0,0 +1,48 @@
...
+@app.task(
+    bind=True,
+    base=requires(Abortable, DMD),
+    name="zen.zenjobs.zing.send_organizers",
+    summary="Send organizer facts to Zing",
+    ignore_result=False,
+)
+def send_organizers(self):
+    facts = {}
+    for root_name in ("Devices", "Groups", "Locations"):
```
StevePC, line 28: Systems?

StevePC (changes requested):

Products/Zing/send_organizers.py
jpeacock-zenoss: Right.  I knew I forgot one.  This is what happens you only use zendmd and don't look in the UI.

StevePC (approved):

PR #4095, opened 2021-07-26
Improve timeout handling on unstable networks.

jpeacock-zenoss:
No new timeouts were added.  Instead, existing timeouts were handled differently to ensure that tasks will exit when a timeout occurrs.

Fixes ZEN-33565.

### DIFF
--- a/Products/ZenRRD/zencommand.py
+++ b/Products/ZenRRD/zencommand.py
@@ -451,7 +451,10 @@ class SshPerformanceCollectionTask(BaseTask):
         @return: Deferred actions to run against a device configuration
         @rtype: Twisted deferred object
         """
-        log.debug("Starting collection task  device=%s", self._devId)
+        log.debug(
+            "Starting collection task  device=%s interval=%s",
+            self._devId, self.interval,
+        )
         self._doTask_start = datetime.now()
         self.state = SshPerformanceCollectionTask.STATE_CONNECTING
         try:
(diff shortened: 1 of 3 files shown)


Products/ZenUtils/Executor.py
```diff
@@ -245,38 +266,47 @@ class ExecutorTask(object):
     @ivar call: The callable called when the task is invoked.
     """
 
-    def __init__(self, deferred, call, timeout, log):
+    def __init__(self, deferred, call, timeout, log, label=None):
```
Deer-WarLord, line 269: Probably we could assign "" as a default value for the label to exclude the if statement below

Products/ZenRRD/zencommand.py
```diff
@@ -491,11 +494,14 @@ def doTask(self):
 
         except Exception as e:
             self.state = TaskStates.STATE_PAUSED
-            err_msg = "Task paused  name=%s device-id=%s ip=%s message=%s" % (
-                self.name,
-                self._devId,
-                self._manageIp,
-                e.message)
+            err_msg = (
```
Deer-WarLord, line 497: Do we need an outer tuple for err_msg here?

jpeacock-zenoss: It's not really a tuple.  There's no trailing comma.  E.g. `()` is a tuple, `(1,)` is a tuple, but `(1)` is not a tuple.  It's just an expression. Python removes the surrounding parenthesis.

The alternative was to use to a backslash, but I don't like how they look.  Using parenthesis to break up a statement onto multiple lines has a better aesthetic, imho.  Here's what the alternative would look like:
```
        err_msg = \
            "Task paused  device=%s ip=%s interval=%s message=%s" % (
                self._devId,
                self._manageIp,
                self.interval,
                e.message,
            )
```

Products/ZenUtils/Executor.py
jpeacock-zenoss: I can do that.

Deer-WarLord (approved):

PR #4127, opened 2022-03-22
Fix UI tracebacks messages for event transforms

d-terletskyi:
Fixes ZEN-32905.

To fix type errors related to getSecurityPermissions() and
getTransformTree() which were observable for the customer through the
UI, the default value for event class was set. After changes when the
customer opens Events -> Even Classes and chose Transforms in the
drop-down menu before selecting an event class, no errors will occur,
instead of them the transforms for Events will be shown.

### DIFF
--- a/Products/ZenUI3/browser/resources/js/zenoss/Security.js
+++ b/Products/ZenUI3/browser/resources/js/zenoss/Security.js
@@ -9,90 +9,92 @@
 
 
 /* package level */
-(function() {
-     // NOTE: All permissions are handled on the server, this is just to
-     // enhance the user experience
-     Ext.ns('Zenoss.Security');
+(function () {
+    // NOTE: All permissions are handled on the server, this is just to
+    // enhance the user experience
+    Ext.ns('Zenoss.Security');
 
-     // Defined in ZenUI3/security/security.py this is a dictionary of all the
-     // permissions the current user has on the current context.
-     var all_permissions = _global_permissions(),
-         callbacks = [];
+    // Defined in ZenUI3/security/security.py this is a dictionary of all the
+    // permissions the current user has on the current context.
+    var all_permissions = _global_permissions(),
+        callbacks = [];
 
-     /**
-      * The main method for ACL on the front end. It asks
-      * if the current user has this permission.
-      * The permissions are defined in Zope
-      *
-      * NOTE: The permissions are not case-sensitive here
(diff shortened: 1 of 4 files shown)


Products/ZenUtils/extdirect/router.py
```diff
@@ -174,8 +178,15 @@ def _processDirectRequest(self, directRequest):
         if isinstance(data, (int, basestring)):
             data = {'id': data}
 
+        # Set default value for event class for transform getting without chosen event class
+        if not data:
+            if action == 'EventClassesRouter' and (method == 'getTransformTree' or method == 'isTransformEnabled'):
+                data = {'uid': 'zport/dmd/Events'}
+            elif action == 'DetailNavRouter' and method == 'getSecurityPermissions':
+                data = {'uid': 'zport/dmd/Events'}
```
jpeacock-zenoss, line 186: Rather using conditionals here to correct issues elsewhere, why not fix the issue where it originates?  The methods `getTransformTree`, `isTransformEnabled`, and `getSecurityPermissions` are each called from one place in the Javascript code.  And they're all called from within a `setContext` function that passes in the `uid` value, so something else is calling `setContext` and passing in no value for `uid`.

We should try to fix the issue at where the problem originates.

Also, `DetailNavRouter` seems to be a general purpose component and not specific to the Event Console. I'm not convinced that setting `uid` to the value "zport/dmd/Events" would be correct in all situations.

jpeacock-zenoss (changes requested):

Products/ZenUtils/extdirect/router.py
d-terletskyi: The solution was improved according to the remarks and suggestions noted above.

jpeacock-zenoss (changes requested):

Products/ZenUI3/browser/resources/js/zenoss/Security.js
```diff
@@ -78,7 +78,7 @@
       **/
      Zenoss.Security.setContext = function(uid) {
          var params = {
-             uid:uid
+             uid:uid || 'zport/dmd/Events'
```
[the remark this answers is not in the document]
d-terletskyi: The solution was improved according to the remarks and suggestions noted above.

jpeacock-zenoss (approved):

PR #4128, opened 2022-03-23
Migrate memcached services to their own image.

jpeacock-zenoss:
Remove files that are now included with the memcached image.

ZEN-30951


jpeacock-zenoss:
retest this

PR #4145, opened 2022-04-19
Feature/zen 33996: Fix adding events with an empty summary

d-terletskyi:


Deer-WarLord (approved):

Products/ZenRRD/zencommand.py
```diff
@@ -502,30 +526,24 @@ def doTask(self):
...
-            self._eventService.sendEvent(
-                STATUS_EVENT,
-                device=self._devId,
-                summary=e.message,
-                component=COLLECTOR_NAME,
-                severity=Error,
+        except defer.CancelledError:
+            message = "Twisted deferred was cancelled."
+            log.error(
+                "Task paused  device=%s ip=%s interval=%s error=%s",
```
jpeacock-zenoss, line 532: This message is not correct.  There is no "pausing" occurring in this exception handler.

Products/ZenRRD/zencommand.py
```diff
@@ -502,30 +526,24 @@ def doTask(self):
...
-                log.exception(err_msg)
-            else:
-                log.error(err_msg)
-            self._eventService.sendEvent(
-                STATUS_EVENT,
-                device=self._devId,
-                summary=e.message,
-                component=COLLECTOR_NAME,
-                severity=Error,
+        except defer.CancelledError:
```
jpeacock-zenoss, line 529: I'm not convinced that just catching CancelledError is a complete fix.  We don't know why the deferred was cancelled.  Simply logging "deferred was cancelled" provides no useful information to whoever is troubleshooting the message.

jpeacock-zenoss (changes requested):

d-terletskyi:
The fix was changed according to the remarks noted above.

d-terletskyi:
The fix was changed according to the remarks noted above.

jpeacock-zenoss (approved):

PR #4146, opened 2022-04-22
ZEN-33994: Traceback 'ModelCatalogError: Exception performing search' appears for the first jobs after CZ/RM deploy 

Deer-WarLord:


jpeacock-zenoss (changes requested):
This is a service migration.  Please implement this in the zenoss-service repository in its migrations directory.

StevePC (changes requested):

PR #4158, opened 2022-05-11
Fix issues with ADM's 'add' directive.

jpeacock-zenoss:
* When _add == False, ADM will still update data.
* Regardless of how _add is set, ADM will not update ZODB when changes are not detected.

Fixes ZEN-32767.


lundybernard (approved):
Changes look good

PR #4225, opened 2022-10-31
zendisc used ip assigned to netapp interface

vsaliieva:
Fixes ZEN-30583.

zendisc creating devices for ip addresses
assigned to a netapp components. There is because of relation ipaddress <-> manageDevice.
So when we are checking for the device existence
we check manageDevice for that ipaddress.
Since we no longer check device existence
from ipaddress<->interface relation when
creating a device that is why zendisc is discovering devices with ip that are already assigned to
the netapp components. Additional check was added
to remove such IPs from the discovering list.


Products/ZenHub/services/DiscoverService.py
```diff
@@ -303,3 +303,24 @@ def remote_moveDevice(self, dev, path):
...
+        already assigned to NetApp interfaces (nodes)
+
+        @param net - network to discover
+        @return:  a list of IPs without IP assigned for NetApp interfaces
+        @rtype: list
+        """
+
+        full_ip_list = net.fullIpList()
+
+        for d in self.dmd.Devices.getSubDevices():
```
StevePC, line 320: why not do dmd.Devices.Storage.NetApp.getSubDevices()?  would be much more efficient.

Products/ZenHub/services/DiscoverService.py
```diff
@@ -303,3 +303,24 @@ def remote_moveDevice(self, dev, path):
...
+            for interface in d.os.interfaces():
+                if '/Storage/NetApp' in interface.getDeviceClassName() and interface.getIpAddressObj():
+                    if net.netmask == interface.getIpAddressObj().netmask and interface.getIp() in full_ip_list:
```
StevePC, line 323: you should only call getIpAddressObj once, and the ip can be pulled from the IpAddress object instead of the interface.

StevePC (commented):
Why are we doing something specific to NetApp.  A decision should be made as to whether or not we consider IpAddresses on components as available or unavailable for Devices, rather than changing the behaviour for individual device or component types, especially at the platform level.
Perhaps it would be better to have an attribute on the IpAddress object that indicated whether or not it was "used" from a device discovery point of view, and then let the ZP set that appropriately.

jpeacock-zenoss (changes requested):
This change targets NetApp specifically in what should be generalized code. zendisc should be able to identify IP addresses already in use and skip them regardless of the type of device or component.

Products/ZenHub/services/DiscoverService.py
```diff
@@ -303,3 +303,25 @@ def remote_moveDevice(self, dev, path):
...
+        """
+
+        full_ip_list = net.fullIpList()
+
+        for d in self.dmd.Devices.getSubDevices():
+            for interface in d.os.interfaces():
+                ip_obj = interface.getIpAddressObj()
+                if ip_obj:
+                    if net.netmask == ip_obj.netmask and ip_obj.getIp() in full_ip_list:
+                        full_ip_list.remove(ip_obj.getIp())
```
jpeacock-zenoss, line 325: This code does not handle the case of a network interface having multiple IP addresses.

There's an "ipaddresses" relationship on the interface that will can reference multiple addresses.  The `getIpAddressObj` method is returning whatever is the first address found in `ipaddresses`.

Something like the following will cover this situation:
```
    for addr in interface.ipaddresses():
        ip = addr.getIp()
        if net.netmask == addr.netmask and ip in full_ip_list:
            full_ip_list.remove(ip)
````

**Review by jpeacock-zenoss — CHANGES_REQUESTED** [review]

File: Products/ZenHub/services/DiscoverService.py [hunk 4]
```diff
@@ -303,3 +303,25 @@ def remote_moveDevice(self, dev, path):
...
+        already assigned to interfaces (device components)
+
+        @param net - network to discover
+        @return:  a list of IPs without addresses assigned for interfaces
+        @rtype: list
+        """
+
+        full_ip_list = net.fullIpList()
+
+        for d in self.dmd.Devices.getSubDevices():
```
jpeacock-zenoss, line 320: I suggest using the `getSubDevicesGen` method instead.  The `getSubDevices` method returns a list while `getSubDevicesGen` method returns a generator.  Since this is called at the root devices organizer, this will return all the devices, and `getSubDevices` would load all the devices into memory.  Using the `getSubDevicesGen` method will load the devices one at at a time as the results are iterated over which, hopefully, won't be as memory intensive (i.e. garbage collection of device objects that aren't needed anymore).

jpeacock-zenoss (changes requested):

Products/ZenHub/services/DiscoverService.py
vsaliieva: added requested change

jpeacock-zenoss (approved):

PR #4230, opened 2023-03-17
ZEN-34083: 'Time Zone' field is blank after upgrade if time zone wasn't default

Deer-WarLord:


Products/ZenUtils/Time.py
jpeacock-zenoss, line 1: The default timezone of a container is UTC regardless of the host's timezone.

Setting the `TZ` environment variable has an effect on what `time.strftime("%Z")` returns.

For example:
```
$ date
Fri Mar 17 08:40:14 AM CDT 2023
$ docker run -it --rm zenoss/resmgr_6.7:6.7.0 bash
[root@aa8b391b61c6] date
Fri Mar 17 13:41:54 UTC 2023
[root@aa8b391b61c6] python -c "import time; print time.strftime('%Z')"
UTC
[root@aa8b391b61c6] exit
$ docker run -it --rm -e "TZ=Europe/Kiev" zenoss/resmgr_6.7:6.7.0 bash
[root@a69dfb71edec /]# date
Fri Mar 17 15:50:29 EET 2023
[root@a69dfb71edec /]# python -c "import time; print time.strftime('%Z')"
EET
[root@a69dfb71edec /]# 
```

It's odd that `time.strftime("%Z")` would return an empty string.

Here's how to reproduce the empty string issue: specify an incorrect value for TZ. For example:
```
$ docker run -it --rm -e "TZ=Europe2/Kiev" zenoss/resmgr_6.7:6.7.0 bash
[root@a69dfb71edec /]# date
Fri Mar 17 13:53:11  2023
[root@a69dfb71edec /]# python -c "import time; print time.strftime('%Z')"

[root@a69dfb71edec /]# 
```

jpeacock-zenoss, line 1: After reviewing the ticket, the issue isn't a misconfigured TZ variable, but the format of the timezone information.  The UI doesn't accept the short-hand timezone name.

jpeacock-zenoss (approved):

PR #4236, opened 2023-05-04
ZEN-32183 allow v3 traps

SerjKrukovZenoss:


webhookbotuser:

webhookbotuser:

webhookbotuser:

Products/ZenEvents/TrapFilter.py
```diff
@@ -167,7 +167,7 @@ def _parseFilterDefinition(self, line, lineNumber):
         snmpVersion = tokens[1].lower()
         if action != "include" and action != "exclude":
             return "Invalid action '%s'; the only valid actions are 'include' or 'exclude'" % tokens[0]
-        elif snmpVersion != "v1" and snmpVersion != "v2":
+        elif snmpVersion != "v1" and snmpVersion != "v2" and snmpVersion != "v3":
```
StevePC, line 170: add v3 to message

StevePC (approved):

Products/ZenEvents/TrapFilter.py
```diff
@@ -167,8 +167,8 @@ def _parseFilterDefinition(self, line, lineNumber):
         snmpVersion = tokens[1].lower()
         if action != "include" and action != "exclude":
             return "Invalid action '%s'; the only valid actions are 'include' or 'exclude'" % tokens[0]
-        elif snmpVersion != "v1" and snmpVersion != "v2":
-            return "Invalid SNMP version '%s'; the only valid versions are 'v1' or 'v2'" % tokens[1]
+        elif snmpVersion != "v1" and snmpVersion != "v2" and snmpVersion != "v3":
+            return "Invalid SNMP version '%s'; the only valid versions are 'v1' or 'v2 or v3'" % tokens[1]
```
StevePC, line 171: incorrect quoting, also doesn't match with test

StevePC (changes requested):

StevePC (approved):

Products/ZenEvents/TrapFilter.py
```diff
@@ -232,8 +233,8 @@ def _parseV1FilterDefinition(self, lineNumber, action, remainingTokens):
                 filterDef.specificTrap = remainingTokens[1]
                 if filterDef.specificTrap != "*" and not filterDef.specificTrap.isdigit():
                     return "Specific trap '%s' invalid; must be non-negative integer" % filterDef.specificTrap
-            else:
-                return "Missing specific trap number or '*'"
+            elif len(remainingTokens) == 1:
+                oid = oidOrGenericTrap
```
jpeacock-zenoss, line 237: Why assign `oid` to `oidOrGenericTrap` again?  That happened on line 226 and hasn't had its value changed.

Additionally, removing the `return` statement says that having `len(remaingTokens) != 2` is OK now.  Is that correct?

jpeacock-zenoss (changes requested):

Products/ZenEvents/TrapFilter.py
SerjKrukovZenoss: Idea behind removing return statement to allow using v1 with specific oid only, so yeah, it should be ok. Original checks are still in place, so in this case with 1 remaining token it would be specific or global oid

jpeacock-zenoss (approved):

PR #4252, opened 2023-08-23
ZEN-34454: Turning off Forward to Cloud for  CZ metrics in UI doesn't seem to work

Deer-WarLord:


Deer-WarLord:
this code shows that in some case the tags info aren't passed to the points before the actual metric is pushed to the Redis 
```
from Products.ZenHub.services.PingPerformanceConfig import PingPerformanceConfig
collector_id = "Austin"
service = PingPerformanceConfig(dmd=dmd, instance=collector_id)
service.remote_getDeviceConfigs(["10.88.120.108"])[0].monitoredIps[0].points
```
We're trying to get these points list [here](https://github.com/zenoss/zenoss-prodbin/blob/03b5a73bb0884048e82e81f16295afd7374cfe91/Products/ZenStatus/PingTask.py#L327 )

jpeacock-zenoss (approved):

PR #4257, opened 2023-10-13
Updated cProps call to use the `.items()` function of the dictionary …

dan-smalley:
…to fix the "too many items to unpack error" when attempting to load a device with cProps defined via the Device API router.


webhookbotuser:

webhookbotuser:

webhookbotuser:

PR #4267, opened 2023-12-07
zendisc SNMP Discovery is not parallelized

vsaliieva:
Fixes ZEN-30834.

Back-port for ZEN-30835 from support/6.x.

### DIFF
--- a/Products/DataCollector/zendisc.py
+++ b/Products/DataCollector/zendisc.py
@@ -72,6 +72,14 @@ def _partitionPingResults(results):
     return (good, bad)
 
 
+# From ZenRRD/zenprocess.py; should move into ZenUtils somewhere
+def chunk(lst, n):
+    """
+    Break lst into n-sized chunks
+    """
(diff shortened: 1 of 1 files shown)


Products/DataCollector/zendisc.py
```diff
@@ -178,27 +186,31 @@ def discoverRanges(self):
...
+            results = yield defer.DeferredList(
+                self.discoverDevice(ip, devicepath="/Network/Router")
+                for ip in batch_of_ips
+            )
+            observedIps.update(batch_of_ips)
+            routers.extend(
+                foundRouter
+                for _, foundRouter in results
+                if foundRouter is not None
             )
```
jpeacock-zenoss, line 213: This code looks incomplete.

I compared the changes in this PR to the actual code in the support/6.x branch and they do not match.

I think there are commits missing.

jpeacock-zenoss (changes requested):

jpeacock-zenoss (approved):

PR #4311, opened 2024-02-05
Change configcache modelapi response keys

odubrovyk:
Implements ZEN-34675.

To avoid JS errors and unnecessary code on the RM Monitor ZP side the configcahe response keys are corrected.


vsaliieva:
retest this

vsaliieva (approved):

PR #4320, opened 2024-02-09
Removed loop prevention code in Auth0

vsaliieva:
Fixes ZEN-34277.

*Removed loop prevention code in Auth0
*Added handler to display a friendly UI message
 instead of a traceback for Unauthorized case


joshw (approved):
Looks good.     From a testing perspective, we need to make sure that the role check does work as expected, so that users who have valid credentials, but no roles for this CZ, don't get access.   It looks like that is handled but just need to verify.

PR #4403, opened 2024-04-12
ZEN-34786: Products.ZenHub.services.SyslogConfig  config - is created for all devices

Deer-WarLord:


jpeacock-zenoss (approved):

Products/ZenEvents/zensyslog.py
```diff
@@ -259,6 +318,13 @@ def expand(self, msg, client_address):
         message = '%s %s %s' % (prettyTime, hostname, body)
         return message
 
+    def syslogProcessorConfChangeListener(self, observable, attrName, oldValue, newValue):
+        self.log.debug("Task %s changed %s. Updating it for task %s", observable.name, attrName, self.name)
+        if attrName == "syslogParsers":
```
StevePC, line 323: Need to handle syslogMsgEvtFieldFilterRules here as well, since _updateConfig() will never be called due to no normal collector config ever being created..

StevePC (changes requested):

Products/ZenEvents/zensyslog.py
Deer-WarLord: done

StevePC: see https://github.com/zenoss/zenoss-prodbin/blob/develop/Products/ZenEvents/zensyslog.py#L401 for how syslogMsgEvtFieldFilterRules is set on update, this is incorrect.

StevePC (changes requested):

Products/ZenEvents/zensyslog.py
Deer-WarLord: updated

StevePC (approved):

PR #4408, opened 2024-04-22
Fix: Scheduled jobs were deleted without execution

vsaliieva:
Fixes ZEN-34769.

### DIFF
--- a/Products/Jobber/model.py
+++ b/Products/Jobber/model.py
@@ -386,30 +386,33 @@ def save_scheduled_jobrecord(log, task_id=None, task=None, *args, **kwargs):
     This function is registered as a handler for the task_prerun
     signal.
     """
-    task = get_app().tasks.get(task.name)
-    if task is None:
-        log.warn("Ignoring unknown task: %s", task.name)
-        return
 
-    if task.ignore_result:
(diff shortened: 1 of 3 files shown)


Products/Jobber/model.py
```diff
@@ -380,6 +380,37 @@ def _save_record(log, storage, record):
...
+
+    record = RedisRecord.from_task(task, task_id,  args, kwargs)
+    record.update(
+        {
+            "status": states.PENDING,
+            "created": time.time(),
+        }
+    )
+
+    saved = _save_record(log, storage, record)
```
jpeacock-zenoss, line 408: I realize that `_save_record` already has a check for whether the job record already exists, but I think there should be a check here before calling `_save_record` to avoid seeing the "No user ID submitted with job ..." error twice in the log.

jpeacock-zenoss (changes requested):

Products/Jobber/model.py
vsaliieva: I added a check before _save_record. 
We agree that there are no other suitable signal 'candidates' to use instead of task_prerun

jpeacock-zenoss (approved):

PR #4432, opened 2024-05-07
Fix: purge_logs is not starting

vsaliieva:
Fixes ZEN-34832.

### DIFF
--- a/Products/Jobber/jobs/purge_logs.py
+++ b/Products/Jobber/jobs/purge_logs.py
@@ -23,7 +23,7 @@ from ..zenjobs import app
     summary="Delete the logs of deleted jobs",
     ignore_result=True,
 )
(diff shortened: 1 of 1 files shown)


Products/Jobber/jobs/purge_logs.py
```diff
@@ -22,8 +22,9 @@
     name="zen.zenjobs.purge_logs",
     summary="Delete the logs of deleted jobs",
     ignore_result=True,
+    scheduled=True,
 )
-def purge_logs(self):
+def purge_logs(self, is_scheduled=True):
```
jpeacock-zenoss, line 27: let's leave out the `is_scheduled=True` since `scheduled=True` was added.

jpeacock-zenoss (changes requested):

PR #4452, opened 2024-05-24
Fixed Zenhubworker deletion during the upgrade

vsaliieva:
Fixes ZEN-34876.

### DIFF
--- a/bin/service-migrate.py
+++ b/bin/service-migrate.py
@@ -1,15 +1,30 @@
 #!/usr/bin/env python
 
+import json
 import sys
 import servicemigration as sm
 sm.require("1.0.0")
 
 TMP_FILENAME = "/tmp/zenoss-service-migration.json"
+SERVICE_NAME_TO_REMOVE = "zenhubiworker"
+
+
+def remove_service_from_file(file_path, service_name):
+    with open(file_path, 'r') as file:
+        data = json.load(file)
+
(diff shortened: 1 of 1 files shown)


jpeacock-zenoss (changes requested):
I would prefer to remove the service definition from the EnterpriseCollector zenpack rather than add this work-around in the prodbin repo.  The EnterpriseCollector zenpack doesn't have to maintain strict backward compatibility.

vsaliieva:
discussed with @Deer-WarLord that we can modify upgrade job to remove this service:
https://zenoss.atlassian.net/browse/ZEN-34876?focusedCommentId=661688 

 
this solution doesn't work, tested by @TheSushiChef

jpeacock-zenoss:
We can avoid the problem of the service being re-installed by the EnterpriseCollector ZenPack by removing the service template from the ZenPack.  The service is found in the `ZenPacks/zenoss/EnterpriseCollector/service_definition` directory.  Removing that directory will delete the template.

We can also restrict the ZenPack to Zenoss versions >=7.20 and >=6.8.0 by modifying the `ZENOSS_COMPAT_VERS` string in the `setup.py` file to look like this:
```
ZENOSS_COMPAT_VERS = ">=6.8.0,!=7.0.*,!=7.1.*"
```

PR #4458, opened 2024-06-07
ZEN-34423: zenmodeler startat and cycletime do not work as probably intended

Deer-WarLord:


Products/DataCollector/zenmodeler.py
```diff
@@ -1047,8 +1051,6 @@ def timeMatches(self):
         Check whether the current time matches a cron-like
         specification, return a straight true or false
         """
-        if self.startat is None:
-            return True
```
jpeacock-zenoss, line 1051: When `timeMatches` is called, `startat` will never be `None`?

Deer-WarLord: yes, we checking this here https://github.com/zenoss/zenoss-prodbin/pull/4458/files#diff-ffa1241dd5b49bebb381f208827c9be117c752e396e1c2f120cdd47cda1a1161R893

Products/DataCollector/zenmodeler.py
```diff
@@ -987,6 +986,11 @@ def checkStop(self, unused=None):
             if not self.options.cycle:
                 self.stop()
             self.finished = []
+            # frequency of heartbeat rate could be 2 times per minute in case we have
+            # cron job modeling faster than 1 minute it'll be trigger a second time
+            if runTime < 60 and self.startat is not None:
+                time.sleep(60)
```
jpeacock-zenoss, line 992: This code will block the reactor; zenmodeler will stop running for 60 seconds and nothing else will happen until afterwards.  We generally do not call `time.sleep` in an async (e.g. Twisted) application.  Is this the intended behavior?

Unfortunately, zenmodeler was never completely ported to using Twisted's `inlineCallbacks` decorator, so it still heavily relies on the in-house "driver" framework.  The `checkStop` method is called from `fillCollectionSlots` which is async (i.e. has a `yield` statement), so we could make `checkStop` async as well.

This `wait` function (https://github.com/zenoss/zenoss-prodbin/blob/develop/Products/ZenUtils/Utils.py#L2097-L2119) pauses execution of the current execution stack while allowing the reactor to process other deferreds.  It's written for `inlineCallbacks`, but `wait`'s implementation should provide some ideas on getting an async "sleep" to work with the "driver" framework.

jpeacock-zenoss (changes requested):

jpeacock-zenoss (approved):

PR #4499, opened 2024-07-23
Feature: ZEN-34969

StevePC:


Products/ZenHub/server/executors/workers.py
```diff
@@ -228,8 +228,8 @@ def execute(self, worker, task):
             )
             self._handle_retry(task, ex)
         except Exception as ex:
+            self.log.exception("Unexpected failure worklist=%s, task details: %r", self.name, dict([(f,getattr(task, f)) for f in task.__slots__]))
```
jpeacock-zenoss, line 231: Let's use a dict comprehension here.  Instead of
```
dict([...])
```
write
```
{f: getattr(task, f, None) for f in task.__slots__}
```

jpeacock-zenoss (changes requested):

jpeacock-zenoss (approved):

PR #4534, opened 2024-08-28
Feature: ZEN-35044

StevePC:


jpeacock-zenoss (changes requested):
It seems like if `meta_type` attribute is missing from the `datamap.target` object in one function, it could be missing when accessed in other functions.  Perhaps all the occurrences where `meta_type` is accessed should be updated to access it using `getattr`?

PR #4663, opened 2025-04-10
swap cz default prod state threshold with device threshold

dbouchillon:
use Devices root zProdStateThreshold as zen_czProdStateThreshold metadata send device prod state threshold as zen_deviceProdStateThreshold


bharatZen (approved):

bharatZen (approved):
The code changes look good. Were you able to verify in dev or preview? by changinh the device threshiold and making sure that the tywo different ones show up in cloud?

PR #4693, opened 2025-07-14
ZEN-35382: An Error during upgrdae '-bash: line 1: /opt/zenoss/bin/fix_MultiRealmIPZenPack.sh: No such file or directory'

Deer-WarLord:


jpeacock-zenoss (changes requested):
Is this script still required?  The latest version of the MultiRealmIP ZenPack is 2.2.5 which was released in 2018.  This script fixes a bug in version 2.2.3.

If we think there are customers still running version 2.2.3 that somehow wasn't patched by a previous ZSD upgrade, then it should be added to the product-assembly repo as a patch file that the upgrade script applies.  This script should not be in the zenoss-prodbin repo.

Deer-WarLord:
I will remove the patching from the upgrade scenario. I agree that in the 6.8.0 release, it should already be fixed.

PR #4700, opened 2025-08-06
ZEN-35404: AttributeError: 'NoneType' object has no attribute 'baseFilename' in the zenjobs container log

Deer-WarLord:


src/Products/Jobber/log.py
```diff
@@ -36,6 +36,7 @@
     LoggingProxy,
     TaskLogFileHandler,
 )
+from ..Zuul.utils import try_unbrain
```
jpeacock-zenoss, line 39: So, it doesn't look like `try_unbrain` is used anywhere in this module.  Please remove unused imports.

src/Products/Jobber/log.py
```diff
@@ -362,10 +363,13 @@ class LogLevelUpdater(object):
...
-        name = (
-            logfilename.rsplit("/", 1)[-1].rsplit(".", 1)[0].replace("-", "_")
-        )
+        try:
+            logfilename = logging._handlers.get("main").baseFilename
+            name = (
+                logfilename.rsplit("/", 1)[-1].rsplit(".", 1)[0].replace("-", "_")
+            )
+        except AttributeError:
+            name = "zenjobs"
```
jpeacock-zenoss, line 372: The name won't always be "zenjobs".  Sometimes it's "config_cache".

jpeacock-zenoss (changes requested):

src/Products/Jobber/log.py
jpeacock-zenoss: Right, which is weird.  Both zenjobs and builder run the same code but are given different configurations.  The primary difference between the two is the logging configuration, the loaded job classes, and the queue they read jobs from.

jpeacock-zenoss: I have different idea on solving this: disable the "loglevel monitor" feature.

The "loglevel monitor" feature does not work as originally intended.

The idea was that a user could make a change to the loglevel config file in CC, updating the file, and the "loglevel monitor" would detect the change and automatically update the log level for the specified loggers.

However, CC does not write updated config files into running containers, so there's no change for the "loglevel monitor" to detect.  The only way to use this feature is to attach to a container and manually edit the loglevel config file, but that change applies only to that one instance.

So, since starting the "loglevel monitor" appears to cause a bug regarding logging initialization, let us disable the "loglevel monitor".

To disable "loglevel monitor", remove the `celery:signal` stanzas from the signals.zcml file that have a `handler=".log.setup_loglevel_monitor"` and a `handler="log.teardown_loglevel_monitor"`.

jpeacock-zenoss (approved):

